PR #8, opened 2022-10-11
Move docs from Pydantic

hramezani:
- Move Pydantic `docs/usage/settings` to this repo under `docs/inxex.md`
- Added docs reqirements
- Make the `make docs` command works and generate the doc

There is no change in the content and content has to be adjusted


samuelcolvin:
@hramezani @PrettyWood what do you think long term? Should we have a separate docs site for pydantic-settings?

I'm wondering about have a separate repo for docs which includes the docs for all parts of pydantic - they're so closely connected that I think they should probably be on the sam domain.

hramezani:
hmmm, didn't think about this. but yeah. I would say it's a good idea to have them in the same domain.
Then we need to just change the `usage/settings` page content of Pydantic docs later after we released V2.

I close this one.

samuelcolvin:
Agreed, I think that's the best solution long term. Lots to do, but we'll get there.

hramezani:
So, in this case, if we add/change something in pedantic-settings. Then we need to update the doc in pydantic repo?
One PR for change and one PR for doc? Or?

samuelcolvin:
Well, either we keep the docs in each repo, then somehow combine them somehow (perhaps via submodules) in a new repo pydantic-docs which takes care of building and publishing docs. Which is kind of complicated, but means you can update docs and code in the same PR.

Or, we move all docs (except docstrings obviously) to a new pydanitc-docs repo, and just accept that any change which requires changes to docs requires a companion PR on `pydantic-docs`.

Thinking about it now, it sounds like the first option would probably be preferable, even if it requires a little more setup.

PR #15, opened 2023-04-13
Reimplement setting source classes

hramezani:
Reimplement pydantic-settings based on the idea proposed in https://github.com/pydantic/pydantic-settings/pull/10#pullrequestreview-1222893531

- Added `PydanticBaseSource` abc
- Other source classes changed to inherit from abc
- Added `DotEnvSettingsSource` inherit from `EnvSettingsSource`

### DIFF
--- a/pydantic_settings/main.py
+++ b/pydantic_settings/main.py
@@ -57,16 +57,16 @@ class BaseSettings(BaseModel):
         _secrets_dir: Optional[StrPath] = None,
     ) -> Dict[str, Any]:
         # Configure built-in sources
-        init_settings = InitSettingsSource(init_kwargs=init_kwargs)
+        init_settings = InitSettingsSource(self.__class__, init_kwargs=init_kwargs)
         env_settings = EnvSettingsSource(
-            self,
+            self.__class__,
             env_nested_delimiter=(
                 _env_nested_delimiter if _env_nested_delimiter is not None else self.__config__.env_nested_delimiter
             ),
             env_prefix_len=len(self.__config__.env_prefix),
         )
         dotenv_settings = DotEnvSettingsSource(
-            self,
+            self.__class__,
             env_file=(_env_file if _env_file != env_file_sentinel else self.__config__.env_file),
             env_file_encoding=(
                 _env_file_encoding if _env_file_encoding is not None else self.__config__.env_file_encoding
@@ -77,10 +77,10 @@ class BaseSettings(BaseModel):
             env_prefix_len=len(self.__config__.env_prefix),
         )
 
-        file_secret_settings = SecretsSettingsSource(self, secrets_dir=_secrets_dir or self.__config__.secrets_dir)
+        file_secret_settings = SecretsSettingsSource(self.__class__, secrets_dir=_secrets_dir or self.__config__.secrets_dir)
         # Provide a hook to set built-in sources priority and add / remove sources
         sources = self.__config__.customise_sources(
-            self,
+            self.__class__,
             init_settings=init_settings,
             env_settings=env_settings,
             dotenv_settings=dotenv_settings,
@@ -136,11 +136,11 @@ class BaseSettings(BaseModel):
         def customise_sources(
             cls,
             settings_cls: type['BaseSettings'],
-            init_settings: SettingsSourceCallable,
-            env_settings: SettingsSourceCallable,
-            dotenv_settings: SettingsSourceCallable,
-            file_secret_settings: SettingsSourceCallable,
-        ) -> Tuple[SettingsSourceCallable, ...]:
+            init_settings: 'PydanticBaseSource',
+            env_settings: 'PydanticBaseSource',
+            dotenv_settings: 'PydanticBaseSource',
+            file_secret_settings: 'PydanticBaseSource',
+        ) -> Tuple['PydanticBaseSource', ...]:
             return init_settings, env_settings, dotenv_settings, file_secret_settings
 
         @classmethod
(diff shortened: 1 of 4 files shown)


pydantic_settings/main.py
```diff
@@ -134,98 +153,166 @@ def parse_env_var(cls, field_name: str, raw_val: str) -> Any:
     __config__: ClassVar[Type[Config]]
 
 
-class InitSettingsSource:
+class PydanticBaseSource(ABC):
```
samuelcolvin, line 156: please can you add a docstring here.

pydantic_settings/main.py
```diff
@@ -134,98 +153,166 @@ def parse_env_var(cls, field_name: str, raw_val: str) -> Any:
...
+    def __init__(self, settings_cls: Type[BaseSettings]):
+        self.settings_cls = settings_cls
+        self.config = settings_cls.__config__
+        self.case_sensitive = self.config.case_sensitive
+
+    @abstractmethod
+    def get_field_value(self, field: ModelField) -> Any:
```
samuelcolvin, line 163: docstring, and methods below.

pydantic_settings/main.py
samuelcolvin: should these all move to a separate `sources.py` module?

pydantic_settings/main.py
```diff
@@ -134,98 +153,166 @@ def parse_env_var(cls, field_name: str, raw_val: str) -> Any:
...
+    def prepare_field(self, field_name: str, field: ModelField, value: Any) -> Any:
+        if field.is_complex():
+            return json.loads(value)
+        return value
+
+    def __call__(self) -> Dict[str, Any]:
+        d: Dict[str, Any] = {}
+
+        for name, field in self.settings_cls.__fields__.items():
+            field_value = self.get_field_value(field)
```
samuelcolvin, line 175: I think this too should be wrapped an exception.

pydantic_settings/main.py
```diff
@@ -134,98 +153,166 @@ def parse_env_var(cls, field_name: str, raw_val: str) -> Any:
...
+
+            try:
+                field_value = self.prepare_field(name, field, field_value)
+            except ValueError as e:
+                raise SettingsError(f'error parsing value for field "{name}"') from e
```
samuelcolvin, line 180: this should probably say which source we're getting data from to to ease debugging.

pydantic_settings/main.py
```diff
@@ -134,98 +153,166 @@ def parse_env_var(cls, field_name: str, raw_val: str) -> Any:
...
+    def __init__(self, settings_cls: Type[BaseSettings]):
+        self.settings_cls = settings_cls
+        self.config = settings_cls.__config__
+        self.case_sensitive = self.config.case_sensitive
+
+    @abstractmethod
+    def get_field_value(self, field: ModelField) -> Any:
+        pass
+
+    def prepare_field(self, field_name: str, field: ModelField, value: Any) -> Any:
```
samuelcolvin, line 166:
```suggestion
    def prepare_field_value(self, field_name: str, field: ModelField, value: Any) -> Any:
```

pydantic_settings/main.py
```diff
@@ -134,98 +153,166 @@ def parse_env_var(cls, field_name: str, raw_val: str) -> Any:
...
         return self.init_kwargs
 
     def __repr__(self) -> str:
         return f'InitSettingsSource(init_kwargs={self.init_kwargs!r})'
 
 
-class EnvSettingsSource:
-    __slots__ = ('env_file', 'env_file_encoding', 'env_nested_delimiter', 'env_prefix_len')
+class SecretsSettingsSource(PydanticBaseSource):
+    __slots__ = ('secrets_dir',)
```
samuelcolvin, line 206:
```suggestion
```

No need to mess around with slots on this one.

pydantic_settings/main.py
```diff
@@ -120,11 +137,13 @@ def prepare_field(cls, field: ModelField) -> None:
         @classmethod
         def customise_sources(
             cls,
-            init_settings: SettingsSourceCallable,
-            env_settings: SettingsSourceCallable,
-            file_secret_settings: SettingsSourceCallable,
-        ) -> Tuple[SettingsSourceCallable, ...]:
-            return init_settings, env_settings, file_secret_settings
+            settings_cls: Type['BaseSettings'],
```
samuelcolvin, line 140: either now or in a separate PR, let's add `from __future__ import annotations as _annotations` so these don't need to be strings.

samuelcolvin (commented):
otherwise LGTM.

pydantic_settings/main.py
hramezani: move to `sources.py` and docstrings added

pydantic_settings/main.py
hramezani: wrapped and a test also added

pydantic_settings/main.py
hramezani: Setting source added to error message and tests updated

pydantic_settings/main.py
hramezani: Done

pydantic_settings/main.py
```diff
@@ -1,23 +1,25 @@
...
 
 from pydantic.config import BaseConfig, Extra
 from pydantic.fields import ModelField
 from pydantic.main import BaseModel
-from pydantic.typing import StrPath, display_as_type, get_origin, is_union
-from pydantic.utils import deep_update, path_type, sequence_like
+from pydantic.typing import StrPath, display_as_type
+from pydantic.utils import deep_update, sequence_like
+
+from pydantic_settings.sources import (
```
samuelcolvin, line 12:
```suggestion
from .sources import (
```

samuelcolvin (approved):
one tiny thing, otherwise LGTM.

Once that's fixed, please merge.

PR #17, opened 2023-04-19
Adopt pydantic-settings with Pydantic V2

hramezani:
Here is the list of changes:

- `class Config` is not valid anymore and has to be replaced with `model_config`. Example:
```
    class Settings(BaseSettings):
        apple: str

        class Config:
            env_prefix = 'foobar_'
```
has to be replaced with:
```
    class Settings(BaseSettings):
        apple: str

        model_config = ConfigDict(env_prefix='foobar_')
```
- `Model.dict()` is not available anymore and `Model.model_dump()` has to be used.
- Defining `env` for a field under `class Config` is not valid and `validation_alias` has to be used. Example:
```
    class Settings(BaseSettings):
        apple: str = ...

        class Config:
            fields = {'apple': {'env': 'BOOM'}}
```
has to be replaced with:
```
    class Settings(BaseSettings):
        apple: str = Field(None, validation_alias='BOOM')
```
- Initializing settings class with field name whenever `validation_alias` is defined does not work anymore. Example:
```
class SettingsModel(BaseSettings):
    foobar: str = Field(validation_alias='foobar_env')

SettingsModel(foobar='test')  # Raises validation error
```
- Defining `env` in `Field` is not valid anymore and `validation_alias` has to be used. Example:
```
    class Settings(BaseSettings):
        foobar: str = Field(..., env='foobar_env_name')
```
has to be replaced with:
```
    class Settings(BaseSettings):
        foobar: str = Field(validation_alias='foobar_env_name')
```
- `customise_sources` is not part of `class Config` anymore and moved to `BaseSettings` class.


pydantic_settings/main.py
```diff
@@ -1,13 +1,10 @@
 from __future__ import annotations as _annotations
 
-import warnings
-from typing import AbstractSet, Any, ClassVar, Dict, List, Optional, Tuple, Type, Union
+from typing import Any, Dict, Optional, Tuple, Type
 
-from pydantic.config import BaseConfig, Extra
-from pydantic.fields import ModelField
+from pydantic import ConfigDict, DirectoryPath, FilePath
+from pydantic._internal._utils import deep_update
```
dmontagu, line 6: Are we okay using pydantic internals here? I think it's fine since it's still our package, just want to make sure.

(If we do this, and I guess even if we don't, it might be a good idea to run the tests for pydantic_settings in the CI for pydantic.)

pydantic_settings/main.py
```diff
@@ -18,7 +15,7 @@
     SecretsSettingsSource,
 )
 
-env_file_sentinel = str(object())
+env_file_sentinel: DotenvType = FilePath('')
```
dmontagu, line 18: What motivated this change? Just curious

pydantic_settings/main.py
```diff
@@ -95,59 +107,14 @@ def _build_values(
...
-        ) -> Tuple[PydanticBaseSettingsSource, ...]:
-            return init_settings, env_settings, dotenv_settings, file_secret_settings
-
-        @classmethod
-        def parse_env_var(cls, field_name: str, raw_val: str) -> Any:
-            return cls.json_loads(raw_val)
-
-    # populated by the metaclass using the Config class defined above, annotated here to help IDEs only
-    __config__: ClassVar[Type[Config]]
+    model_config = ConfigDict(  # type: ignore[typeddict-item]
```
dmontagu, line 110: Should we make a subclass of ConfigDict that adds more items? Maybe that doesn't work well.

Probably we'll want to make this work better with the pydantic type hints at some point (annoying that TypedDicts are generally treated as invariant..). But I guess that's outside the scope of a pydantic settings issue.

pydantic_settings/utils.py
```diff
@@ -0,0 +1,24 @@
...
+    'is_mount': 'mount point',
+    'is_symlink': 'symlink',
+    'is_block_device': 'block device',
+    'is_char_device': 'char device',
+    'is_fifo': 'FIFO',
+    'is_socket': 'socket',
+}
+
+
+def path_type(p: Path) -> str:
```
dmontagu, line 15: Minor thing, I'd suggest renaming this to `path_type_label` though, to indicate that the returned value is purely meant for display purposes, not for equality checks or anything like that.

dmontagu (approved):
LGTM! 

I did try to break some stuff related to the implementation of `field_is_complex` and the only thing I could get to not work was generic dataclasses, which had an easy fix.

I didn't look at the changes to the tests _super_ closely, but from what I could tell they all seemed reasonable. (If you want specific feedback on any of that let me know.)

pydantic_settings/main.py
hramezani: > Are we okay using pydantic internals here? I think it's fine since it's still our package, just want to make sure.

Yes, as we maintain Pydantic I think it's fine

> (If we do this, and I guess even if we don't, it might be a good idea to run the tests for pydantic_settings in the CI for pydantic.)

Agree

pydantic_settings/main.py
hramezani: make mypy happy

samuelcolvin (commented):
Otherwise LGTM.

Sorry for the slow response.

pydantic_settings/main.py
hramezani: `SettingsConfigDict` added

pydantic_settings/main.py
```diff
@@ -1,13 +1,10 @@
 from __future__ import annotations as _annotations
 
-import warnings
-from typing import AbstractSet, Any, ClassVar, Dict, List, Optional, Tuple, Type, Union
+from typing import Any, Dict, Optional, Tuple, Type
 
-from pydantic.config import BaseConfig, Extra
-from pydantic.fields import ModelField
+from pydantic import ConfigDict, DirectoryPath, FilePath
```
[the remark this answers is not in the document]
hramezani: `DirectoryPath` and `FilePath` replaced by `Path`

pydantic_settings/main.py
```diff
@@ -48,40 +45,55 @@ def __init__(
...
 
+    @classmethod
+    def customise_sources(
+        cls,
+        settings_cls: Type[BaseSettings],
+        init_settings: PydanticBaseSettingsSource,
+        env_settings: PydanticBaseSettingsSource,
+        dotenv_settings: PydanticBaseSettingsSource,
+        file_secret_settings: PydanticBaseSettingsSource,
+    ) -> Tuple[PydanticBaseSettingsSource, ...]:
```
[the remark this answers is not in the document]
hramezani: I created an issue for this https://github.com/pydantic/pydantic-settings/issues/19

hramezani:
Thanks @samuelcolvin for the review. I've addressed all of them.

pydantic_settings/main.py
```diff
@@ -48,40 +55,55 @@ def __init__(
             )
         )
 
-    def _build_values(
+    @classmethod
+    def customise_sources(
```
samuelcolvin, line 59:
```suggestion
    def settings_customise_sources(
```

pydantic_settings/main.py
```diff
@@ -18,7 +16,16 @@
...
+env_file_sentinel: DotenvType = Path('')
+
+
+class SettingsConfigDict(ConfigDict):
+    case_sensitive: bool
+    env_prefix: str
+    env_file: Optional[DotenvType]
+    env_file_encoding: Optional[str]
+    env_nested_delimiter: Optional[str]
+    secrets_dir: Optional[Union[str, Path]]
```
samuelcolvin, line 29: These should become `NotRequired`, then be omited below.

samuelcolvin: having looked, I think this is fine.

PR #18, opened 2023-04-21
Adopt pydantic-settings docs for Pydantic V2

hramezani:
It's based on the change on https://github.com/pydantic/pydantic-settings/pull/17 to make pydantic-settings compatible with Pydantic V2.
All the docs and examples changed to be compatible


docs/examples/settings_add_custom_source.py
samuelcolvin, line 1: please move these inline, into the markdown files.

README.md
```diff
@@ -5,3 +5,5 @@
 [![license](https://img.shields.io/github/license/pydantic/pydantic-settings.svg)](https://github.com/pydantic/pydantic-settings/blob/main/LICENSE)
 
 **Work in Progress**, see https://github.com/pydantic/pydantic/pull/4492
+
+{!docs/index.md!}
```
samuelcolvin, line 9: I don't think we should include this, since it won't work in github.

samuelcolvin (commented):
Otherwise looks like a great start.

docs/examples/settings_add_custom_source.py
hramezani: I've converted these examples to be compatible with `V2` and added them here in `examples` directory to be compatible with Pydantic documents.
Also, I've added them directly in `index.md` as well.
When we decided about the docs repo and the way that we want to handle them we can we can use them directly.

docs/index.md
```diff
@@ -0,0 +1,534 @@
...
+    'pg_dsn': Url('postgres://user:pass@localhost:5432/foobar'),
+    'amqp_dsn': Url('amqp://user:pass@localhost:5672/'),
+    'special_function': <built-in function cos>,
+    'domains': set(),
+    'more_settings': {'foo': 'bar', 'apple': 1},
+}
+"""
+```
+
+<!--- {!.tmp_examples/settings_main.md!} --->
```
hramezani, line 72: As you can see the example code is injected directly here in the md file and this line is commented. Whenever we decided about the docs, we can uncomment this line and remove the above example code

docs/examples/settings_add_custom_source.py
hramezani: see https://github.com/pydantic/pydantic-settings/pull/18/files#r1175087113

README.md
hramezani: My plan was to make new docs available for the users because the old docs are outdated and we need to somehow share the new docs with the user.

samuelcolvin (commented):
we need to remove the `.py` files and the comments.

At some point we should enable `pytest-examples` to test the docs examples.

samuelcolvin:
LGTM, please enable auto-merge if you're happy with this.

hramezani:
@samuelcolvin I just create an issue for `pytest-examples` integration https://github.com/pydantic/pydantic-settings/issues/33

PR #22, opened 2023-04-25
Implement proper support for nested complex env values

hramezani:
Port https://github.com/pydantic/pydantic/pull/4216

### DIFF
--- a/pydantic_settings/sources.py
+++ b/pydantic_settings/sources.py
@@ -290,6 +290,25 @@ class EnvSettingsSource(PydanticBaseEnvSettingsSource):
 
     @staticmethod
     def next_field(field: Optional[FieldInfo], key: str) -> Optional[FieldInfo]:
+        """
+        Find the field in a sub model by key(env name)
+
+        By having the following models:
+
+            class SubSubModel(BaseSettings):
+                dvals: Dict
+
+            class SubModel(BaseSettings):
+                vals: List[str]
+                sub_sub_model: SubSubModel
+
+            class Cfg(BaseSettings):
+                sub_model: SubModel
+
+        Then:
+            next_field(sub_model, 'vals') Returns the `vals` field of `SubModel` class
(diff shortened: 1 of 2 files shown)


pydantic_settings/sources.py
```diff
@@ -288,6 +288,16 @@ def _field_is_complex(self, field: FieldInfo) -> Tuple[bool, bool]:
 
         return True, allow_parse_failure
 
+    @staticmethod
+    def next_field(field: Optional[FieldInfo], key: str) -> Optional[FieldInfo]:
+        if not field or origin_is_union(get_origin(field.annotation)):
```
samuelcolvin, line 293: can you add a docstring?

pydantic_settings/sources.py
```diff
@@ -307,8 +317,24 @@ def explode_env_vars(
...
+                target_field = self.next_field(target_field, key)
                 env_var = env_var.setdefault(key, {})
+
+            # get proper field with last_key
+            target_field = self.next_field(target_field, last_key)
+
+            # check if env_val maps to a complex field and if so, parse the env_val
+            if target_field and env_val:
+                is_complex, allow_json_failure = self._field_is_complex(target_field)
+                if is_complex:
```
samuelcolvin, line 332: do we have a way to disable json parsing?

samuelcolvin (commented):
otherwise I think looks good, but I'm a bit unclear how this works, will be easier to review when `next_field` has a docstring.

pydantic_settings/sources.py
hramezani: added!

pydantic_settings/sources.py
hramezani: I've tried to disable it here but it doesn't work
I haven't looked deeply on it but seems we need it here

samuelcolvin:
thanks.

PR #31, opened 2023-04-26
Change the behavior of field_is_complex to be more similar to pydantic v1

dmontagu:
I discovered this discrepancy while helping Sebastian with FastAPI. @hramezani I'll defer to you whether the change makes sense. Feel free to modify the implementation, add tests, etc. Or request that I do it 😄

### DIFF
--- a/pydantic_settings/sources.py
+++ b/pydantic_settings/sources.py
@@ -13,7 +13,7 @@ from pydantic import BaseModel
 from pydantic._internal._typing_extra import origin_is_union
 from pydantic._internal._utils import deep_update, lenient_issubclass
 from pydantic.fields import FieldInfo
-from typing_extensions import get_args, get_origin
+from typing_extensions import get_origin
 
 from pydantic_settings.utils import path_type_label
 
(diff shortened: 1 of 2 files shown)


samuelcolvin (commented):
i think the behaviour is fine, but we should keep the method on the source so uses can override it more easily.

pydantic_settings/sources.py
```diff
@@ -4,15 +4,16 @@
...
 from pathlib import Path
-from typing import TYPE_CHECKING, Any, Dict, List, Mapping, Optional, Tuple, Type, Union
+from typing import TYPE_CHECKING, Any, Dict, List, Mapping, Optional, Sequence, Tuple, Type, Union
 
 from pydantic import BaseModel
 from pydantic._internal._typing_extra import origin_is_union
 from pydantic._internal._utils import deep_update, lenient_issubclass
 from pydantic.fields import FieldInfo
-from typing_extensions import get_origin
+from typing_extensions import get_args, get_origin
```
hramezani, line 16:
```suggestion
from typing_extensions import get_origin
```

hramezani (approved):
LGTM 👍

hramezani:
Thanks @dmontagu

PR #34, opened 2023-04-27
Fix nested model problem when case_sensitive=False

hramezani:
Fixes https://github.com/pydantic/pydantic-settings/issues/9

### DIFF
--- a/pydantic_settings/sources.py
+++ b/pydantic_settings/sources.py
@@ -150,6 +150,7 @@ class PydanticBaseEnvSettingsSource(PydanticBaseSettingsSource):
 
         By having the following models:
 
+            ```py
             class SubSubSub(BaseModel):
                 VaL3: str
 
@@ -165,6 +166,7 @@ class PydanticBaseEnvSettingsSource(PydanticBaseSettingsSource):
                 nested: Sub
 
                 model_config = ConfigDict(env_nested_delimiter='__')
+            ```
 
         Then:
             _replace_field_names_case_insensitively(
@@ -360,6 +362,7 @@ class EnvSettingsSource(PydanticBaseEnvSettingsSource):
 
         By having the following models:
 
+            ```py
             class SubSubModel(BaseSettings):
                 dvals: Dict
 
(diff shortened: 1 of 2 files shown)


pydantic_settings/sources.py
```diff
@@ -144,6 +144,61 @@ def _extract_field_info(self, field: FieldInfo, field_name: str) -> List[Tuple[s
...
+            Returns {'VAL1': 'v1', 'SUB_sub': {'Val2': 'v2', 'SUB_sub_SuB': {'VaL3': 'v3'}}}
+        """
+        values: Dict[str, Any] = {}
+
+        for name, value in field_values.items():
+            sub_model_field: Optional[FieldInfo] = None
+
+            if not field.annotation:
+                values[name] = value
+                continue
```
samuelcolvin, line 183: are you sure this is necessary? It looks like where this is called `field.annotations` is always a `BaseModel` - also the next line assumes `field.annotation.model_fields` always exists.

samuelcolvin (commented):
otherwise LGTM.

pydantic_settings/sources.py
hramezani: It's here only to make mypy happy
```
Item "None" of "Optional[Type[Any]]" has no attribute "model_fields"
```

hramezani: Done!

hramezani:
@samuelcolvin I've also added a comment to ignore fields with `validation_alias`. a test also has been added

pydantic_settings/sources.py
```diff
@@ -144,6 +144,63 @@ def _extract_field_info(self, field: FieldInfo, field_name: str) -> List[Tuple[s
 
         return field_info
 
+    def _replace_field_names_case_insensitively(self, field: FieldInfo, field_values: Dict[str, Any]) -> Dict[str, Any]:
+        """
+        Replace field names in values dict by looking in models fields insensitively.
+
+        By having the following models:
+
+            class SubSubSub(BaseModel):
```
samuelcolvin, line 153: one comment (doesn't matter much here since I guess we won't add this to the documentation) - if you want code in docsting, please use markdown with code fences so they examples render properly.

hramezani: Nice one. Done!

Issue #11, opened 2022-12-07
Extra.Forbid is not applied when using BaseSettings with Config.env_file defined


jberthier-ma:
### Initial Checks

- [X] I have searched GitHub for a duplicate issue and I'm sure this is something new
- [X] I have searched Google & StackOverflow for a solution and couldn't find anything
- [X] I have read and followed [the docs](https://pydantic-docs.helpmanual.io) and still think this is a bug
- [X] I am confident that the issue is with pydantic (not my code, or another library in the ecosystem like [FastAPI](https://fastapi.tiangolo.com) or [mypy](https://mypy.readthedocs.io/en/stable))

### Description

Reading the documentation and as BaseSettings inherits from BaseModel it'd be expected to have `Config.extra` parameters working when `Config.env_file` is defined.

It appears that without Config.env files it works as expected, but when it comes from env_files the extra config is ignored.

If this is expected I can make a PR to make the documentation more explicit on this specific point.

### Example Code

```Python
from pydantic import BaseSettings, Extra

""""
# .env.test
title="cool stuff"
other=42
"""

class Plop(BaseSettings):
    title: str

    class Config:
        extra: Extra = Extra.forbid
        env_file = ".env.test"

Plop()
```

### Python, Pydantic & OS Version

```Text
pydantic version: 1.10.2
            pydantic compiled: True
                 install path: .venv/lib/python3.11/site-packages/pydantic
               python version: 3.11.0 (main, Oct 24 2022, 19:55:51) [GCC 9.4.0]
                     platform: Linux-5.17.0-051700-generic-x86_64-with-glibc2.31
     optional deps. installed: ['dotenv', 'typing-extensions']
```

### Affected Components

- [ ] [Compatibility between releases](https://pydantic-docs.helpmanual.io/changelog/)
- [X] [Data validation/parsing](https://pydantic-docs.helpmanual.io/usage/models/#basic-model-usage)
- [ ] [Data serialization](https://pydantic-docs.helpmanual.io/usage/exporting_models/) - `.dict()` and `.json()`
- [ ] [JSON Schema](https://pydantic-docs.helpmanual.io/usage/schema/)
- [ ] [Dataclasses](https://pydantic-docs.helpmanual.io/usage/dataclasses/)
- [ ] [Model Config](https://pydantic-docs.helpmanual.io/usage/model_config/)
- [ ] [Field Types](https://pydantic-docs.helpmanual.io/usage/types/) - adding or changing a particular data type
[...]

PR #39, opened 2023-04-28
Considering extra config in dotenv source

hramezani:
Fixes https://github.com/pydantic/pydantic-settings/issues/11

### DIFF
--- a/pydantic_settings/sources.py
+++ b/pydantic_settings/sources.py
@@ -472,11 +472,21 @@ class DotEnvSettingsSource(EnvSettingsSource):
     def __call__(self) -> Dict[str, Any]:
         data: Dict[str, Any] = super().__call__()
 
+        data_lower_keys: List[str] = []
+        if not self.settings_cls.model_config.get('case_sensitive', False):
+            data_lower_keys = [x.lower() for x in data.keys()]
+
         # As `extra` config is allowed in dotenv settings source, We have to
         # update data with extra env variabels from dotenv file.
-        for k, v in self.env_vars.items():
-            if v is not None and k not in data:
-                data[k] = v
+        for env_name, env_value in self.env_vars.items():
+            if env_value is not None:
+                env_name_without_prefix = env_name[self.env_prefix_len :]
+                first_key, *_ = env_name_without_prefix.split(self.env_nested_delimiter)
+
+                if (data_lower_keys and first_key not in data_lower_keys) or (
+                    not data_lower_keys and first_key not in data
+                ):
+                    data[first_key] = env_value
 
         return data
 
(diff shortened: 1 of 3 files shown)


pydantic_settings/sources.py
```diff
@@ -449,12 +449,7 @@ def __init__(
         super().__init__(settings_cls, env_nested_delimiter, env_prefix_len)
 
     def _load_env_vars(self) -> Mapping[str, Optional[str]]:
-        env_vars = super()._load_env_vars()
-        dotenv_vars = self._read_env_files(self.settings_cls.model_config.get('case_sensitive', False))
-        if dotenv_vars:
-            env_vars = {**dotenv_vars, **env_vars}
-
-        return env_vars
+        return self._read_env_files(self.settings_cls.model_config.get('case_sensitive', False))
```
samuelcolvin, line 452: I'm a bit surprised that `DotEnvSettingsSource` inherits from `EnvSettingsSource` - are you sure that's how it should be?

It might be cleaner if they were separate sources and we use the normal logic to merge them?

pydantic_settings/sources.py
```diff
@@ -474,14 +469,27 @@ def _read_env_files(self, case_sensitive: bool) -> Mapping[str, Optional[str]]:
 
         return dotenv_vars
 
+    def __call__(self) -> Dict[str, Any]:
+        data: Dict[str, Any] = super().__call__()
+
+        # As `extra` config is allowed in dotenv settings source, We have to
+        # update data with extra env variabels from dotenv file.
+        for k, v in self.env_vars.items():
+            if v is not None and k not in data:
```
samuelcolvin, line 478: why `v is not None`?

samuelcolvin (commented):
otherwise seems ok

pydantic_settings/sources.py
hramezani: Actually, the env settings source and dotenv setting source were in the same class and all the logic was the same in `V1`(EnvSettingsSource). There was only one logic related to dotenv, loading env variables from dotenv file.

There was some logic like `explode_env_vars` which has to be done on both sources.
Actually, dotenv is not a separate source, it's a just some functionality for loading env variables.

We can:

- Keep these classes as is
- Merge these two classes into one class
- Separate these two classes as two separate sources and it will introduce a breaking change

pydantic_settings/sources.py
hramezani: Otherwise `test_env_file_none` fails

hramezani:
I've added a commit and changed the logic to handle case insensitive and sub-model

pydantic_settings/sources.py
samuelcolvin: okay, sounds good.

samuelcolvin:
thanks so much.

PR #59, opened 2023-05-19
Use updated FieldInfo from https://github.com/pydantic/pydantic/pull/5800

dmontagu:
Chicken and egg problem for updating pydantic; I'm proposing to disable pydantic-settings testing in https://github.com/pydantic/pydantic/pull/5800 so we can get these two PRs merged, then re-enable it once things are back to working.

### DIFF
--- a/pydantic_settings/sources.py
+++ b/pydantic_settings/sources.py
@@ -7,9 +7,9 @@ from abc import ABC, abstractmethod
 from collections import deque
 from dataclasses import is_dataclass
 from pathlib import Path
-from typing import TYPE_CHECKING, Any, List, Mapping, Sequence, Tuple, Union
+from typing import TYPE_CHECKING, Any, List, Mapping, Sequence, Tuple, Union, cast
 
-from pydantic import BaseModel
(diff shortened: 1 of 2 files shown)


pydantic_settings/sources.py
```diff
@@ -121,19 +121,23 @@ def _extract_field_info(self, field: FieldInfo, field_name: str) -> list[tuple[s
...
+        else:
+            v_alias = field.validation_alias
 
         if v_alias:
             if isinstance(v_alias, list):  # AliasChoices, AliasPath
                 for alias in v_alias:
                     if isinstance(alias, str):  # AliasPath
                         field_info.append((alias, self._apply_case_sensitive(alias), True if len(alias) > 1 else False))
                     elif isinstance(alias, list):  # AliasChoices
+                        first_arg = cast(str, alias[0])  # first item of an AliasPath must be a str
```
hramezani, line 138:
```suggestion
                        first_arg = cast(str, alias[0])  # first item of an AliasChoices must be a str
```

hramezani (approved):

Issue #19, opened 2023-04-24
Implemet protected namespaces


hramezani:
Implement protected namespace in `BaseSettings`. we need to:

1. add docstrings to this and other methods
2. rename these methods to all use a consistent prefix, maybe `settings_`
3. Implement https://github.com/pydantic/pydantic/issues/4915 and use that to make `settings_` a protected namespace.

PR #64, opened 2023-06-02
Update pydantic to 2.0b1

hramezani:
Also, Fixes https://github.com/pydantic/pydantic-settings/issues/19


hramezani:
please review

Issue #63, opened 2023-06-01
feature request：load the config should log the load source by debug level


Teklvtn:
When pydantic loads the configuration, it records what data source the configuration is read from, and provides debug-level log output.i think if the feature is added, this library will work better.

By reading the source code, I found that it is currently unable to support recording the configuration source. This can be achieved by encapsulating the chainmap, or by adding an additional dict to record. Due to my limited level, I am afraid that direct modification of the source code will not completely cover it. testing, thus introducing more bugs, so I hope the development team can add this feature.

hramezani:
> I tried to submit the code to pydantic v1, and later found that the development team separated the setting from pydantic into a sub-project

Yeah. `pydantic-settings` is a separate package after `V2`

Yeah `deep_update` is in `Pydantic` package and `pydantic-settings` used it. it will be deprecated in `V2` and we might need to move it to `pydantic-settings`.

I think even by changing the `deep_update` it is hard to say the field value comes from which source. Consider the following field:
```py
a: int = Field(validation_alias='a_alias')
```

and two sources with these values:

```py
s1 = {'a': 1}
s2 = {'a_alias': 2}
```

`deep_update` returns a dict like `{'a': 1, 'a_alias': 2}`. Do you know at this point which value has to be selected for the field? you may find it but I am afraid it would be complicated.

At the end `Pydantic` validates the data and selects data for fields (based on field_name, alias, alias_generator, and some config).

hramezani:
We may can log values loaded from sources and at the end log the final values that has to be passed to pydantic. something like:

```
DEBUG    pydantic_settings.main:main.py:120 {'foo': 'foo_secret_value_str'} loaded from SecretsSettingsSource
DEBUG    pydantic_settings.main:main.py:120 {'foo': 'foo_env_value_str'} loaded from DotEnvSettingsSource
DEBUG    pydantic_settings.main:main.py:120 {} loaded from EnvSettingsSource
DEBUG    pydantic_settings.main:main.py:120 {} loaded from InitSettingsSource
DEBUG    pydantic_settings.main:main.py:123 {'foo': 'foo_env_value_str'} is the final settings values
```

we can easily have it by 

```diff
diff --git a/pydantic_settings/main.py b/pydantic_settings/main.py
index 88478f1..05a7f5a 100644
--- a/pydantic_settings/main.py
+++ b/pydantic_settings/main.py
@@ -1,5 +1,6 @@
 from __future__ import annotations as _annotations

[...]

PR #68, opened 2023-06-02
feature：logging config loaded source

oslijunw:
for :#63

### DIFF
--- a/pydantic_settings/main.py
+++ b/pydantic_settings/main.py
@@ -18,7 +18,7 @@ from .sources import (
     SecretsSettingsSource,
 )
 
-log = logging.getLogger('pydantic')
+log = logging.getLogger('pydantic.settings')
 env_file_sentinel: DotenvType = Path('')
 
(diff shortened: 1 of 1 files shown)


oslijunw:
I don't know how to fix the lint to pass the first test, I haven't changed much in the code, so can't figure out how to solve this problem with the ci process

pydantic_settings/main.py
```diff
@@ -16,6 +18,7 @@
     SecretsSettingsSource,
 )
 
+log = logging.getLogger('pydantic')
```
Kludex, line 21: Maybe `pydantic.settings`? I'm not sure what's the best name here.

oslijunw: tks for your reply, I will change that

hramezani:
Close as per comment [here](https://github.com/pydantic/pydantic-settings/issues/63#issuecomment-1577441967)

PR #75, opened 2023-06-13
✨ Make `SettingsConfigDict` public

Kludex:


pydantic_settings/__init__.py
```diff
@@ -10,11 +10,12 @@
 
 __all__ = (
```
Kludex, line 11: Sorted alphabetically. I can revert if wanted.

hramezani (approved):

PR #76, opened 2023-06-13
Tweak sources to have `case_sensitive` and `env_prefix` usable as args

### LINKED ISSUE #74: Add `env_prefix` to `BaseSettings` args
Currently `BaseSettings` can take the following config values from its `__init__` method:

https://github.com/pydantic/pydantic-settings/blob/c05410f4bd380fa2769bb9a674d32db0f2353d0f/pydantic_settings/main.py#L47-L54

Then these values are used if provided, else the ones from the `SettingsConfigDict`, e.g.:

https://github.com/pydantic/pydantic-settings/blob/c05410f4bd380fa2769bb9a674d32db0f2353d0f/pydantic_settings/main.py#L109-L114

Is there a reason why `env_prefix` isn't part of this scheme? If this is possible, I can make a PR implementing this.

Viicos:
Fixes #74.

Ended up doing a bit of refactoring, the idea is to avoid accessing `self.config` (even `self.settings_cls.model_config` in some places, which points to the same config object) in sources, and determine the values to use in `_settings_build_values` instead.

Alternatively we could have some kind of "context" object, created by merging args values from `_settings_build_values` and values from the `model_config`. This way we wouldn't have to pass each config value by hand to the sources' `__init__`.

Waiting for review before fixing tests and updating documentation.

Selected Reviewer: @hramezani

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -90,7 +90,7 @@ print(Settings().model_dump())
 
 By default, the environment variable name is the same as the field name.
 
-You can change the prefix for all environment variables by setting the `env_prefix` config setting, 
+You can change the prefix for all environment variables by setting the `env_prefix` config setting,
 or via the `_env_prefix` keyword argument on instantiation:
 
 ```py
--- a/docs/index.md
+++ b/docs/index.md
@@ -86,6 +86,37 @@ print(Settings().model_dump())
 
     Check the [Environment variable names documentation](#environment-variable-names) for more information.
 
+## Additionnal config values
+
+The following extra configuration values are available when using `BaseSettings`:
+
+* `case_sensitive`: Whether environment variables names should be read with case-sensitivity. Defaults to `False`.
+* `env_prefix`: Prefix for all environment variables. Defaults to `''`.
+* `env_file`: The env file(s) to load settings values from. Defaults to `None`.
+* `env_file_encoding`: The env file encoding, e.g. `'latin-1'`. Defaults to `None`.
+* `env_nested_delimiter`: The nested env values delimiter. Defaults to `None`.
+* `secrets_dir`: The secret files directory. Defaults to `None`.
+
+All of these values will be described in the following sections.
+
+!!! note
+    For better auto-completion, `SettingsConfigDict` can be used.
+
+All of these configuration options can be set during initialization:
+
+```py
+from pydantic_settings import BaseSettings, SettingsConfigDict
+
+
+class Settings(BaseSettings):
+    v1: str = ''
+    model_config = SettingsConfigDict(env_prefix='my_prefix_')
+
+
+# Initialization values take priority over `model_config`.
+Settings(_env_prefix='other_prefix_')
+```
+
 ## Environment variable names
 
 By default, the environment variable name is the same as the field name.
@@ -316,7 +347,7 @@ in the `BaseSettings` class:
 
 ```py test="skip" lint="skip"
 class Settings(BaseSettings):
-    model_config = SettingsConfigDict(env_file='.env', env_file_encoding = 'utf-8')
+    model_config = SettingsConfigDict(env_file='.env', env_file_encoding='utf-8')
 ```
 
 2. Instantiating the `BaseSettings` derived class with the `_env_file` keyword argument
--- a/docs/index.md
+++ b/docs/index.md
@@ -86,42 +86,12 @@ print(Settings().model_dump())
 
     Check the [Environment variable names documentation](#environment-variable-names) for more information.
 
-## Additionnal config values
-
-The following extra configuration values are available when using `BaseSettings`:
-
-* `case_sensitive`: Whether environment variables names should be read with case-sensitivity. Defaults to `False`.
-* `env_prefix`: Prefix for all environment variables. Defaults to `''`.
-* `env_file`: The env file(s) to load settings values from. Defaults to `None`.
-* `env_file_encoding`: The env file encoding, e.g. `'latin-1'`. Defaults to `None`.
-* `env_nested_delimiter`: The nested env values delimiter. Defaults to `None`.
-* `secrets_dir`: The secret files directory. Defaults to `None`.
-
-All of these values will be described in the following sections.
-
-!!! note
-    For better auto-completion, `SettingsConfigDict` can be used.
-
-All of these configuration options can be set during initialization:
-
-```py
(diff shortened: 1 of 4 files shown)


pydantic_settings/main.py
```diff
@@ -90,38 +94,46 @@ def settings_customise_sources(
...
+        env_file = _env_file if _env_file != env_file_sentinel else self.model_config.get('env_file')
+        env_file_encoding = (
+            _env_file_encoding if _env_file_encoding is not None else self.model_config.get('env_file_encoding')
+        )
+        env_nested_delimiter = (
+            _env_nested_delimiter
+            if _env_nested_delimiter is not None
+            else self.model_config.get('env_nested_delimiter')
+        )
+        secrets_dir = _secrets_dir or self.model_config.get('secrets_dir')
```
Viicos, line 116: Should we use the `_val or self.model_config.get(val)` form for each of this settings instead, as it is done for `secrets_dir`? (kept it the way it was for each value)

pydantic_settings/sources.py
```diff
@@ -331,18 +340,18 @@ class EnvSettingsSource(PydanticBaseEnvSettingsSource):
...
         env_nested_delimiter: str | None = None,
-        env_prefix_len: int = 0,
-    ):
-        super().__init__(settings_cls)
-
-        self.env_nested_delimiter: str | None = env_nested_delimiter
-        self.env_prefix_len: int = env_prefix_len
+    ) -> None:
+        super().__init__(settings_cls, case_sensitive, env_prefix)
+        self.env_nested_delimiter = env_nested_delimiter
```
Viicos, line 348: `env_nested_delimiter` is `str | None` here, isn't it an issue regarding [this part](https://github.com/pydantic/pydantic-settings/blob/d4a7201166c34692525eea1601a9f612ee26f418/pydantic_settings/sources.py#L475-L484)?

Viicos:
Please review

pydantic_settings/sources.py
```diff
@@ -521,16 +530,16 @@ def __init__(
...
-        env_prefix_len: int = 0,
-    ):
-        self.env_file: DotenvType | None = env_file
-        self.env_file_encoding: str | None = env_file_encoding
-
-        super().__init__(settings_cls, env_nested_delimiter, env_prefix_len)
+    ) -> None:
+        super().__init__(settings_cls, case_sensitive, env_prefix, env_nested_delimiter)
+        self.env_file = env_file
+        self.env_file_encoding = env_file_encoding
```
hramezani, line 539:
```suggestion
        self.env_file = env_file
        self.env_file_encoding = env_file_encoding
        super().__init__(settings_cls, case_sensitive, env_prefix, env_nested_delimiter)
```

pydantic_settings/sources.py
hramezani: If `env_name='foo'` then `f'{env_name}{self.env_nested_delimiter}'` will be converted to `'fooNone'` which is not raising an error but it's better to handle it better.
we can return at the beginning of the function by checking the `env_nested_delimiter is None`.
If you want to fix it, please do it in a separate PR. otherwise I will fix it

hramezani:
@Viicos Thanks for this PR.
The implementation LGTM!

Please continue with the tests and documentation

hramezani:
please update

Viicos:
Basic tests added, updated docstrings and documentation (not sure the added section is required, as each section describing `BaseSettings` related config options is already describing both behaviors -- `model_config` vs init args).

Viicos:
Please reveiew

pydantic_settings/main.py
hramezani: Yes

pydantic_settings/main.py
```diff
@@ -38,8 +38,11 @@ class BaseSettings(BaseModel):
     All the bellow attributes can be set via `model_config`.
 
     Args:
+        _case_sensitive: Whether environment variables names should be read with case-sensitivity. Defaults to `None`.
+        _env_prefix: Prefix for all environment variables. Defaults to `None`.
         _env_file: The env file(s) to load settings values from. Defaults to `Path('')`.
         _env_file_encoding: The env file encoding. e.g. `'latin-1'`. Defaults to `None`.
+        _env_file_encoding: The env file encoding, e.g. `'latin-1'`. Defaults to `None`.
```
hramezani, line 45: Isn't this one duplicate?

pydantic_settings/main.py
```diff
@@ -19,7 +19,7 @@
 env_file_sentinel: DotenvType = Path('')
 
 
-class SettingsConfigDict(ConfigDict, total=False):
+class SettingsConfigDict(ConfigDict):
```
hramezani, line 22: Why did you remove the `total=False`?

docs/index.md
```diff
@@ -86,6 +86,37 @@ print(Settings().model_dump())
 
     Check the [Environment variable names documentation](#environment-variable-names) for more information.
 
+## Additionnal config values
```
hramezani, line 89: I would suggest removing this section and adding doc for `_case_sensitive` to the section `### Case-sensitivity` and `_env_prefix` to the section `## Environment variable names`.

The related config keys are documented there

hramezani:
please update

docs/index.md
Viicos: Ok will do

pydantic_settings/main.py
Viicos: Error while rebasing

pydantic_settings/main.py
Viicos: Error while rebasing

pydantic_settings/main.py
Viicos: In fact better to keep the strict `None` equality check, otherwise this could lead to unexpected results (e.g. with `_case_sensitive=False`, `case_sensitive = _case_sensitive or self.model_config.get('case_sensitive')` would evaluate to `self.model_config.get('case_sensitive')`).

Viicos:
Pleae review

pydantic_settings/sources.py
Viicos: I will take a look

hramezani (approved):

hramezani:
Thanks @Viicos

Issue #78, opened 2023-06-21
pydantic.Json[dict[str, str]] raise error


aleksey925:
### Initial Checks

- [X] I have searched GitHub for a duplicate issue and I'm sure this is something new
- [X] I have searched Google & StackOverflow for a solution and couldn't find anything
- [X] I have read and followed [the docs](https://docs.pydantic.dev) and still think this is a bug
- [X] I am confident that the issue is with pydantic (not my code, or another library in the ecosystem like [FastAPI](https://fastapi.tiangolo.com) or [mypy](https://mypy.readthedocs.io/en/stable))

### Description

The following code raise an error

```
Traceback (most recent call last):
  File "/Users/alexey.petrunnik/PycharmProjects/temp/temp6.py", line 12, in <module>
    s = Settings()
        ^^^^^^^^^^
  File "pydantic/env_settings.py", line 39, in pydantic.env_settings.BaseSettings.__init__
  File "pydantic/main.py", line 341, in pydantic.main.BaseModel.__init__
pydantic.error_wrappers.ValidationError: 1 validation error for Settings
data
  JSON object must be str, bytes or bytearray (type=type_error.json)
```

.env file
```
DATA='{"key":"value"}'
```

### Example Code

```Python
from pydantic import BaseSettings, Json


class Settings(BaseSettings):
    data: Json[dict[str, str]]

    class Config:
        env_file = '.env'
        env_file_encoding = 'utf-8'


s = Settings()
print(s.data)
```

### Python, Pydantic & OS Version

```Text
pydantic version: 1.10.9
            pydantic compiled: True
                 install path: /Users/alexey.petrunnik/Library/Application Support/pdm/venv/lib/python3.11/site-packages/pydantic
               python version: 3.11.2 (main, Feb 16 2023, 02:55:59) [Clang 14.0.0 (clang-1400.0.29.202)]
                     platform: macOS-13.4-arm64-arm-64bit
     optional deps. installed: ['dotenv', 'typing-extensions']
```

### Affected Components

- [ ] [Compatibility between releases](https://docs.pydantic.dev/changelog/)
- [X] [Data validation/parsing](https://docs.pydantic.dev/usage/models/#basic-model-usage)
- [ ] [Data serialization](https://docs.pydantic.dev/usage/exporting_models/) - `.dict()` and `.json()`
- [ ] [JSON Schema](https://docs.pydantic.dev/usage/schema/)
[...]

PR #79, opened 2023-06-21
Add Pydantic Json field support

hramezani:
Fixes https://github.com/pydantic/pydantic-settings/issues/78

### DIFF
--- a/pydantic_settings/sources.py
+++ b/pydantic_settings/sources.py
@@ -9,7 +9,7 @@ from dataclasses import is_dataclass
 from pathlib import Path
 from typing import TYPE_CHECKING, Any, List, Mapping, Sequence, Tuple, Union, cast
 
-from pydantic import AliasChoices, AliasPath, BaseModel
+from pydantic import AliasChoices, AliasPath, BaseModel, Json
 from pydantic._internal._typing_extra import origin_is_union
 from pydantic._internal._utils import deep_update, lenient_issubclass
 from pydantic.fields import FieldInfo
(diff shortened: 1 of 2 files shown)


hramezani:
please review

pydantic_settings/sources.py
```diff
@@ -602,7 +602,9 @@ def read_env_file(
         return file_vars
 
 
-def _annotation_is_complex(annotation: type[Any] | None) -> bool:
+def _annotation_is_complex(annotation: type[Any] | None, metadata: list[Any]) -> bool:
+    if any(str(md) == 'Json' for md in metadata):
```
hramezani, line 606: Not sure I am doing the right way to check if the field is a `Json` field.

dmontagu: Yeah this seems to be right:
```suggestion
    if any(isinstance(md, Json) for md in metadata):
```

(Also need to add `Json` to the set of imports from `pydantic` at the top)

hramezani: Thanks for your suggestion. Done

pydantic_settings/sources.py
```diff
@@ -602,7 +602,9 @@ def read_env_file(
         return file_vars
 
 
-def _annotation_is_complex(annotation: type[Any] | None) -> bool:
+def _annotation_is_complex(annotation: type[Any] | None, metadata: list[Any]) -> bool:
+    if any(isinstance(md, Json) for md in metadata):  # type: ignore[misc]
```
dmontagu, line 606:
```suggestion
    if any(isinstance(md, Json) for md in metadata):
```
Was the type ignore here necessary? I didn't think I needed it to run mypy with no errors but maybe I missed something. Just making sure

hramezani: It failed without type ignore
Here is the log: https://github.com/pydantic/pydantic-settings/actions/runs/5357133312/jobs/9717574129

dmontagu (approved):

PR #82, opened 2023-06-22
Add default value to source classes init params

hramezani:
This makes source classes inheritable in custom source classes. So, the custom source class doesn't need to override the `__init__` to provide a fallback value from the config.

Selected Reviewer: @dmontagu

### DIFF
--- a/pydantic_settings/sources.py
+++ b/pydantic_settings/sources.py
@@ -22,7 +22,12 @@ if TYPE_CHECKING:
 
 
 DotenvType = Union[Path, List[Path], Tuple[Path, ...]]
-env_file_sentinel: DotenvType = Path('')
+
+# This is used as default value for `_env_file` in `Settings` class and
+# `env_file` in `DotEnvSettingsSource`. Then passing `None` as `_env_file`
+# during initialization of `Settings` has a special meaning and it will
+# override the `env_file` specified in config.
+ENV_FILE_SENTINEL: DotenvType = Path('')
 
 
 class SettingsError(ValueError):
@@ -531,13 +536,13 @@ class DotEnvSettingsSource(EnvSettingsSource):
     def __init__(
         self,
         settings_cls: type[BaseSettings],
-        env_file: DotenvType | None = env_file_sentinel,
+        env_file: DotenvType | None = ENV_FILE_SENTINEL,
         env_file_encoding: str | None = None,
         case_sensitive: bool | None = None,
         env_prefix: str | None = None,
         env_nested_delimiter: str | None = None,
     ) -> None:
-        self.env_file = env_file if env_file != env_file_sentinel else settings_cls.model_config.get('env_file')
+        self.env_file = env_file if env_file != ENV_FILE_SENTINEL else settings_cls.model_config.get('env_file')
         self.env_file_encoding = (
             env_file_encoding if env_file_encoding is not None else settings_cls.model_config.get('env_file_encoding')
         )
--- a/pydantic_settings/sources.py
+++ b/pydantic_settings/sources.py
@@ -22,6 +22,7 @@ if TYPE_CHECKING:
 
 
 DotenvType = Union[Path, List[Path], Tuple[Path, ...]]
+env_file_sentinel: DotenvType = Path('')
 
 
 class SettingsError(ValueError):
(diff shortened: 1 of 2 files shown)


pydantic_settings/sources.py
```diff
@@ -22,6 +22,7 @@
 
 
 DotenvType = Union[Path, List[Path], Tuple[Path, ...]]
+env_file_sentinel: DotenvType = Path('')
```
dmontagu, line 25:
```suggestion
ENV_FILE_SENTINEL: DotenvType = Path('')
```

I generally prefer to have module-level constants be all-caps (unless they are type aliases, like the one just above this line, in which case PascalCase is fine).

Also, can we add documentation of how this sentinel works? I don't understand what the  meaning is vs. using `None` — I think it should be easy to determine that just from reading source code, and since I can't tell what it's doing by looking at the diff in this PR, I think it would be good to add a comment either in the docstring of `DotEnvSettingsSource.__init__` or right under this line defining `env_file_sentinel` explaining what the distinction is between providing the value `None` and the value `env_file_sentinel`.

pydantic_settings/sources.py
```diff
@@ -23,6 +23,9 @@
 
 DotenvType = Union[Path, List[Path], Tuple[Path, ...]]
 
+#
+ENV_FILE_SENTINEL: DotenvType = Path('')
```
dmontagu, line 27: You said you added some comments above this — but it looks like they haven't made their way in? I just see `#` lol

pydantic_settings/sources.py
```diff
@@ -23,6 +23,12 @@
 
 DotenvType = Union[Path, List[Path], Tuple[Path, ...]]
 
+# This is used as default value for `_env_file` in `Settings` class and
+# `env_file` in `DotEnvSettingsSource`. Then passing `None` as `_env_file`
+# during initialization of `Settings` has a special meaning and it will
+# override the `env_file` specified in config.
+ENV_FILE_SENTINEL: DotenvType = Path('')
```
dmontagu, line 30: Can we make this a bit more clear of the exact behavior so I know what will happen without needing to check the logic (right now it says "special meaning", but not what that special meaning is)? Here's an attempt — I haven't checked the logic closely, can you confirm if this is correct?

```suggestion
# This is used as default value for `_env_file` in `Settings` class and
# `env_file` in `DotEnvSettingsSource`. When this value is used, environment
# variables will be read from the `model_config`-specified file. Using this
# as the default instead of `None` allows us to use `None` to mean that environment
# variables should not be read from any env file.
ENV_FILE_SENTINEL: DotenvType = Path('')
```

Another approach we could take is putting this in the `pydantic_settings.main.BaseSettings.__init__` docstring, and then we could shorten/simplify the above comment:
    def __init__(
        __pydantic_self__,
        _case_sensitive: bool | None = None,
        _env_prefix: str | None = None,
        _env_file: DotenvType | None = env_file_sentinel,
        _env_file_encoding: str | None = None,
        _env_nested_delimiter: str | None = None,
        _secrets_dir: str | Path | None = None,
        **values: Any,
    ) -> None:
        """
        TODO: Properly document all additional kwargs here.

        When `_env_file == ENV_FILE_SENTINEL`, the `model_config`-specified env file will be used.
        When `_env_file is None`, environment variables will not be read from any env file.
        """
I do think we need to document all these keyword arguments eventually, but even if you don't want to do that now, I think it's okay to put a placeholder docstring in like this and it would address my concern.

dmontagu: Oh I just realized that this is documented in the `pydantic_settings.main.BaseSettings` docstring. I think we can just augment that docstring as part of this PR if you don't mind. I would change the current value:
```
        _env_file: The env file(s) to load settings values from. Defaults to `Path('')`.
```
to
```
        _env_file: The env file(s) to load settings values from. Defaults to `Path('')`, which
        means that the value from `model_config['env_file']` should be used. You can also pass
        `None` to indicate that environment variables should not be loaded from an env file.
```
or similar

dmontagu: If we change the `BaseSettings` docstring, I think we can change the comment here to just 
```suggestion
# This is used as default value for `_env_file` in the `BaseSettings` class and
# `env_file` in `DotEnvSettingsSource` so that it can be distinguished from `None`.
# See the docstring of `BaseSettings` for more details.
ENV_FILE_SENTINEL: DotenvType = Path('')
```

dmontagu (approved):

PR #86, opened 2023-06-26
Fix formatting of index.md

Lithimlin:
This change should hopefully fix some weird code formatting that are happening on the website version of the documentation.

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -318,6 +318,7 @@ Once you have your `.env` file filled with variables, *pydantic* supports loadin
 in the `BaseSettings` class:
 
 ```py test="skip" lint="skip"
+
 class Settings(BaseSettings):
     model_config = SettingsConfigDict(env_file='.env', env_file_encoding='utf-8')
 ```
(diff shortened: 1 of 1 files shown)


samuelcolvin:
have you tried this and checked if it works?

Lithimlin:
Truthfully, no. I forgot that that was actually a thing I could also do locally.

Lithimlin:
I'm experiencing some trouble getting the page to build at all with mkDocs.
I'm currently running into the following issue after manually installing `mike` and `mkdocs-exclude`:

```
$ mkdocs serve
WARNING  -  Config value 'plugins': Plugin 'mike' option 'alias_type': Unrecognised configuration name: alias_type
Aborted with 1 Configuration Warnings in 'strict' mode!
```

hramezani:
Thanks @Lithimlin for this patch.

I've fixed the problem in https://github.com/pydantic/pydantic-settings/pull/89 by making the code block testable and adding some highlighting to the code block

PR #87, opened 2023-06-26
Fix garbled up docs

fredrikaverpil:
### Why this change?

The docs are garbled up in a few places.

Screenshot 2023-06-26 at 08 26 44

### What was done?

- Removed the occurrences of `test="skip" lint="skip"` which followed the three backticks to mark a code block.

### Concerns, notes, side-effects etc

- I'm not sure if instead there was some missing plugin or something that should instead be added rather than removing these declarations. However, I couldn't find these declarations elsewhere in the docs, so I figured I'd just remove them.


fredrikaverpil:
😅 okay so it seems all tests failed with this change.

Issue #88, opened 2023-06-26
Badly formatted code blocks garble up docs


fredrikaverpil:
![image](https://github.com/pydantic/pydantic-settings/assets/994357/810205e6-4076-4184-a3d0-9f927955c0cd)

This gets repeated whenever you have three backticks followed by `test="skip" lint="skip"`. I tried removing them in #87 but then all CI test checks [failed](https://github.com/pydantic/pydantic-settings/actions/runs/5374619268?pr=87). 😅

fredrikaverpil:
It's possible that there's an issue with pytest-examples where it doesn't work well with the main pydantic docs including the markdown file from this repo.

hramezani:
Thanks @fredrikaverpil for reporting this.

It got fixes in https://github.com/pydantic/pydantic-settings/commit/584a57b1c0b9833b130af59f4801ed5c093e04b0

fredrikaverpil:
@hramezani out of curiosity, since you removed the additional arguments in your first commit, and I did too (but had failed tests), did you do something else?

hramezani:
@fredrikaverpil I made the tests with `test="skip"` testable. for example, the first test was:

```py
class Settings(BaseSettings):
    model_config = SettingsConfigDict(env_file='.env', env_file_encoding='utf-8')
```

It needed `BaseSettings` and `SettingsConfigDict` to be imported. So, I imprted them by:

`from pydantic_settings import BaseSettings, SettingsConfigDict`

PR #89, opened 2023-06-26
Fix code block formatting

hramezani:
Fixes https://github.com/pydantic/pydantic-settings/issues/88


Kludex (approved):
I did what I did, so there was less text on the snippet.

Should we highlight the lines? https://squidfunk.github.io/mkdocs-material/reference/code-blocks/#highlighting-specific-lines

In any case, I'm fine with this.

hramezani:
Thanks @Kludex for your suggestion.

Highlight added to some lines.

Issue #95, opened 2023-07-02
Mypy issues with BaseSettings and SettingsConfigDict - v2


axeng:
It seems that mypy has some issues with the new BaseSettings.
I did the same tests with pydantic v1 and it did not throw these errors.

Using this example:
```python
from pydantic_settings import BaseSettings, SettingsConfigDict

class TestSettings(BaseSettings):
    model_config = SettingsConfigDict(env_file=".env")

    test_field: str

settings = TestSettings()
```

Mypy is throwing the following:
```shell
test.py:4: error: Incompatible types (expression has type "str", TypedDict item "env_file" has type "Path | list[Path] | tuple[Path, ...] | None")  [typeddict-item]
test.py:8: error: Missing named argument "test_field" for "TestSettings"  [call-arg]
Found 2 errors in 1 file (checked 1 source file)
```

Here the Mypy configuration I am using:
```toml
[tool.mypy]
ignore_missing_imports = true
plugins = [
    "pydantic.mypy",
]
disallow_untyped_defs = true
disallow_any_unimported = true
no_implicit_optional = true
check_untyped_defs = true
warn_unused_ignores = true
show_error_codes = true
warn_unreachable = true
```

----------

`pydantic` version: 2.0
`pydantic-core` version: 2.0.1
`pydantic-settings` version: 2.0
`mypy` version: 1.4.1

Selected Assignee: @dmontagu

hramezani:
Thanks @axeng for reporting this issue.

The first error:
> test.py:4: error: Incompatible types (expression has type "str", TypedDict item "env_file" has type "Path | list[Path] | tuple[Path, ...] | None")  [typeddict-item]

Can be fixed by including `str` in the allowed type for `env_file` config.

The second error:
> test.py:8: error: Missing named argument "test_field" for "TestSettings"  [call-arg]

There was [some logic related to setting management in V1 mypy plugin](https://github.com/pydantic/pydantic/blob/0e8a387d6752dd859676da53298601ccae6a23a7/pydantic/mypy.py#L315) which has been removed because settings management is not part of `Pydantic V2`.
I think we can create a mypy plugin for `pydantic-settings` that inherit from `Pydantic V2` mypy plugin and add specific logic for `pydantic-settings`.

What do you think @samuelcolvin @dmontagu ?

PR #96, opened 2023-07-03
Improve `DotenvType` to accept `str`

hramezani:
Fixes first error reported in https://github.com/pydantic/pydantic-settings/issues/95

```
test.py:4: error: Incompatible types (expression has type "str", TypedDict item "env_file" has type "Path | list[Path] | tuple[Path, ...] | None")  [typeddict-item]
```

### DIFF
--- a/pydantic_settings/sources.py
+++ b/pydantic_settings/sources.py
@@ -21,7 +21,7 @@ if TYPE_CHECKING:
     from pydantic_settings.main import BaseSettings
 
 
-DotenvType = Union[Path, List[Path], Tuple[Path, ...]]
+DotenvType = Union[Path, str, List[Union[Path, str]], Tuple[Union[Path, str], ...]]
 
 # This is used as default value for `_env_file` in the `BaseSettings` class and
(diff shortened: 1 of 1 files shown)


hramezani:
please review

Kludex (approved):
We don't test these?

hramezani:
> We don't test these?

We have some tests that include the string `env_file` in the config.

Which kind of tests do you mean?

Kludex:
> > We don't test these?
> 
> We have some tests that include the string `env_file` in the config.
> 
> Which kind of tests do you mean?

Any test that would avoid a regression here. 👀

hramezani:
> > > We don't test these?
> > 
> > 
> > We have some tests that include the string `env_file` in the config.
> > Which kind of tests do you mean?
> 
> Any test that would avoid a regression here. 👀

It's a mypy error and we don't have mypy tests here.


> test.py:4: error: Incompatible types (expression has type "str", TypedDict item "env_file" has type "Path | list[Path] | tuple[Path, ...] | None")  [typeddict-item]

PR #97, opened 2023-07-03
Add explicit type hint to `BaseSettings.model_config`

Viicos:
Noticed this small issue when working inside `settings_customise_sources`. Before this PR (by using either `cls` or `settings_cls`):

image

After:

image

What's weird is that my IDE seems to infer the type correctly in other places, even without this improvement:

image

By the way, is the `settings_cls` argument necessary, as `settings_customise_sources` is already a class method and we have access to `cls`?

Selected Reviewer: @dmontagu

### DIFF
--- a/pydantic_settings/main.py
+++ b/pydantic_settings/main.py
@@ -1,7 +1,7 @@
 from __future__ import annotations as _annotations
 
 from pathlib import Path
-from typing import Any
+from typing import Any, ClassVar
 
 from pydantic import ConfigDict
 from pydantic._internal._utils import deep_update
@@ -153,7 +153,7 @@ class BaseSettings(BaseModel):
             # to an informative error and much better than a confusing error
             return {}
 
-    model_config = SettingsConfigDict(
+    model_config: ClassVar[SettingsConfigDict] = SettingsConfigDict(
         extra='forbid',
         arbitrary_types_allowed=True,
         validate_default=True,


Viicos:
I think the `mypy` issue comes from the fact that `model_config` is defined in [two places](https://github.com/pydantic/pydantic/blob/0ed48ae141e2a50694bc57ee356bdac13c4af406/pydantic/main.py#L92-L138) on `BaseModel`:

```python
    if typing.TYPE_CHECKING:
        # Here we provide annotations for the attributes of BaseModel.
        # Many of these are populated by the metaclass, which is why this section is in a `TYPE_CHECKING` block.
        # However, for the sake of easy review, we have included type annotations of all class and instance attributes
        # of `BaseModel` here:

        # Class attributes
        model_config: ClassVar[ConfigDict]
```

And later on:

```python
    __slots__ = '__dict__', '__pydantic_fields_set__', '__pydantic_extra__', '__pydantic_private__'

    model_config = ConfigDict()
    __pydantic_complete__ = False
    __pydantic_root_model__ = False
```

Would it be safe to remove the first one in the `if TYPE_CHECKING` block, and modify the remaining one to `model_config: ClassVar[ConfigDict] = ConfigDict()`?

hramezani:
Thanks @Viicos for this patch.

Yeah, seems the mypy error will be fixed by the change you mentioned on `pydantic` side.

You have to create a PR on pydantic and suggest the change. We(other people on the team and I) will review the PR. they may have different opinions on that.

davidhewitt:
I noticed the lint is using pydantic 2.0b3 and since beta 3 we had https://github.com/pydantic/pydantic/pull/6260 which added the type annotation to `model_config` which is proposed to be changed in https://github.com/pydantic/pydantic/pull/6393

Perhaps there is no change needed in pydantic and the mypy issue will be resolved by updating the linting requirements to 2.0? We could merge #99 and then rebase this PR to confirm.

Viicos:
@davidhewitt Good catch, I will rebase once #99 gets merged. It seems to work with a MRE:

https://mypy-play.net/?mypy=latest&python=3.11&gist=ba7caebef09f0b3aa5397d05dfef4541

(the error appears if you remove the `if TYPE_CHECKING` block).

hramezani:
@Viicos As #99 has been merged, Could you please rebase this PR on top of `main`?

Viicos:
Rebasing fixed it.

Pease review

hramezani (approved):

hramezani:
Thanks @Viicos

PR #107, opened 2023-07-05
dotenv no error w/ an env key shorter than prefix (#106)

### LINKED ISSUE #106: DotEnv ValueError when env variable name is shorter than env_prefix
Example:

```
class DramatiqSettings(BaseSettings):
    broker_url: RedisDsn

        env_file=".env", env_prefix="plato_worker_", extra="ignore"  # type: ignore
    )
```
I also have another settings class that holds my general application settings.

When processing my PLATO_DEBUG env var, the following line fails:
https://github.com/pydantic/pydantic-settings/blob/373af71ae8ac0d2c3530857ce51ef28c9e9178d2/pydantic_settings/sources.py#L583

```
ValueError: not enough values to unpack (expected at least 1, got 0)
```

It's because PLATO_DEBUG is shorter than the env_prefix for my DramatiqSettings class.

Possible Solution:
Check if the env var name actually starts with the env prefix before stripping the name.

Selected Assignee: @samuelcolvin

jvllmr:
fixes #106 

I created a test that tests a dotenv config class with a prefix to test the effect of my changes

### DIFF
--- a/pydantic_settings/sources.py
+++ b/pydantic_settings/sources.py
@@ -578,7 +578,7 @@ class DotEnvSettingsSource(EnvSettingsSource):
         # As `extra` config is allowed in dotenv settings source, We have to
         # update data with extra env variabels from dotenv file.
         for env_name, env_value in self.env_vars.items():
-            if env_value is not None:
+            if env_name.startswith(self.env_prefix) and env_value is not None:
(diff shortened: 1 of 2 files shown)


hramezani (approved):

hramezani:
Thanks @jvllmr

Issue #110, opened 2023-07-06
Support setting model through class kwargs


wookie184:
`pydantic.BaseModel` supports setting config through class kwargs:
```python
from pydantic import BaseModel
class Foo(BaseModel, extra='forbid'):
    pass
```

It would be nice for `pydantic_settings.BaseSettings` to also support this:
```python
from pydantic_settings import BaseSettings
class Foo(BaseSettings, env_file='.env'):
    pass
```
Currently an error is raised:
```python
Traceback (most recent call last):
  File "...\test.py", line 2, in <module>
    class Foo(BaseSettings, env_file='.env'):
  File "...\Lib\site-packages\pydantic\_internal\_model_construction.py", line 124, in __new__
    cls: type[BaseModel] = super().__new__(mcs, cls_name, bases, namespace, **kwargs)  # type: ignore
                           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "<frozen abc>", line 106, in __new__
TypeError: Foo.__init_subclass__() takes no keyword arguments
```

This pydantic issue may be relevant: https://github.com/pydantic/pydantic/issues/6024

Thanks

Selected Assignee: @samuelcolvin

hramezani:
Thanks @wookie184 for this issue 🙏 

I've created a PR to support this.

hramezani:
Support for setting `pydantic-settings` configs throgh kwargs added.
It will be available in next release.

PR #112, opened 2023-07-07
Add support for adding config through class kwargs

hramezani:
Closes https://github.com/pydantic/pydantic-settings/issues/110

Selected Reviewer: @dmontagu

### DIFF
--- a/pydantic_settings/main.py
+++ b/pydantic_settings/main.py
@@ -28,8 +28,12 @@ class SettingsConfigDict(ConfigDict, total=False):
     secrets_dir: str | Path | None
 
 
-# Extend config_keys by pydantic settings config keys to
+# Extend `config_keys` by pydantic settings config keys to
 # support setting config through class kwargs.
+# Pydantic uses `config_keys` in `pydantic._internal._config.ConfigWrapper.for_model`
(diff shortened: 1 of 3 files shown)


hramezani:
please review

pydantic_settings/main.py
```diff
@@ -27,6 +28,11 @@ class SettingsConfigDict(ConfigDict, total=False):
     secrets_dir: str | Path | None
 
 
+# Extend config_keys by pydantic settings config keys to
+# support setting config through class kwargs.
+config_keys |= set(SettingsConfigDict.__annotations__.keys())
```
Kludex, line 33: This seems super magical... Can you add a reference to the pydantic code that we are using here?

Kludex (changes requested):

pydantic_settings/main.py
hramezani: I improved the comment and explained it more by mentioning the class and method name that the magic happen there on Pydantic.

Please let me know by comment or suggestion if you think it is not clear yet.

Kludex (approved):

jordantshaw:
Which release is this going with? @hramezani

hramezani:
> Which release is this going with? @hramezani

Next release will happen soon and this will be included

Issue #139, opened 2023-08-12
Support `keyring` encrypted credential storage


lmmx:
## Outline

- Setting env. variables for auth config is awkward, so `python-dotenv` can be used to load them from a `.env` text file instead
- Keyrings can be preferable to plain text secret storage for security, and Python has a `keyring` module [[source]](https://keyring.readthedocs.io/en/latest/index.html)

  > These recommended keyring backends are supported:
  >
  > -  macOS Keychain
  > -  Freedesktop Secret Service supports many DE including GNOME (requires secretstorage)
  > -  KDE4 & KDE5 KWallet (requires dbus)
  > -  Windows Credential Locker

- This practice is used by other well known tools such as the GitHub CLI tool `gh` and `twine` [[source]](https://twine.readthedocs.io/en/stable/contributing.html#where-twine-gets-configuration-and-credentials)

  > **Where Twine gets configuration and credentials**
  > A user can set the repository URL, username, and/or password via command line, `.pypirc` files, environment variables, and keyring.

- This could be made an optional dependency if desired (perhaps to keep package size minimal/consistent).

## Impact

When I `pip install keyring` on Linux after first installing `pydantic` and `pydantic-settings` the additional dependencies are:

```
Installing collected packages: zipp, pycparser, more-itertools, jeepney, jaraco.classes, importlib-metadata, cffi, cryptography, SecretStorage, keyring
```

## Usage

Once installed, secret access is achieved like so:

```py
import keyring
my_secret = keyring.get_password("MY_SECRET", "secret_username")
```

The `gh` tool sets the username to an empty string, indicating that it's used as a simple key-value secret store.

You can also access specific keyrings, also known as 'collections' (for instance if you wanted to have different applications using different keys with the same name, say a different API key for different services). [For reference](https://github.com/jaraco/keyring/issues/151#issuecomment-1111685038)

## Proposed implementation

Essentially we are replacing `os.environ.get(validation_alias)` for `keyring.get_password(validation_alias)`

In this library, both environment variables and `.env` configured variables are loaded into the `env_vars` attribute.

`EnvSettingsSource` calls `_load_env_vars()` at initialisation:

https://github.com/pydantic/pydantic-settings/blob/5933ea6ca51d58342378eed5db12d3f6f6dee8df/pydantic_settings/sources.py#L376-L381

 `DotEnvSettingsSource` subclasses `EnvSettingsSource` and overrides the `_load_env_vars()` method

[...]

PR #140, opened 2023-08-12
feat: contribute KeyringSettingsSource class to load keyring variables

lmmx:
## Proposal

PR to fulfill the feature proposal:

- #139

## Details

- A `keyring_backend` is provided, and behaves similarly to the `env_file`: if the named backend is not found, nothing is loaded. If not specified, the default backend is used.

## Request for review

- Is the approach of `try`/`except ImportError: pass` acceptable? An alternative would be to set a variable that could be used to gate the calling of the `keyring` module.
- How should I implement error handling for the access of the variables? (As mentioned in the original proposal). If the keyring is locked it will raise an error. I haven't added this handling to keep the original proposal simple.

### DIFF
--- a/pydantic_settings/__init__.py
+++ b/pydantic_settings/__init__.py
@@ -3,6 +3,7 @@ from .sources import (
     DotEnvSettingsSource,
     EnvSettingsSource,
     InitSettingsSource,
+    KeyringSettingsSource,
     PydanticBaseSettingsSource,
     SecretsSettingsSource,
 )
@@ -13,6 +14,7 @@ __all__ = (
     'DotEnvSettingsSource',
     'EnvSettingsSource',
     'InitSettingsSource',
+    'KeyringSettingsSource',
     'PydanticBaseSettingsSource',
     'SecretsSettingsSource',
     'SettingsConfigDict',
--- a/pydantic_settings/main.py
+++ b/pydantic_settings/main.py
@@ -14,6 +14,7 @@ from .sources import (
     DotenvType,
     EnvSettingsSource,
     InitSettingsSource,
+    KeyringSettingsSource,
     PydanticBaseSettingsSource,
     SecretsSettingsSource,
 )
@@ -25,6 +26,7 @@ class SettingsConfigDict(ConfigDict, total=False):
     env_file: DotenvType | None
     env_file_encoding: str | None
     env_nested_delimiter: str | None
+    keyring_backend: str | None
     secrets_dir: str | Path | None
 
 
@@ -54,6 +56,7 @@ class BaseSettings(BaseModel):
             `None` to indicate that environment variables should not be loaded from an env file.
         _env_file_encoding: The env file encoding, e.g. `'latin-1'`. Defaults to `None`.
         _env_nested_delimiter: The nested env values delimiter. Defaults to `None`.
+        _keyring_backend: The keyring backend, e.g. `'SecretService Keyring'`. Defaults to `None`.
         _secrets_dir: The secret files directory. Defaults to `None`.
     """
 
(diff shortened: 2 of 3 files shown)


lmmx:
# TODO list

- [ ] pip compile
- [ ] Optional dependency import
- [ ] Settings source edit
- [ ] raise `SettingsError` upon keyring variable access error
- [x] Linux support
- [ ] Mac OSX support
- [ ] Tests
- [ ] Documentation

## pip compile

> You can add `keyring` to the [linting requirements](https://github.com/pydantic/pydantic-settings/blob/main/requirements/linting.in) and then run `pip-compile --output-file=requirements/linting.txt requirements/linting.in`

## Optional dependency import

Ah yes I see:

- `TYPE_CHECKING` conditional block that imports the optional dependency else sets it to None

https://github.com/pydantic/pydantic/blob/d70abd88b4ae70431d534ce424788eb4494928e1/pydantic/networks.py#L16-L22

- Helper function called `import_email_validator()` that may raise `ImportError`, followed by differential declaration of a class that relies on it dependent on the `TYPE_CHECKING` context (if so, just declaring as `Annotated[str, ...]`).

https://github.com/pydantic/pydantic/blob/d70abd88b4ae70431d534ce424788eb4494928e1/pydantic/networks.py#L161-L194

## Settings source edit

>  Also, I think we need to include `KeyringSettingsSource` in sources if `keyring` is installed.

I hadn't thought of that, I just added it and let it default to `None`, I'll review...

## raise `SettingsError` upon keyring variable access error

:+1: 

## Linux support

Implemented via `secretstorage` as described

## Mac support

OS X has a `security` module which `keyring` can access and a `dump-keychain` method which would provide equivalent functionality. I need to look at this in more detail.

## Tests

I used this as a dependency in an early stage project `magic-scrape` [here](https://github.com/lmmx/magic-scrape) (dependency declared [here](https://github.com/lmmx/magic-scrape/blob/5817004abc7e8c78eaab0a42f553f55b826d0fc8/pyproject.toml#L25)) and by patching the `pydantic_settings.KeyringSettingsSource.__call__` method (equivalently I could also have patched the `_read_keyring` method) I have been able to test the integration of this with my package.

This already helped me uncover and fix bugs in the implementation (in case sensitivity handling), so testing should be easy to build off of this.

## Documentation

:+1:

lmmx:
I thought about the way pydantic is importing optional dependencies and as I recalled seeing a cleaner way in pandas, I reviewed their approach, `import_optional_dependency`. [source](https://github.com/pandas-dev/pandas/blob/e92ea577ccce3ae8cc7da2baba44d72854f4947f/pandas/compat/_optional.py#L92) -> [example usage](https://github.com/pandas-dev/pandas/blob/e92ea577ccce3ae8cc7da2baba44d72854f4947f/pandas/io/formats/style_render.py#L48)

The rest of the comment explains why I'm choosing a combination of both, if it's of interest!

Theirs is not type equivalent: in pydantic the TYPE_CHECKING block ensures the type of the module name is always an imported module's type, whereas in pandas the return type of the helper function is `Optional`. This is only permissible there because they use `--skip-unannotated` and said helper function has no annotated return type.

Their is however cleaner: the import site is not required to be wrapped in `if TYPE_CHECKING`, and indeed their `_optional.py` module has scaled well to handle many optional deps.

pydantic's approach uses `global` to redefine the module which was imported as `None` in the else condition of the `if TYPE_CHECKING` block used to spoof the module type as if it were a non-Optional dependency.

We can parameterise `globals()[package_name]` instead and put the `TYPE_CHECKING` condition within the helper function, thereby achieving the same outcome (spoofed always-module that's actually `Optional`) but without the overhead at site of use. I expect this would look cleaner and hope this would therefore be acceptable in my PR.

I'll put it together and fall back to the template of function if I've misjudged part of this or if you don't like it.

cisaacstern:
May I ask what the status of this PR is?

hramezani:
Replaces by https://github.com/pydantic/pydantic-settings/pull/729

PR #150, opened 2023-08-23
python-dotenv should be optional

### LINKED ISSUE #149: python-dotenv should be optional
From `pyproject.toml`:
```
requires-python = '>=3.7'
dependencies = [
    'pydantic>=2.0.1',
    'python-dotenv>=0.21.0',
]
dynamic = ['version']
```
`python-dotenv` is always installed.

From `pydantic_settings.sources`:
https://github.com/pydantic/pydantic-settings/blob/919a20b77527ecc1cd6eeb0a09ca22cc21486fb8/pydantic_settings/sources.py#L620-L632
- here is the logic of the optional `python-dotenv`
- and the wrong message in exception to install as pydantic extra (which already does not have this extra)

---

One more place where the project depends on `python-dotenv` is `tests.test_settings`:
https://github.com/pydantic/pydantic-settings/blob/919a20b77527ecc1cd6eeb0a09ca22cc21486fb8/tests/test_settings.py#L39-L42
~~But even here checks are made to skip tests if `python-dotenv` is not installed (one of the tests with such a check):~~
https://github.com/pydantic/pydantic-settings/blob/919a20b77527ecc1cd6eeb0a09ca22cc21486fb8/tests/test_settings.py#L663-L680
While making PR I understood the reason for making it in tests.
Finally, I changed the CI workflow.

Selected Assignee: @dmontagu

maxzhenzhera:
Closes #149


hramezani:
@maxzhenzhera Thanks for this PR 🙏 

Based on [Samuel's comment on the issue](https://github.com/pydantic/pydantic-settings/issues/149#issuecomment-1691938797), we want to keep `python-dotenv` as a dependency.

So, I close this PR. Feel free to create another PR if you want to remove extra codes for handling optional dependency

PR #152, opened 2023-08-26
Env var extraction to classmethod

vindex10:
Hi all! Thank you for a great project :) Its irreplaceability was motivating me all the way to opening this PR. I understand that my explanation may be hard to digest, please ask questions and make comments, I'm really motivated to make this change come true :)

Big goal: be able to nest settings while preserving encapsulation of the nested setting classes. In other words, to be able to re-use the settings file of my script when I plug the script to a bigger project. The only change - env_prefix.

Motivation: nested env variables were included a bit of ad-hoc, making parent Settings class majorly responsible for parsing nested fields. I understand, a part of the reason is that parent class is instantiated, while nested classes are not, but the extraction logic was tightly bound to instance methods/variables.

Change: make extraction logic (i.e. `__call__` method) a classmethod `read_model_fields`. Example

```
res = {}
for field in settings.fields:
    env_prefix = env_prefix_from_field(field)
    res[field.name] = field.annotation.model.read_fields(env_prefix=env_prefix)
```

This naturally would allow using a python script with PydanticSettings, and then seamlessly plug it into a bigger code without modification of the script and its config themselves (even if the script used env_prefix, or included some other nested settings, we dont care anymore).

---

In this PR I tried to make a first step towards the big goal. It is still quite large, that's why I'm explaining the motivation as best as I can.

Some things I though of:

1. PydanticBaseSettingsSource - I left unchanged, although most of the methods became `@classmethods` in the child sources.
2. I worked mainly inside  `PydanticBaseEnvSettingsSource` and children.
3. `PydanticBaseEnvSettingsSource` supports `read_model_fields` which is a classmethod and its output is equivalent to output of `__call__` for each child source. Most of the changes in the code were caused to make this possible.
4. New methods are equivalents of the corresponding instance methods
    * prepare_field_value_from_env_vars
    * get_field_value_from_env_vars
    now to create a custom source, one has to implement these for BaseEnvSource. Instance methods are not used for extraction there.
5. I introduced SourceConfig for each source, to be able to pass the config to classmethods. Now it is a bit odd mixture of private classes / methods which could possibly be moved outside.
6. I preserved `load_env_vars` instance methods, and I use them in the root settings class, which is instantiated, to read in advance all the variables (either from secret dir, dotenv or env vars). Then the goal is to pass subsets of the "loaded env vars" when extraction of variables from the nested fields is needed.

---

Nice side-effects:

* Now all EnvSettings support aliases

Thank you very much for your attention!

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -64,7 +64,7 @@ class Settings(BaseSettings):
     # export my_prefix_more_settings='{"foo": "x", "apple": 1}'
     more_settings: SubModel = SubModel()
 
-    model_config = SettingsConfigDict(env_prefix='my_prefix_')  # (5)!
+    model_config = SettingsConfigDict(env_prefix='my_prefix_', extra='ignore')  # (5)!
 
 
 print(Settings().model_dump())
@@ -273,7 +273,7 @@ You may also populate a complex type by providing your own source class.
 ```py
 import json
 import os
-from typing import Any, List, Tuple, Type
+from typing import Any, List, Mapping, Tuple, Type
 
 from pydantic.fields import FieldInfo
 
@@ -285,8 +285,15 @@ from pydantic_settings import (
 
 
 class MyCustomSource(EnvSettingsSource):
-    def prepare_field_value(
-        self, field_name: str, field: FieldInfo, value: Any, value_is_complex: bool
+    @classmethod
+    def prepare_field_value_from_env_vars(
+        cls,
+        field_name: str,
+        field: FieldInfo,
+        value: Any,
+        value_is_complex: bool,
+        env_vars: Mapping[str, str | None],
+        config: EnvSettingsSource._EnvSettingsSource__SourceConfig,
     ) -> Any:
         if field_name == 'numbers':
             return [int(x) for x in value.split(',')]
--- a/pydantic_settings/sources.py
+++ b/pydantic_settings/sources.py
@@ -7,9 +7,9 @@ from abc import ABC, abstractmethod
 from collections import deque
 from dataclasses import is_dataclass
 from pathlib import Path
-from typing import TYPE_CHECKING, Any, List, Mapping, Sequence, Tuple, Union, cast
+from typing import TYPE_CHECKING, Any, List, Mapping, Sequence, Tuple, Union, cast, Literal
 
-from pydantic import AliasChoices, AliasPath, BaseModel, Json
+from pydantic import AliasChoices, AliasPath, BaseModel, Json, ConfigDict
 from pydantic._internal._typing_extra import origin_is_union
 from pydantic._internal._utils import deep_update, lenient_issubclass
 from pydantic.fields import FieldInfo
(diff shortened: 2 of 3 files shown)


hramezani:
Thanks @vindex10 for this patch 🙏 

Unfortunately, the patch is very big and hard to follow.

on the other hand, the main problem that you are going to address is still not clear to me.

> Big goal: be able to nest settings while preserving encapsulation of the nested setting classes. In other words, to be able to re-use the settings file of my script when I plug the script to a bigger project. The only change - env_prefix.

Could you please provide a minimal example, test, or whatever you can to explain the problem?

vindex10:
Hi @hramezani  :) Thanks for having a look! I'm sharing several examples below.

Imagine a small script "sub" with a settings class AppSub:

```py
import os


class AppSub(BaseSettings):
    xxx: str = 'bar'
    yyy: int = 1

    model_config = SettingsConfigDict(env_prefix='sub_')


os.environ["sub_xxx"] = "hello"
os.environ["sub_yyy"] = "2"


appsub = AppSub()
assert appsub.xxx == "hello"
assert appsub.yyy == 2
```

This script now has a well defined set of parameters: `sub_xxx` and `sub_yyy`.

Then, there appears a bigger project "top", which wants to use "sub" as utility:

```py
import os


class AppSub(BaseSettings):
    xxx: str = 'bar'
    yyy: int = 1

    model_config = SettingsConfigDict(env_prefix='sub_')

class AppTop(BaseSettings):
    aaa: str | None = None
    bbb: AppSub = AppSub()

    model_config = SettingsConfigDict(env_prefix='top_', env_nested_delimiter="__")

# <some env definition>

apptop = AppTop()

assert apptop.aaa == "world"
assert apptop.bbb.xxx == "hello"
assert apptop.bbb.yyy == 2
```

With current version of pydantic-settings, these env vars should be used:

```py
os.environ["top_aaa"] = "world"
os.environ["top_bbb__xxx"] = "hello"
os.environ["top_bbb__yyy"] = "2"
```

which I found counterintuitive. I know that my child model has a set of parameters I have to provide (`sub_xxx` and `sub_yyy`), so I would expect this to work:

```py
os.environ["top_aaa"] = "world"
os.environ["top_bbb__sub_xxx"] = "hello"
os.environ["top_bbb__sub_yyy"] = "2"
```

but it doesnt. Because not only AppTop fields are read by AppTop-class, but also all the nested ones. In my opinion it would be more natural to let AppSub read its own fields.

---
How this could be implemented in the code.

Currently, the read fields are returned from the `__call__` method. Which also calls `explode_env_vars` to drill down the nested fields.

What if we could call `__call__` of the nested model inside`explode_env_vars` to take care of its own fields?

`__call__` is an instance method, but we don't have instantiated `AppSub` yet.

What if  `__call__` was a classmethod? I didn't try, may be an option. I created a classmethod `read_model_fields` instead, which is called from `__call__`. Then, assuming `explode_env_vars` magically became a classmethod itself, we can call AppSub.read_model_fields from inside AppTop.__call__. 

No classes were instantiated here :) 

---

One more example, includes nested fields of nested fields. Let's assume a small script includes a nested model:

```py
import os


class AppSubSub(BaseSettings):
    mmm: str = 'bar'

    model_config = SettingsConfigDict(env_prefix='', env_nested_delimiter='')


class AppSub(BaseSettings):
    xxx: str = 'bar'
    yyy: int = 1
    sub: AppSubSub = AppSubSub()

    model_config = SettingsConfigDict(env_prefix='sub_', env_nested_delimiter='_')

os.environ["sub_xxx"] = "hello"
os.environ["sub_yyy"] = "2"
os.environ["sub_sub_mmm"] = "world"

app = AppSub()

assert app.xxx == "hello"
assert app.yyy == 2
assert app.sub.mmm == "world"
```

Now we want to plug it into a bigger project, as before:

```py
import os


class AppSubSub(BaseSettings):
    mmm: str = 'bar'

    model_config = SettingsConfigDict(env_prefix='', env_nested_delimiter='')


class AppSub(BaseSettings):
    xxx: str = 'bar'
    yyy: int = 1
    sub: AppSubSub = AppSubSub()

    model_config = SettingsConfigDict(env_prefix='sub_', env_nested_delimiter='_')

class AppTop(BaseSettings):
    aaa: str | None = None
    bbb: AppSub = AppSub()

    model_config = SettingsConfigDict(env_prefix='top_', env_nested_delimiter="__")

# <define some env vars>

app = AppTop()

assert app.aaa == "world"
assert app.bbb.xxx == "hello"
assert app.bbb.yyy == 2
assert app.bbb.sub.mmm == "world"
```

Currently, this is the way to satisfy asserts:

```py
os.environ["top_aaa"] = "world"
os.environ["top_bbb__xxx"] = "hello"
os.environ["top_bbb__yyy"] = "2"
os.environ["top_bbb__sub__mmm"] = "world"
```

Notice how `sub_mmm` became `sub__mmm`, because nested env delimiter got inherited from AppTop.

This is how I would expect it to be:

```py
os.environ["top_aaa"] = "world"
os.environ["top_bbb__sub_xxx"] = "hello"
os.environ["top_bbb__sub_yyy"] = "2"
os.environ["top_bbb__sub_sub_mmm"] = "world"
```

Because, again, the child "app" has a well defined set of config params, would be nice to preserve their format.

hramezani:
Thanks @vindex10 for the complete description and great examples.

I think the current approach is a desired approach and we don't want to change it because it would be a breaking change and most of Pydantic users already depend on this behavior.

BTW, I would like to see what @samuelcolvin thinks about it

PR #158, opened 2023-08-31
chore(dev-dependencies): use dev dependency and not requirements

sylvainmouquet:
Hello,

I propose to replace the usage of requirements by dev-dependencies.
The fact to stay in the context of the current venv is a better practice to close the dependencies used by the project in a jail

Sylvain


hramezani:
Thanks @sylvainmouquet for this patch 🙏 

I am not sure that we want to use `pdm` here in `pydantic-settings` because we are happy with the current setup and it suitable for our needs.

But, Let's see what @samuelcolvin think.

PR #160, opened 2023-09-04
Fix broken links

hramezani:


docs/index.md
```diff
@@ -85,17 +85,17 @@ print(Settings().model_dump())
 1. The environment variable name is overridden using `validation_alias`. In this case, the environment variable
    `my_auth_key` will be read instead of `auth_key`.
 
-    Check the [`Field` documentation](../fields/) for more information.
+    Check the [`Field` documentation](fields.md) for more information.
```
samuelcolvin, line 88:
```suggestion
    Check the [`Field` documentation](../usage/fields.md) for more information.
```
I think is right, but please check.

And below.

samuelcolvin (changes requested):

docs/index.md
hramezani: Samuel, I think the links are correct. we have same line `without ../usage/` in [json_schema.md](https://github.com/pydantic/pydantic/blob/a380749d367e1b34ac175c42672769ea6b0b2cd6/docs/usage/json_schema.md?plain=1#L171) which is in same level of pydantic_settings.md

samuelcolvin (approved):
:facepalm: LGTM

PR #165, opened 2023-09-11
Add note about nested model should inherit BaseModel

hramezani:


docs/index.md
```diff
@@ -276,6 +276,10 @@ print(Settings().model_dump())
 """
 ```
 
+1. Sub model has to inherit from `pydantic.BaseModel`, Otherwise you may get unexpected results.
+
+2. Sub model has to inherit from `pydantic.BaseModel`, Otherwise you may get unexpected results.
```
Kludex, line 281: Can we point out some reference as what this means? 🤔

Kludex (approved):

docs/index.md
hramezani: I've added more detail to it. If you think it is not enough, I have to add a separate section for it.
what do you think @Kludex?

PR #198, opened 2023-12-04
feat: ignore empty env vars

niventc:
As suggested at https://github.com/pydantic/pydantic/discussions/3796, we often have environment variables set but empty, and would prefer to use the default value for the field instead of raising an exception.

This adds the ability to ignore empty environment variables by setting `env_ignore_empty=True`, e.g.

```
os.environ['app_a'] = ''

class Settings(BaseSettings):
        a: str = 'default'

        model_config = SettingsConfigDict(env_prefix="app_", env_ignore_empty=True)

assert Settings().a == 'default'
```

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -253,7 +253,10 @@ except ValidationError as e:
 
 ## Parsing environment variable values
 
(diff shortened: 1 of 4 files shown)


hramezani:
Thanks @niventc

PR #206, opened 2024-01-05
Add support for parsing environment "None" strings to None.

### LINKED ISSUE #199: [Question] Is it still possible to pass None through environment variables?
It worked in Pydantic 1.* and I've read that we can use ENV_VAR=None|none|null to parse as None in python. Does it still work for pydantic 2.0?
short example:
```python
import os

class Settings(BaseSettings):
    ENV_VAR: str | None = None

if __name__ == "__main__":
    os.environ["ENV_VAR"] = "None"
    print(Settings().ENV_VAR)  # will be string "None" insted of an actual None
```

kschwab:
Resolves #199.

Adds support for parsing environment "None" type(str) to None type(None).

### DIFF
--- a/pydantic_settings/main.py
+++ b/pydantic_settings/main.py
@@ -26,6 +26,7 @@ class SettingsConfigDict(ConfigDict, total=False):
     env_file_encoding: str | None
     env_ignore_empty: bool
     env_nested_delimiter: str | None
+    env_parse_none: bool
     secrets_dir: str | Path | None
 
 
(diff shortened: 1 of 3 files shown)


pydantic_settings/main.py
```diff
@@ -26,6 +26,7 @@ class SettingsConfigDict(ConfigDict, total=False):
     env_file_encoding: str | None
     env_ignore_empty: bool
     env_nested_delimiter: str | None
+    env_parse_none: bool
```
kschwab, line 29: If desired, we could change this to `env_parse_none_str: str | None` with default as `None`. Users could then define what string value should be parsed (e.g. "null", "void", "None", etc.) into `None`.

hramezani: Yes, let's change it to `env_parse_none_str: str | None`

hramezani (approved):

hramezani:
Thanks @kschwab

Issue #210, opened 2024-01-19
Toml and maybe even yaml support 


samuelcolvin:
![Screenshot_20240119-085927](https://github.com/pydantic/pydantic-settings/assets/4039449/788b353f-a9d7-4870-976d-f361d70a260a)

Google suggestions makes me think toml parsing support would be very popular, we might as well also do yaml, it shouldn't be much more work.

We should use the builtin toml parser when possible, and have optional dependency groups.

Smixi:
Hello,

Learning pydantic and pydantic settings. 
I'm not used to work with optionnal dependencies. I used the same mecanism in pydantic with email-validator.
If the current code is OK for you I will add TOML, 
I basically copied the docs example from Jsonsource (which I added too).
The builtin tomlib is only available from python 3.11, but the project is >=3.8. 

>=3.11 = tomlib
<3.11 = fallback toml reader lib.

Do you have any recommandation for the tomlib you want me to use ?

kjithin:
Hello,

We should be able to create Toml & Yaml source  by inheriting InitSettingsSource. Toml and Yaml parsing libraries generates a dict which the InitSettingsSource also uses. I added the source for Toml for reference.

Is there any downside to inheriting from InitSettingsSource?

Smixi:
@kjithin I update the PR, I guess it's better to use the base class you mentionned :).

hramezani:
Added in https://github.com/pydantic/pydantic-settings/pull/211

PR #211, opened 2024-01-20
feat: adding json, yaml and toml sources

Smixi:
feat: https://github.com/pydantic/pydantic-settings/issues/210


samuelcolvin:
I think you mean toml, not JSON - please update the title.

hramezani:
Thanks @Smixi for the PR 🙏 

A couple of things to address:

- If the dependency is installed, we need to add the source class by default in `settings_customise_sources`. So, no need to override the `settings_customise_sources` to use the new source classes. They should have less priority than the existing sources to be backward-compatible.
- CI is broken for Python 3.12. it seems the `pyyaml 6.0` can't be installed. Have you tried the `pyyaml 6.0.1`?
- You've added the `JsonConfigSettingsSource` as well. not sure we want that. @samuelcolvin what do you think?

pydantic_settings/sources.py
```diff
@@ -19,10 +20,48 @@
 from pydantic_settings.utils import path_type_label
 
 if TYPE_CHECKING:
+    if sys.version_info.minor >= 11:
```
hramezani, line 23:
```suggestion
    if sys.version_info >= (3, 11):
```

Smixi:
> Thanks @Smixi for the PR 🙏
> 
> A couple of things to address:
> 
> * If the dependency is installed, we need to add the source class by default in `settings_customise_sources`. So, no need to override the `settings_customise_sources` to use the new source classes. They should have less priority than the existing sources to be backward-compatible.
> * CI is broken for Python 3.12. it seems the `pyyaml 6.0` can't be installed. Have you tried the `pyyaml 6.0.1`?
> * Documentation has to be added in https://github.com/pydantic/pydantic-settings/blob/main/docs/index.md
> * You've added the `JsonConfigSettingsSource` as well. not sure we want that. @samuelcolvin what do you think?

If they are default loaded, it means the default settings class is in settings_customise_sources, which will result in the API breaking (people having custom class order will fail) or does it make sense ?

hramezani:
Here is the source settings priority right now:

- init_settings
- env_settings
- dotenv_settings
- file_secret_settings

So, if we add new sources at the end there shouldn't be a breaking change and there will be new sources that users can use by defining the file path and encoding in config.

It means that releasing the new `pydantic-settings` won't break existing settings classes and users can benefit from new sources by defining the required configs.

Smixi:
My concerns is more about user having their own custom orders. They will need to reimplement the `settings_customise_sources` to add the new parameters (like I did in the tests) ?
Pyaml 6.0.1 seems to fix the issue

hramezani:
> My concerns is more about user having their own custom orders. They will need to reimplement the settings_customise_sources to add the new parameters (like I did in the tests) ?

That's a good point. didn't think about this.
So, please revert the change and document how to use the new sources. sorry for this

Smixi:
@hramezani  That should be good. I just don't know how to deal with the coverage.

pydantic_settings/sources.py
```diff
@@ -664,6 +704,103 @@ def __repr__(self) -> str:
         )
 
 
+class ConfigFileSourceMixin(ABC):
+    def _read_files(self, files: PathType | None) -> dict[str, Any | None]:
+        if files is None:
+            return {}
+        if isinstance(files, (str, os.PathLike)):
+            files = [files]
+        vars: dict[str, Any | None] = {}
```
hramezani, line 713:
```suggestion
        vars: dict[str, Any] = {}
```

pydantic_settings/sources.py
```diff
@@ -664,6 +704,103 @@ def __repr__(self) -> str:
...
+        for file in files:
+            file_path = Path(file).expanduser()
+            if file_path.is_file():
+                vars.update(self._read_file(file_path))
+        return vars
+
+    @abstractmethod
+    def _read_file(self, path: Path) -> dict[str, Any | None]:
```
hramezani, line 721:
```suggestion
    def _read_file(self, path: Path) -> dict[str, Any]:
```

pydantic_settings/sources.py
```diff
@@ -664,6 +704,103 @@ def __repr__(self) -> str:
         )
 
 
+class ConfigFileSourceMixin(ABC):
+    def _read_files(self, files: PathType | None) -> dict[str, Any | None]:
```
hramezani, line 708:
```suggestion
    def _read_files(self, files: PathType | None) -> dict[str, Any]:
```

pydantic_settings/sources.py
```diff
@@ -664,6 +704,103 @@ def __repr__(self) -> str:
...
+            yaml_file_encoding
+            if yaml_file_encoding is not None
+            else settings_cls.model_config.get('yaml_file_encoding')
+        )
+        self.yaml_data = self._read_files(self.yaml_file_path)
+        super().__init__(settings_cls, self.yaml_data)
+
+    def _read_file(self, file_path: Path) -> dict[str, Any | None]:
+        with open(file_path, encoding=self.yaml_file_encoding) as yaml_file:
+            import_yaml()
```
hramezani, line 800: Isn't it better to move `import_yaml` out of the block to check the import before opening the file?

pydantic_settings/sources.py
Smixi: As in env settings, I think this is allowed in config files to have null values, as it's defined in json and in yaml specs ?

hramezani:
> @hramezani That should be good. I just don't know how to deal with the coverage.

I think it can be tested like [the old dotenv test in Pydantic V1](https://github.com/pydantic/pydantic/blob/3ddb509786089f9df0658303545ae4b66db2d47c/tests/test_settings.py#L909)

and then you can add a new step to the CI to [uninstall deps and run the tests](https://github.com/pydantic/pydantic/blob/3ddb509786089f9df0658303545ae4b66db2d47c/.github/workflows/ci.yml#L242) to run the above tests

.github/workflows/ci.yml
```diff
@@ -56,6 +56,15 @@ jobs:
         COVERAGE_FILE: .coverage.${{ runner.os }}-py${{ matrix.python }}
         CONTEXT: ${{ runner.os }}-py${{ matrix.python }}
 
+    - name: uninstall deps
+      run: pip uninstall -y tomlkit PyYAML python-dotenv
```
hramezani, line 60: `python-dotenv` is not an optional dep. no need to test for it

Smixi: Should I remove it from the test then ? It was previously set here:
```python
@pytest.mark.skipif(dotenv, reason='python-dotenv is installed')
def test_dotenv_not_installed(tmp_path):
    p = tmp_path / '.env'
    p.write_text('a=b')
```

hramezani: Yes, `python-dotenv` is no an optional dep and will be installed with `pydantic-settings`

hramezani (approved):

hramezani:
LGTM. Thanks @Smixi.

@samuelcolvin I will keep this PR open. please take a look if you want.

korolkevich:
Hi @Smixi,
Thank you for your work.

I have a question:
If we have nested YAML configuration:
`app:
  pass: pass
  login: login
`
How do we replace the default configuration with environment variables, as it is done in Dynaconf (app__pass='new_val' or app.pass='new_val'), and handle deeper configurations? I couldn't find this in your PR.

Smixi:
@korolkevich This PR is only a loader for variables. 
I think your use case is more about using [this mecanism](https://docs.pydantic.dev/latest/concepts/pydantic_settings/#changing-priority) and declare env_settings before the yaml setting source

docs/index.md
```diff
@@ -534,6 +534,66 @@ Last, run your application inside a Docker container and supply your newly creat
...
+        return (TomlConfigSettingsSource(settings_cls),)
+```
+
+This will be able to read the following "config.toml" file, located in your working directory:
+
+```toml
+foobar = "Hello"
+[nested]
+nested_field = "world!"
+"""
```
samuelcolvin, line 594: I don't understand what this is?

samuelcolvin (approved):
otherwise LGTM.

Thank you so much.

hramezani:
Thanks @Smixi

PR #204, opened 2024-01-02
Allow nested env var source to override nested init source.

kschwab:
Resolves #203

Allows for nested `EnvSettingsSource` to override nested `InitSettingsSource`.

Issue #196, opened 2023-11-29
Optional (or Union) types cannot be loaded from JSON secrets


daBrado:
If one tries to use an optional settings field with a complex type, it cannot be assigned to from a secrets file that contains JSON.  For example:

```python
from os import environ
from os.path import join
from tempfile import TemporaryDirectory

from pydantic import BaseModel
from pydantic_settings import BaseSettings


class Foo(BaseModel):
    a: int


class Settings(BaseSettings):
    foo: Foo


class SettingsOptional(BaseSettings):
    foo: Foo | None = None


with TemporaryDirectory() as tmpdir:
    with open(join(tmpdir, "foo"), "w") as f:
        f.write('{"a":10}')

    bar = Settings(_secrets_dir=tmpdir)
    print(bar)  # foo=Foo(a=10), as expected.

    # Errors out with:
    #   Traceback (most recent call last):
    #     File "/.../example.py", line 41, in <module>
    #       bar2 = SettingsOptional(_secrets_dir=tmpdir)
    #              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    #     File "/.../python3.11/site-packages/pydantic_settings/main.py", line 71, in __init__
    #       super().__init__(
    #     File "/.../lib/python3.11/site-packages/pydantic/main.py", line 164, in __init__
    #       __pydantic_self__.__pydantic_validator__.validate_python(data, self_instance=__pydantic_self__)
    #   pydantic_core._pydantic_core.ValidationError: 1 validation error for SettingsOptional
    #   foo
    #     Input should be a valid dictionary or instance of Foo [type=model_type, input_value='{"a":10}', input_type=str]
    #       For further information visit https://errors.pydantic.dev/2.5/v/model_type
    bar2 = SettingsOptional(_secrets_dir=tmpdir)
```

This seems inconsistent.  I wouldn't expect making the `foo` field optional to disallow the ability to load its value as JSON.

hramezani:
Thanks @daBrado for reporting this issue and sorry for the late response.

Yeah, seems like a bug to me and the reason is by making `foo` optional(`foo: Foo | None = None`), `pydantic-settings` does not consider the field as a complex field. 

But this is not the case in env source. if you set `FOO='{"a":10}'` env variable it `SettingsOptional` works as expected.

PR #214, opened 2024-01-22
Add CLI Settings Source

### LINKED ISSUE #209: Feature Request: Add CLI argument parsing for model initialization
Hi @hramezani,

I wanted to propose adding CLI argument parsing into Pydantic settings. Under the hood, it is essentially the same as environment variable parsing, so not very difficult to add (relatively speaking). This would be nice to have as it would complete all sources of ingestion at the application level (FILES, ENV, CLI). I am in the process of completing a rough draft and wanted to see if there was interest in bringing something like this into main. Would love to contribute and take feedback on what would be desired if interested. 

As with [parsing environment variables example](https://docs.pydantic.dev/latest/concepts/pydantic_settings/#parsing-environment-variable-values), nested settings would take precedence, etc. Essentially, the below:
```python
# your environment
export V0=0
export SUB_MODEL='{"v1": "json-1", "v2": "json-2"}'
export SUB_MODEL__V2=nested-2
export SUB_MODEL__V3=3
export SUB_MODEL__DEEP__V4=v4
```
would be equivalent to:
```bash
# your cli
app --v0=0 --sub_model='{"v1": "json-1", "v2": "json-2"}' --sub_model.v2 nested-2 \
--sub_model.v3 3 --sub_model.deep.v4 v4
```

Then, within the application it would simply be:
```python

class DeepSubModel(BaseModel):  
    v4: str


class SubModel(BaseModel):  
    v1: str
    v2: bytes
    v3: int
    deep: DeepSubModel


class Settings(BaseSettings):
    model_config = SettingsConfigDict(env_nested_delimiter='__')

    v0: str
    sub_model: SubModel


print(Settings.parse_cli().model_dump())
"""
{
    'v0': '0',
    'sub_model': {'v1': 'json-1', 'v2': b'nested-2', 'v3': 3, 'deep': {'v4': 'v4'}},
}
"""
```

Thoughts?

kschwab:
Resolves #209.

This PR implements a `CliSettingsSource` with the following main features outlined below. Refer to [docs/index.md](https://github.com/pydantic/pydantic-settings/pull/214/files#diff-b4d68dc855d0f9476d3f2ee343853bd21bf82ea9960d0cf06661baa244439dd6R462) for more details.

### Fields

* Same expectations as setting environment variables
* Proposed `_cli_parse_args` can accept arg array or `bool` set to `True` for `sys.argv`

```python

class DeepSubModel(BaseModel):
    v4: str


class SubModel(BaseModel):
    v1: str
    v2: bytes
    v3: int
    deep: DeepSubModel


class Settings(BaseSettings):
    v0: str
    sub_model: SubModel


print(Settings(_cli_parse_args=[
    '--v0=0',
    '--sub_model={"v1": "json-1", "v2": "json-2"}',
    '--sub_model.v2', 'nested-2',
    '--sub_model.v3', '3',
    '--sub_model.deep.v4', 'v4'
]).model_dump())
'''
{'v0': '0', 'sub_model': {'v1': 'json-1', 'v2': b'nested-2', 'v3': 3, 'deep': {'v4': 'v4'}}}
'''
```

### Lists

* Proposed intermixing of three list styles:
  * JSON style `--field='[1,2,3]'`
  * Argparse style `--field 1 --field 2 --field 3`
  * Lazy style `--field=1,2,3`

```python
from typing import List


class Settings(BaseSettings):
    my_list: List[int]

print(Settings(_cli_parse_args=[
    '--my_list', '[1,2]',
    '--my_list', '3,4',
    '--my_list', '5', '--my_list', '6'
]).model_dump())
'''
{'my_list': [1, 2, 3, 4, 5, 6]}
'''
```

### Dictionaries

* Proposed intermixing of two dictionary styles:
  * JSON style `--field='{"k1": 1, "k2": 2}'`
  * Env Var style `--field k1=1 --field k2=2`
* These styles can be used in conjunction with list forms as well:
  *  `--field k1=1,k2=2 --field k3=3 --field '{"k4: 4}'` etc.

```python
from typing import Dict


class Settings(BaseSettings):
    my_dict: Dict[str, str]


print(Settings(_cli_parse_args=[
    '--my_dict', '{"k1": "a", "k2": "b"}',
    '--my_dict', '{"k3": "c"}, {"k4": "d"}',
    '--my_dict', '{"k5": "e"}', '--my_dict', '{"k6": "f"}',
    '--my_dict', '[k7=g,k8=h]',
    '--my_dict', 'k9=i,k10=j',
    '--my_dict', 'k11=k', '--my_dict', 'k12=l',
]).model_dump())
'''
{'my_dict': {
    'k1': 'a',
    'k2': 'b',
    'k3': 'c',
    'k4': 'd',
    'k5': 'e',
    'k6': 'f',
    'k7': 'g',
    'k8': 'h',
    'k9': 'i',
    'k10': 'j',
    'k11': 'k',
    'k12': 'l',
}}
'''
```

### Subcommands and Positionals

* Proposed help text generation based off of field descriptions and class documentation
* Proposed `_cli_prog_name` has same meaning as in argparse
* Proposed `CliSubCommand` annotation for marking subcommands
* Proposed `CliPositionalArg` annotation for marking positional arguments

```python
from pydantic_settings import BaseSettings, CliSubCommand, CliPositionalArg, PydanticBaseSettingsSource


class FooPlugin(BaseModel, use_attribute_docstrings=True):
    '''git-plugins-foo - Extra deep foo plugin command'''

    my_feature: bool = False
    '''Enable my feature on foo plugin'''


class BarPlugin(BaseModel, use_attribute_docstrings=True):
    '''git-plugins-bar - Extra deep bar plugin command'''

    my_feature: bool = False
    '''Enable my feature on bar plugin'''


class Plugins(BaseModel, use_attribute_docstrings=True):
    '''git-plugins - Fake plugins for GIT'''

    foo: CliSubCommand[FooPlugin]
    '''Foo is fake plugin'''

    bar: CliSubCommand[BarPlugin]
    '''Bar is also a fake plugin'''


class Clone(BaseModel, use_attribute_docstrings=True):
    '''git-clone - Clone a repository into a new directory'''

    repository: CliPositionalArg[str]
    '''The repository to clone'''

    directory: CliPositionalArg[str]
    '''The directory to clone into'''

    local: bool = False
    '''When the resposity to clone from is on a local machine, bypass ...'''

    shared: bool = False
    '''Force the clone process from a reposity on a local filesystem ...'''


class Init(BaseModel, use_attribute_docstrings=True):
    '''git-init - Create an empty Git repository or reinitialize an existing one'''

    directory: CliPositionalArg[str]
    '''The directory to clone into'''

    quiet: bool = False
    '''Only print error and warning messages; all other input will be suppressed'''

    bare: bool = False
    '''Create a bare repository. If GIT_DIR environment is not set, it is set ...'''


class Git(BaseSettings, use_attribute_docstrings=True):
    '''git - The stupid content tracker'''

    clone: CliSubCommand[Clone]
    '''Clone a repository into a new directory'''

    init: CliSubCommand[Init]
    '''Create an empty Git repository or reinitialize an existing one'''

    plugins: CliSubCommand[Plugins]
    '''Fake GIT plugin commands'''


print(Git(_cli_prog_name='git', _cli_parse_args=['--help']))
'''
usage: git [-h] {clone,init,plugins} ...

git - The stupid content tracker

options:

subcommands:
  {clone,init,plugins}
    clone               Clone a repository into a new directory
    init                Create an empty Git repository or reinitialize an existing one
    plugins             Fake GIT plugin commands
'''


print(Git(_cli_prog_name='git', _cli_parse_args=['clone', '--help']))
'''
usage: git clone [-h] [--local bool] [--shared bool] REPOSITORY DIRECTORY

git-clone - Clone a repository into a new directory

positional arguments:
  REPOSITORY     The repository to clone
  DIRECTORY      The directory to clone into

options:
  -h, --help     show this help message and exit
  --local bool   When the resposity to clone from is on a local machine, bypass ...
  --shared bool  Force the clone process from a reposity on a local filesystem ...
'''


print(Git(_cli_prog_name='git', _cli_parse_args=['plugins', 'bar', '--help']))
'''
usage: git plugins bar [-h] [--my_feature bool]

git-plugins-bar - Extra deep bar plugin command

options:
  -h, --help         show this help message and exit
  --my_feature bool  Enable my feature on bar plugin
'''
```

###


kschwab:
@hramezani and @samuelcolvin 

I am most of the way through but hit this limitation in the underlying [environment variable parsing](https://docs.pydantic.dev/latest/concepts/pydantic_settings/#parsing-environment-variable-values):

> JSON is only parsed in top-level fields, if you need to parse JSON in sub-models, you will need to implement validators on those models.

Obviously, this would be very nice to have from a CLI perspective (e.g., nested lists etc.). I didn't know this was the case until I started writing nested tests. Do you have suggestions on what the best course of action is here? Ideally, I would love for it to be handled in environment parsing. If not, from CLI perspective the next easiest path would be nested dictionaries of strings. Is that something I can throw to Pydantic and have a reasonable expectation it will sort it out?

kschwab:
Sorry @hramezani, I should have been more specific with my question. I am at a fork and am looking for guidance on which direction to take.

Is there any background on why nested JSON isn't supported in `EnvSettingsSource`? Was there something that scared us away from adding it? Is it a massive pile of work? etc. 

My preference would be to solve it there so that the CLI and ENV are consistent. Just trying to ascertain if that is a reasonable direction or not.

kschwab:
Got it. So it's either a bug or something invalid in the test case. Below is an example of the issue I am seeing:

```python
import os
from typing import Optional, List


class Obj(BaseModel):
    val: int

class Child(BaseModel):
    num_list: Optional[List[int]] = None
    obj_list: Optional[List[Obj]] = None
    str_list: Optional[List[str]] = None
    union_list: Optional[List[Obj | int]] = None

class Cfg(BaseSettings, env_nested_delimiter='__'):
    num_list: Optional[List[int]] = None
    obj_list: Optional[List[Obj]] = None
    union_list: Optional[List[Obj | int]] = None
    str_list: Optional[List[str]] = None
    child: Optional[Child] = None

os.environ['CHILD__NUM_LIST'] = '[1,2,3]'
print(Cfg().model_dump())
'''
  File "/home/kschwab/pydantic-settings/counter_example.py", line 23, in <module>
    print(Cfg().model_dump())
  File "/home/kschwab/pydantic-settings/pydantic_settings/main.py", line 93, in __init__
    super().__init__(
  File "/home/kschwab/.local/lib/python3.10/site-packages/pydantic/main.py", line 171, in __init__
    self.__pydantic_validator__.validate_python(data, self_instance=self)
pydantic_core._pydantic_core.ValidationError: 1 validation error for Cfg
child.num_list
  Input should be a valid list [type=list_type, input_value='[1,2,3]', input_type=str]
    For further information visit https://errors.pydantic.dev/2.6/v/list_type
'''
```

hramezani:
It's a bug. Here is the related [issue](https://github.com/pydantic/pydantic-settings/issues/196)

pydantic_settings/main.py
```diff
@@ -94,6 +113,7 @@ def settings_customise_sources(
         cls,
         settings_cls: type[BaseSettings],
         init_settings: PydanticBaseSettingsSource,
+        cli_settings: PydanticBaseSettingsSource,
```
hramezani, line 116: Please remember we can't change the signature of this function because it will break user code.

Probably, we can add a special `model_validate_cli` that temporarily adds the CLI source at the start of sources, then does validation

kschwab: Yep, this is one of main points of discussion that needed to be had. There are actually a few that need feedback, I'll try and comment the others when I get time.

These are the decisions that needed consideration:
- Do we want break backwards compatibility by adding CLI source as a core source? - No.
  - Benefit here is that CLI source fits into exact same paradigm as other core sources "out of the box".
- Should CLI source have prioritization levels?
  - If yes, it means we will need to have a separate `cli_settings_customise_sources` hook for users to arrange prioritization of CLI sources relative to others
  - Alternatively, we could do something like I outlined [here](https://github.com/pydantic/pydantic-settings/issues/209#issuecomment-1900855951), but it makes it awkward with how args are passed in. Specifically, users have the potential to define args in two different places which I don't like (`_cli_parse_args` and instatiation of `CliSettingsSource`). So, we would need to do something there to make it one location.
- Should the CLI source have a fixed prioritization level, and if so, should it always be the topmost priority?
  - After playing with the implementation, this is the one makes the most sense to me, which is what I think you're advocating for above. Typically users will probably treat `_cli_parse_args` as an alternative to "init_kwargs", or with the intention of topmost priority. This does eliminate use cases where this is not true, but I imagine that is a very small minority.

hramezani: > Should CLI source have prioritization levels?

I would say we can add it as first priority if users want to benefit from it. I've added [an issue](https://github.com/pydantic/pydantic-settings/issues/216) to address the `settings_customise_sources` signature problem in V3

> Should the CLI source have a fixed prioritization level, and if so, should it always be the topmost priority?

Yes, if the user wants to use the CLI source it has to be the top priority source.

kschwab: Sounds good. I'll head in that direction.

pydantic_settings/sources.py
```diff
@@ -515,7 +514,7 @@ def _field_is_complex(self, field: FieldInfo) -> tuple[bool, bool]:
         return True, allow_parse_failure
 
     @staticmethod
-    def next_field(field: FieldInfo | None, key: str) -> FieldInfo | None:
+    def next_field(field: FieldInfo | Any | None, key: str) -> FieldInfo | None:
```
kschwab, line 517: I ended up tackling https://github.com/pydantic/pydantic-settings/issues/196. Please review if this looks OK.

kschwab:
The bulk of the functionality is now complete. Will continue to work on cleanup, comments, docs etc. over next few days. The only key remaining feedback items needed for any adjustments are:

- Subcommand and Positional Argument Implementation (refer to GIT CLI example for reference)
  - [ ] Does the `CliSubCommand` and `CliPositionalArg` annotations for denoting their respective meanings look OK?
    - There are a few ways to accomplish this. I took the stance of a more explicit annotation at the user level for denoting subcommands and positional arguments. Overall, it is quite clean and makes intent clear when reading the model. The alternative approach is implicit coercion, with union fields into subcommands and required fields into positionals. This takes flexibility away from user, who may want a similar experience to that as environment variable overrides. Therefore, the default experience is in line with how Pydantic is already structured, and deviation from the expectation is expressed through these annotations.
  - [ ] Is single subparser per model OK? 
    - I took an opinionated stance of only supporting a single subcommand group (single subparser) per model. If we want to support multiple subparsers, we would likely need to go back to an earlier version I had [here](https://github.com/pydantic/pydantic-settings/pull/214/commits/2ef847323d3638881d8ac5a0cdfad9fa063536e0#diff-f20069a79c955c52519f37001105df187ba738b9a4dd49af010c723ff810200cR2011). I changed from that direction however because:
      - Better control of generated help text. With a union implementation, the help text for the subcommand gets pulled from the class doc, which works so long as it is not long. However, in our application, we work with larger models that do have verbose class documentation so it can flood the help text. Simple workaround would be parsing the first sentence of the class doc etc. but not worth it IMO. With current implementation, the subcommand help text is the field doc string making it very straightforward. Then, the class doc itself is displayed when using help on the subcommand itself. e.g. `git --help` shows condensed help for the `clone` subcommand, whereas `git clone --help` shows the expanded class doc help text there.
      - Explicit naming of subcommands. In union implementation, the subcommand name must be coerced from the class name or some other mechanism. With current implementation, it is whatever the subcommand field name is.
      - Avoids the need for [Discriminated Unions](https://docs.pydantic.dev/2.5/concepts/unions/#discriminated-unions). Absolutely an option but requires more boilerplate, which takes away from the clean aesthetic. This was actually the forcing issue that caused me to change direction. The normal union selection was not strong enough so users would have needed to add discriminated unions for any subcommand.
  - [ ] Positional arguments always placed before subcommands?
    - Another opinionated stance, although of smaller concern. As is today, there is no requirement with regards to field ordering within the definition of a model. However, the resultant CLI will reorder positional args before subcommands. This is mostly to prevent users from shooting themselves in the foot. Perhaps this is best to just change to a flag, i.e., turn the safety off?

Outside of the above, I think everything else is pretty straightforward and aligns with precedence already set by environment variables.

pydantic_settings/main.py
```diff
@@ -214,11 +223,12 @@ def _settings_build_values(
         sources = self.settings_customise_sources(
             self.__class__,
             init_settings=init_settings,
-            cli_settings=cli_settings,
             env_settings=env_settings,
             dotenv_settings=dotenv_settings,
             file_secret_settings=file_secret_settings,
         )
+        if _cli_parse_args:
+            sources = (cli_settings,) + sources
```
kschwab, line 231: If `_cli_parse_args` is provided, it will always be topmost source.

pydantic_settings/sources.py
kschwab: Opened in PR https://github.com/pydantic/pydantic-settings/pull/217.

kschwab:
Hi @hramezani, need some help with resolving the below:

1. Many of the doc examples use CLI `--help` to view generated help text etc. However, this results in an exit, which pytest does not like. How to update pytest such that exit is allowed in these expected cases?
2. Many of the examples use the [recently merged](https://github.com/pydantic/pydantic/pull/6563) `use_attribute_docstrings` flag, which does not appear available in tests. How to update the tests such that the pydantic version includes this commit?
3. Is there a way to generate the docs locally for viewing?

hramezani:
> Many of the doc examples use CLI --help to view generated help text etc. However, this results in an exit, which pytest does not like. How to update pytest such that exit is allowed in these expected cases?

You can use
```py 
```py test="skip"
```
to skip running on those blocks 

> Many of the doc examples use the https://github.com/pydantic/pydantic/pull/6563 use_attribute_docstrings flag, which does not appear available in tests. How to update the tests such that the pydantic version includes this commit?

This will be available on `Pydantic 2.6`. It is in beta and will be released soon. We are planing to release the next `pydantic-settings` after `Pydantic 2.6`. So, in the end, we can use `use_attribute_docstrings` flag.

> Is there a way to generate the docs locally for viewing?

Not right now. `pydantic-settings` docs will be rendered under pydantic docs. But I will work on it to make it possible to render docs in `pydantic-settings`

kschwab:
Hi @hramezani, almost ready to move this into PR. There are three final items:

1. I do not plan to add coverage for [this line](https://github.com/pydantic/pydantic-settings/pull/214/files#diff-b4b68ace3cb0cf2820d1d882735748b675a379c39463ccc89700a9618a80a2a2R716), unless there is a safe/easy way in pytest to modify `sys.argv`. Is this acceptable? If so, how to exclude?
2. I would like to exclude [this function](https://github.com/pydantic/pydantic-settings/pull/214/files#diff-b4b68ace3cb0cf2820d1d882735748b675a379c39463ccc89700a9618a80a2a2R957) from coverage or leverage the original test case if you can help locate. I pulled this from [pydantic._internal._repr.py](https://github.com/pydantic/pydantic/blob/main/pydantic/_internal/_repr.py#L85). Please provide guidance on desired direction.
3. Also related to above number two, I am hitting a `mypy` linting issue related to `typing_extensions.TypeAliasType` [here](https://github.com/pydantic/pydantic-settings/pull/214/files#diff-b4b68ace3cb0cf2820d1d882735748b675a379c39463ccc89700a9618a80a2a2R965). This is directly copied from the `_repr` version, how would you like me to proceed?

hramezani:
> I do not plan to add coverage for [this line](https://github.com/pydantic/pydantic-settings/pull/214/files#diff-b4b68ace3cb0cf2820d1d882735748b675a379c39463ccc89700a9618a80a2a2R717), unless there is a safe/easy way in pytest to modify sys.argv. Is this acceptable? If so, how to exclude?

Can't we use `monkeypatch` fixture like `monkeypatch.setattr("sys.argv", ["test1", "-f", "test2"])`?

> I would like to exclude [this function](https://github.com/pydantic/pydantic-settings/pull/214/files#diff-b4b68ace3cb0cf2820d1d882735748b675a379c39463ccc89700a9618a80a2a2R957) from coverage or leverage the original test case if you can help locate. I pulled this from [pydantic._internal._repr.py](https://github.com/pydantic/pydantic/blob/main/pydantic/_internal/_repr.py#L85). Please provide guidance on desired direction.

There are some tests in pydantic repo

- https://github.com/pydantic/pydantic/blob/c1dff153c8ba2894e2b060cf6e53de6851ba4cd9/tests/test_utils.py#L83
- https://github.com/pydantic/pydantic/blob/c1dff153c8ba2894e2b060cf6e53de6851ba4cd9/tests/test_utils.py#L106

> Also related to above number two, I am hitting a mypy linting issue related to typing_extensions.TypeAliasType [here](https://github.com/pydantic/pydantic-settings/pull/214/files#diff-b4b68ace3cb0cf2820d1d882735748b675a379c39463ccc89700a9618a80a2a2R965). This is directly copied from the _repr version, how would you like me to proceed?

Please ignore it

kschwab:
Thanks @hramezani for the feedback. Ready for review 👍🏾

pydantic_settings/sources.py
```diff
@@ -2,25 +2,31 @@
...
 
+import typing_extensions
 from dotenv import dotenv_values
 from pydantic import AliasChoices, AliasPath, BaseModel, Json, TypeAdapter
-from pydantic._internal._typing_extra import WithArgsTypes, origin_is_union
+from pydantic._internal._typing_extra import WithArgsTypes, origin_is_union, typing_base
 from pydantic._internal._utils import deep_update, is_model_class, lenient_issubclass
 from pydantic.fields import FieldInfo
-from typing_extensions import get_args, get_origin
+from pydantic.v1.utils import Representation
```
hramezani, line 21: Why did you import it from `pydantic.v1`?
```suggestion
from pydantic._internal._repr import Representation
```

pydantic_settings/main.py
hramezani: Shouldn't it be `if cli_parse_args`?

I mean to enable it by either `_cli_parse_args` or `cli_parse_args` config.

pydantic_settings/sources.py
```diff
@@ -674,6 +695,309 @@ def __repr__(self) -> str:
...
+
+    def _metavar_format_recurse(self, obj: Any) -> str:
+        """Pretty metavar representation of a type. Adapts logic from `pydantic._repr.display_as_type`."""
+        if isinstance(obj, FunctionType):
+            return obj.__name__
+        elif obj is ...:
+            return '...'
+        elif isinstance(obj, Representation):
+            return repr(obj)
+        elif isinstance(obj, typing_extensions.TypeAliasType):
```
hramezani, line 973:
```suggestion
        elif isinstance(obj, typing_extensions.TypeAliasType):  # type: ignore
```

pydantic_settings/sources.py
kschwab: Yeah, I feel like this is similar to the `typing.Literal` vs `typing_extensions.Literal` [issue](https://github.com/pydantic/pydantic-settings/pull/214/files/20a83f1aaa3b0aa17573ca791a387f5c772ad137#diff-b4b68ace3cb0cf2820d1d882735748b675a379c39463ccc89700a9618a80a2a2R982). 

When I first ported the test case, [the `Representation` line](https://github.com/pydantic/pydantic-settings/pull/214/files/20a83f1aaa3b0aa17573ca791a387f5c772ad137#diff-b4b68ace3cb0cf2820d1d882735748b675a379c39463ccc89700a9618a80a2a2R972) wasn't covered. Originally, I had the suggestion above and instantiated from `pydantic._internal`, but it still did not exercise the line. I suspect in Pydantic the coverage is coming from elsewhere, I'm not sure how they're hitting it. However, the `pydantic.v1` variant does. Since both are possible, I'll just update to check for both.

tests/test_settings.py
```diff
@@ -1993,3 +2008,557 @@ class Settings(BaseSettings):
...
+    ],
+)
+def test_cli_metavar_format_310(value_gen, expected):
+    value = value_gen()
+    assert CliSettingsSource(SimpleSettings)._metavar_format(value) == expected
+
+
+@pytest.mark.skipif(sys.version_info < (3, 12), reason='requires python 3.12 or higher')
+def test_cli_metavar_format_type_alias_312():
+    exec(
```
kschwab, line 2559: Is this OK? I didn't know how else to get coverage on the `TypeAliasType` without breaking earlier versions of Python (i.e. `type TypeAliasInt = int` is only available starting in Python 3.12), so used exec here to avoid an interpreter error.

pydantic_settings/sources.py
```diff
@@ -674,6 +696,309 @@ def __repr__(self) -> str:
...
+        elif get_origin(obj) in (typing_extensions.Literal, typing.Literal):
+            args = self._metavar_format_list(list(map(repr, self._get_modified_args(obj))))
+            return f'{{{args}}}' if ',' in args else args
+        elif isinstance(obj, WithArgsTypes):
+            args = self._metavar_format_list(list(map(self._metavar_format_recurse, self._get_modified_args(obj))))
+            return f'{obj.__qualname__}[{args}]'
+        elif obj is type(None):
+            return self.env_parse_none_str
+        elif is_model_class(obj):
+            return 'JSON'
```
kschwab, line 992: Any `BaseModel` type in CLI help text will be displayed as `JSON`, is this OK? I felt it is more confusing seeing a class name(s) at the CLI help text as opposed to "JSON". You'll have to refer to the model anyways to know how to structure the JSON input. e.g.

```python

class SubModel(BaseModel):
    v1: int

class Settings(BaseSettings):
    sub_model: SubModel

# Displaying models as "JSON" 
Settings(_cli_parse_args=['--help'])
"""
usage: example.py [-h] [--sub_model JSON] [--sub_model.v1 int]

options:
  -h, --help          show this help message and exit
sub_model options:
  --sub_model JSON    set sub_model from JSON string
  --sub_model.v1 int 
"""

# Displaying models as class name
Settings(_cli_parse_args=['--help'])
"""
usage: example.py [-h] [--sub_model SubModel] [--sub_model.v1 int]

options:
  -h, --help          show this help message and exit
sub_model options:
  --sub_model SubModel    set sub_model from JSON string
  --sub_model.v1 int 
"""
```

pydantic_settings/sources.py
```diff
@@ -459,7 +479,9 @@ def prepare_field_value(self, field_name: str, field: FieldInfo, value: Any, val
         """
         is_complex, allow_parse_failure = self._field_is_complex(field)
         if is_complex or value_is_complex:
-            if value is None:
+            if isinstance(value, EnvNoneType):
```
kschwab, line 482: This was a missing case I found from https://github.com/pydantic/pydantic-settings/pull/204, setting a complex field to `None`.

kschwab: Also noted that I did not add an explicit test case. Will add.

kschwab: Added test cases.

pydantic_settings/sources.py
hramezani: I think it is ok

pydantic_settings/main.py
```diff
@@ -59,6 +66,15 @@ class BaseSettings(BaseModel):
         _env_nested_delimiter: The nested env values delimiter. Defaults to `None`.
         _env_parse_none_str: The env string value that should be parsed (e.g. "null", "void", "None", etc.)
             into `None` type(None). Defaults to `None` type(None), which means no parsing should occur.
+        _cli_prog_name: The CLI program name to display in help text. Defaults to `None` if _cli_parse_args is `None`.
+            Otherwse, defaults to sys.argv[0].
+        _cli_parse_args: The list of CLI arguments to parse. Defaults to None.
+            If set to `True`, defaults to sys.argv[1:].
+        _cli_hide_none_type: Hide NoneType values in CLI help text. Defaults to `False`.
```
samuelcolvin, line 73:
```suggestion
        _cli_hide_none_type: Hide `None` values in CLI help text. Defaults to `False`.
```

docs/index.md
```diff
@@ -459,6 +459,531 @@ class Settings(BaseSettings):
...
+
+By default, the experience is tailored towards use case #1 and builds on the foundations established in [parsing
+environment variables](#parsing-environment-variables). If your use case primarily falls into #2, you will likely want
+to enable [enforcing required arguments at the CLI](#enforce-required-arguments-at-cli).
+
+### The Basics
+
+To get started, let's look at a basic example for defining a Pydantic settings CLI:
+
+```py test="skip"
```
samuelcolvin, line 478: This example isn't very "basic", better to start with a trivial example.

samuelcolvin: also, would be good if we didn't have to skip tests for most of these.

docs/index.md
```diff
@@ -459,6 +459,531 @@ class Settings(BaseSettings):
...
+    """The Settings class documentation will show in top level help text."""
+
+    v0: str
+    """the top level v0 option"""
+
+    sub_model: SubModel
+    """The help summary for SubModel related options. This will be placed at top of group."""
+
+
+Settings(_cli_prog_name='app', _cli_parse_args=['--help'])  # (1)!
```
samuelcolvin, line 513: I would try to avoid using the `_` kwargs to `__init__`, and use class kwargs instead in most places in the docs - I think that approach is clearer and easier to understand.

docs/index.md
```diff
@@ -459,6 +459,531 @@ class Settings(BaseSettings):
...
+its priority](#changing-priority).
+
+#### Enable CLI Argument Parsing
+
+`cli_parse_args: Optional[Union[List[str], bool]] = None`
+
+* Default = `None`
+* If `True`, parse from `sys.argv[1:]`
+* If `List[str]`, parse from `List[str]`
+* If `False` or `None`, do not parse CLI arguments
```
samuelcolvin, line 573: this looks like it should be in a docstring.

docs/index.md
```diff
@@ -459,6 +459,531 @@ class Settings(BaseSettings):
...
+
+#### Lists
+
+CLI argument parsing of lists supports intermixing of any of the below three styles:
+
+  * JSON style `--field='[1,2]'`
+  * Argparse style `--field 1 --field 2`
+  * Lazy style `--field=1,2`
+
```
samuelcolvin, line 582: 3 seems excessive, surely just Argparse and lazy should be allowed.

Or Argparse and JSON if we need json elsewhere?

docs/index.md
```diff
@@ -459,6 +459,531 @@ class Settings(BaseSettings):
...
+from pydantic import BaseModel
+
+from pydantic_settings import (
+    BaseSettings,
+    CliPositionalArg,
+    CliSubCommand,
+)
+
+
+class FooPlugin(BaseModel, use_attribute_docstrings=True):
```
samuelcolvin, line 653: rather than using `use_attribute_docstrings=True` everywhere, we would be better of using `Field(description='...')` in all but one example.

pydantic_settings/sources.py
```diff
@@ -674,6 +715,320 @@ def __repr__(self) -> str:
         )
 
 
+class CliSettingsSource(EnvSettingsSource):
```
samuelcolvin, line 718: I haven't read all of this, I assume @hramezani has reviewed it?

samuelcolvin (commented):
In general, I think this looks great. Thank you so much.

Writing so much documentation is amazing, but I would lean a bit towards smaller examples each demonstrating one thing.

hramezani:
please update

docs/index.md
kschwab: > This example isn't very "basic", better to start with a trivial example.

Sure, will update.

> also, would be good if we didn't have to skip tests for most of these.

Agreed. I tried to keep them where possible. The issue we encountered was for doc examples showing CLI `--help` which result in an argparse `SystemExit` causing pytest to fail. If there is a way to have the doc tests handle `SystemExit`, we could re-enable these.

docs/index.md
kschwab: Sure, will update.

docs/index.md
kschwab: I would lean towards JSON and Argparse, mostly to maintain consistency with JSON input and environment variable parsing. Of course, it's annoying to write `[...]` any time you need a list at the CLI, but it's pretty minor. 

However, from an implementation point of view we would have to go out of our way to exclude one of the styles. It's actually easier to support all three. The list parsing work was really dictated by the dict key=val list parsing. If we removed that, we could simplify the list parsing implementation.

docs/index.md
kschwab: Sure, will update.

kschwab:
@hramezani @samuelcolvin Thank you for the feedback, really appreciate it and will make updates accordingly. We found a couple of changes we'll also bring in after playing with it internally:

1. The enum support should be moved deeper into the environment variable parsing. Will open a separate issue and address it there.
2. We plan to integrate this with other existing parsers (one of which is pytest), so will be providing some support to "export" the internally generated parser and likely a pytest specific example. Still working on how to best accomplish this but will push once ready.

kschwab:
> 1. The [enum support](https://github.com/pydantic/pydantic-settings/pull/214/commits/cb9c1c36ff9973954ef04a880483eb796c3d0a7c) should be moved deeper into the environment variable parsing. Will open a separate issue and address it there.

Opened #252 to address this.

docs/index.md
kschwab: @hramezani @samuelcolvin 

I think we're at an impasse for the doc related tests on the `--help` output. I updated the examples to use a custom CLI source so the help text could be directly grabbed from the root parser. While somewhat cringey, it did allow for a real doc test on the `--help` output.

However, I had forgotten Python 3.9 -> 3.10 `argparse` changed the help text slightly, meaning we would also have to update the examples to account for the difference. This is ok in the [unittests](https://github.com/pydantic/pydantic-settings/blob/9679c107daf910ffca6ea769b4d1ed1eb47e1024/tests/test_settings.py#L2486), but for the help docs it will just further obfuscate the examples. Therefore, looking for guidance on how to best proceed given the below options:

1. Use pytest="skip" again for doc tests related to `--help` output?
2. Update the docs to break up the python code from the expected help text?
3. Other ideas?

kschwab: I skipped them as it is cleanest from docs perspective.

docs/index.md
kschwab: I ended removing all instances of `use_attribute_docstrings`. The user should know from Pydantic if they want to enable this or not.

docs/index.md
kschwab: I left all three options in as it is simplest given we support dict key=val list parsing.

docs/index.md
kschwab: @hramezani I removed these as I agree they can be found in API docs. How to link to the APIs so user can see above API details?

kschwab:
@hramezani @samuelcolvin @dmontagu updates are complete and ready for review.

pydantic_settings/sources.py
kschwab: @hramezani please resolve this comment once satisfied.

hramezani:
> @hramezani @samuelcolvin @dmontagu updates are complete and ready for review.

Thanks @kschwab for the update.
We will review it later and probably this will be included in the next release

tests/test_settings.py
```diff
@@ -25,6 +29,7 @@
     dataclasses as pydantic_dataclasses,
 )
 from pydantic.fields import FieldInfo
+from pydantic.v1.utils import Representation
```
hramezani, line 32: we shouldn't import it from `pydantic.v1`

kschwab: Sure, this was to cover [line 1308](https://github.com/pydantic/pydantic-settings/blob/0263f0d83711b4af7288d35d05dfaa73b80e9c1f/pydantic_settings/sources.py#L1308). The original [tests](https://github.com/pydantic/pydantic/blob/main/tests/test_utils.py#L63-L108) were not providing coverage (likely coming from a different case). We can:

1. Remove line 1308.
2. Ignore coverage for line 1308 and remove `pydantic.v1.utils.Representation`.
3. Find test case that covers line 1308 and remove `pydantic.v1.utils.Representation`.

Please let me know how you would like to proceed.

hramezani: I would say try to cover the line by test and if it is really not possible ignore it.
BTW, the v1 import has to be deleted

kschwab: hmm, not sure what changed but it passes now when using the internal Representation in the test. Updated.

hramezani:
@kschwab Sorry for the late response. Finally, I found time to work on this and test it.
The implementation is good and the doc is very well 👍 

There are something that need to be fixed:
* If you enable `SettingsConfigDict(cli_parse_args=True)`, then the CLI source will be the first priority and there is no way to change the priority. This is not like other sources that we've added before like `toml` and `yaml`. you include them and change the priority as well. So, this needs to be changed and we should have the same behavior as other new sources.

* It doesn't work with aliases. with the following model:
``` py
class Settings(BaseSettings):
    apple: str = Field(alias='alias')

```
we can't use `--alias foo`.
As alias support is included in other sources, we have to have it in CLI as well.

* It doesn't work with case_sensitive config and always expects the field name. With the following model:
```py
class Settings(BaseSettings):
    foo: str

    model_config = SettingsConfigDict(cli_parse_args=True, case_sensitive=False)  # default is `False`
```
Then `--FOO bar` doesn't work. in env source, we convert all the env names to lowercase in when `case_sensitive=False`

* It doesn't work well with dataclasses in sub model. for example with:
```py
@pydantic_dataclasses.dataclass
class MyDataclass:
    foo: int

class Settings(BaseSettings):
    n: MyDataclass
```
You have to pass --n '{"foo": 123} but --n.foo 123 doesn't work. but it works for normal basemodel submodels

Thanks for your great work here ❤️

kschwab:
Thanks @hramezani for the review, we'll address the feedback items soon.

With respect to the `CliSettingsSouce` prioritization, originally, we had prioritization that could be changed but removed it after our discussion [here](https://github.com/pydantic/pydantic-settings/pull/214/files/5795c7cf6973406fdf70d7983a73d8486236c364#r1463979911). The conclusion was to use fixed prioritization so as to not break backwards compatibility, and then revisit dynamic prioritization once [this issue](https://github.com/pydantic/pydantic-settings/issues/216) is fixed in V3.

We can change it back to the original if you would prefer, let me know your thoughts.

hramezani:
> Thanks @hramezani for the review, we'll address the feedback items soon.
> 
> With respect to the `CliSettingsSouce` prioritization, originally, we had prioritization that could be changed but removed it after our discussion [here](https://github.com/pydantic/pydantic-settings/pull/214/files/5795c7cf6973406fdf70d7983a73d8486236c364#r1463979911). The conclusion was to use fixed prioritization so as to not break backwards compatibility, and then revisit dynamic prioritization once [this issue](https://github.com/pydantic/pydantic-settings/issues/216) is fixed in V3.
> 
> We can change it back to the original if you would prefer, let me know your thoughts.

Yeah. that was my first thought. but after talking with the team we decided to have the CLI source like other sources.
Can't we implement the new CLI source like other new sources(toml, yaml) and include it like them? it would be good to 
Sorry about that.

kschwab:
No worries! 

> Can't we implement the new CLI source like other new sources(toml, yaml) and include it like them?

I haven't looked at the latest toml, yaml source changes and how it was integrated, but will try and align with it if possible. I don't think there should be any issues. The only thing that comes to mind is the external parser support, but I think it should be fine. We'll make the updates and see where we land 🙂

hramezani:
> @hramezani I've addressed the issues raised. For case-insensitive matching, this is not supported on external parsers, it is only supported on the internal parser.

Thanks @kschwab for addressing the issues. I will check it later.

samuelcolvin (approved):
Thanks so much @kschwab for working on this, and so sorry for my slow response 🙏 .

Overall, I think this is AWESOME, and I can't wait to use it.

I have one minor suggestion for an improvement from trying this out (can be fixed now or in future):

It would be nice if the `--help` output differentiated between required and not-required fields, and I guess showed the default value for non-required fields.

E.g. for this code:

```py
from devtools import debug


class Settings(BaseSettings, use_attribute_docstrings=True, cli_parse_args=True):
    """
    Settings for the application.
    """
    foo: str
    """Set the foo value."""
    bar: int = 123


s = Settings()
debug(s)
```

The help output was:

```
usage: demo.py [-h] [--foo str] [--bar int]

Settings for the application.

options:
  --foo str   Set the foo value. Required
  --bar int   Optional, default=123.
```

E.g. we append `' Required'` or `' Optional, default={default!r}'` to the description, or use that as the description if there is none.

kschwab:
Thanks @samuelcolvin for the review.

I added the differentiation with the last commit. Required fields are marked as `(required)` and optional fields are marked as `(default: val)`. This also aligns with argparse [ArgumentDefaultsHelpFormatter](https://docs.python.org/3/library/argparse.html#argparse.ArgumentDefaultsHelpFormatter) in how it delineates optional/default arguments.

```
usage: demo.py [-h] [--foo str] [--bar int]

Settings for the application.

options:
  --foo str   Set the foo value. (required)
  --bar int   (default: 123)
```

hramezani (approved):

hramezani:
@kschwab Thanks for your update.

I am going to merge the PR if you are done with your updates.

It will be released soon 🚀

hramezani:
Thanks @kschwab for your great effort here ❤️

kschwab:
Thanks @hramezani! Glad to see it go in 👍🏾

PR #217, opened 2024-01-24
Fix for JSON on optional nested types.

kschwab:
Resolves https://github.com/pydantic/pydantic-settings/issues/196

Updated to recursively search the annotation for any model that matches key. Note that there is a possibility that it could get the wrong field in case of union with complex types that both have the same key. Not sure how that case should be handled. For now, it picks first match.

### DIFF
--- a/tests/test_settings.py
+++ b/tests/test_settings.py
@@ -1233,7 +1233,7 @@ def test_secrets_nested_optional_json(tmp_path):
         a: int
 
     class Settings(BaseSettings):
-        foo: Foo | None = None
+        foo: Optional[Foo] = None
 
     p = tmp_path / 'foo'
     p.write_text('{"a": 10}')
--- a/tests/test_settings.py
+++ b/tests/test_settings.py
@@ -1229,16 +1229,18 @@ def test_secrets_path_json(tmp_path):
 
 
 def test_secrets_nested_optional_json(tmp_path):
+    p = tmp_path / 'foo'
+    p.write_text('{"a": 10}')
+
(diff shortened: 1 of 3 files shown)


hramezani:
Thanks @kschwab for this PR 🙏 

The test case that you've added is great. Could you please add another test case based on the code snippet in reported issue?

kschwab:
@hramezani good catch. #196 was slightly different issue compared to mine. I have pushed a fix and added test.

tests/test_settings.py
```diff
@@ -1212,6 +1228,19 @@ class Settings(BaseSettings):
...
+    class Foo(BaseModel):
+        a: int
+
+    class Settings(BaseSettings):
+        foo: Optional[Foo] = None
+
+    p = tmp_path / 'foo'
+    p.write_text('{"a": 10}')
+
+    assert Settings(_secrets_dir=tmp_path).model_dump() == {'foo': {'a': 10}}
```
hramezani, line 1241:
```suggestion
    p = tmp_path / 'foo'
    p.write_text('{"a": 10}')

    class Foo(BaseModel):
        a: int

    class Settings(BaseSettings):
        foo: Optional[Foo] = None

        model_config = SettingsConfigDict(secrets_dir=tmp_path)

    assert Settings().model_dump() == {'foo': {'a': 10}}
```

hramezani (approved):

hramezani:
Thanks @kschwab for this. I've just updated the tests.

kschwab:
Thanks @hramezani!

PR #218, opened 2024-01-25
Ensure extra='forbid' is enforced in DotEnvSettingsSource when env_prefix is specified

### LINKED ISSUE #215: env_prefix prevents env_file from working
### Initial Checks

- [X] I confirm that I'm using Pydantic V2

### Description

Assuming this is a bug because I found nothing in the docs explaining this behavior.

Spent forever trying to understand why env_file wasn't working. Commenting out env_prefix cause env_file to start working. This can't be by design right? I should be able to define both.

### Example Code

```Python
        env_file=[
            "my.env",
        ],
        # When uncommented pydantic does not attempt to load any files defined in env_file 
        # env_prefix="my_app_",
    )
```

### Python, Pydantic & OS Version

```Text
pydantic version: 2.5.3
        pydantic-core version: 2.14.6
          pydantic-core build: profile=release pgo=true
                 install path: /usr/local/lib/python3.12/site-packages/pydantic
               python version: 3.12.1 (main, Dec 19 2023, 20:23:36) [GCC 10.2.1 20210110]
                     platform: Linux-6.5.11-linuxkit-x86_64-with-glibc2.31
             related packages: pydantic-settings-2.1.0 mypy-1.8.0 typing_extensions-4.9.0
```

Edit
===
I found the issue. The issue is that setting `env_prefix` causes pydantic to ignore non-prefixed env vars defined in you dotenv files.

I was expecting it to fail because I had un-prefixed vars defined in mysettings.env for example:

    MY_SETTING_ONE=ksjdlksjd
    MY_SETTING_TWO=zzzzzzz

Changing those to this worked:

    MY_APP_MY_SETTING_ONE=ksjdlksjd
    MY_APP_MY_SETTING_TWO=zzzzzzz

So I think the bug is pydantic should not be ignoring un-prefixed fields in dotenv files.

Because while this correctly throws `Extra inputs are not permitted`:

    MY_APP_SETTING_THAT_DOESNT_EXIST=ksjdlksjd

This throws no errors at all:

    SETTING_THAT_DOESNT_EXIST=ksjdlksjd

kjithin:
Fix #215 

The current implementation in DotEnvSettingsSource allows non-prefixed variables even when extra='forbid'(default) is set . This commit addresses the issue by raising a SettingsError when extra is set to 'forbid' and variables are encountered without using the specified prefix.

### DIFF
--- a/pydantic_settings/sources.py
+++ b/pydantic_settings/sources.py
@@ -647,7 +647,10 @@ class DotEnvSettingsSource(EnvSettingsSource):
         # update data with extra env variabels from dotenv file.
         for env_name, env_value in self.env_vars.items():
             if not is_extra_allowed and not env_name.startswith(self.env_prefix):
-                raise SettingsError(f'unable to load environment variables from \'{self.env_file}\' file due to the presence of variables without the specified prefix - \'{self.env_prefix}\'')
+                raise SettingsError(
+                    f"unable to load environment variables from '{self.env_file}' file "
+                    f"due to the presence of variables without the specified prefix - '{self.env_prefix}'"
+                )
             if env_name.startswith(self.env_prefix) and env_value is not None:
                 env_name_without_prefix = env_name[self.env_prefix_len :]
                 first_key, *_ = env_name_without_prefix.split(self.env_nested_delimiter)
--- a/pydantic_settings/sources.py
+++ b/pydantic_settings/sources.py
@@ -640,12 +640,14 @@ class DotEnvSettingsSource(EnvSettingsSource):
         data: dict[str, Any] = super().__call__()
 
         data_lower_keys: list[str] = []
+        is_extra_allowed = self.config.get('extra') != 'forbid'
         if not self.case_sensitive:
             data_lower_keys = [x.lower() for x in data.keys()]
-
         # As `extra` config is allowed in dotenv settings source, We have to
         # update data with extra env variabels from dotenv file.
         for env_name, env_value in self.env_vars.items():
+            if not is_extra_allowed and not env_name.startswith(self.env_prefix):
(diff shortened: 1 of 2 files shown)


kjithin:
please review

hramezani:
Thanks @kjithin for working on this 🙏 

I think raising `SettingsError` is not the right approach. because it is not consistent with the dotenv source without `env_prefix`.

e.g. if we have a settings model like:

```py


class Foo(BaseSettings):
    a: str

    model_config = SettingsConfigDict(env_file="my.env")

```
with `my.env`:

```
a=foo
b=bar
```

we will get
```
pydantic_core._pydantic_core.ValidationError: 1 validation error for Foo
b
  Extra inputs are not permitted [type=extra_forbidden, input_value='bar', input_type=str]
    For further information visit https://errors.pydantic.dev/2.5/v/extra_forbidden
```

So, we want to have the same behavior in the case of the user providing `env_prefix` as well.
it can probably fixed by touching the next block when we only collect envs that start with `env_prefix`. we need to collect all of them.

hramezani:
please update

kjithin:
Got it. Initially, I explored that option. However, when collecting all the environment variables, a challenge arose: it would accept a valid property without a prefix. This occurred because the expected output for this function is a dictionary without prefixes in the keys.

for e.g.

```python


class Foo(BaseSettings):
    a: str
    b: str

    model_config = SettingsConfigDict(env_file="my.env", env_prefix="prefix_")
```

with `my.env`

```
prefix_a=foo
b=bar
```

The identical ValidationError will only be triggered during the final model validation against the generated dictionary, specifically if it includes any additional properties. In the above case, when collecting all the env variables, both `a` & `b` properties will be accepted without any issue as they are valid property names.

hramezani (approved):

Issue #220, opened 2024-01-26
env_prefix applies to fields with alias when defined in dotenv file


gsakkis:
Follow up to https://github.com/pydantic/pydantic-settings/issues/219, probably with the same root cause.

According to the docs `env_prefix does not apply to fields with alias`. This is not true when it comes to variables in dotenv files. 

MCVE:
```py
from pydantic import Field
from pydantic_settings import BaseSettings, SettingsConfigDict


class Settings(BaseSettings):
    model_config = SettingsConfigDict(env_prefix="xxx__", env_file=".env")

    foo: str = ""
    bar_alias: str = Field("", alias="bar")


print(Settings().model_dump())
```

Populating the fields via env vars works as advertised; `env_prefix` applies to the non-alias field but not to the alias:
```console
$ xxx__foo=1 xxx__bar=2 python testsettings.py
{'foo': '1', 'bar_alias': ''}
```

Populating them via a dotenv file applies `env_prefix` to both fields:

```console
cat .env
xxx__foo=1
xxx__bar=2

$ python testsettings.py
{'foo': '1', 'bar_alias': '2'}
```

hramezani:
Thanks @gsakkis for reporting this 🙏 

My fix for https://github.com/pydantic/pydantic-settings/issues/219 probably will fix this issue as well.
I try to add a test case for this issue as well

hramezani:
I've created https://github.com/pydantic/pydantic-settings/pull/221 to fix it. it would be great to take a look or test it if you want

> Please add a test case or two where the aliased field is nested settings. I'm seeing discrepancies between env vars and dotenv for this too but it's probably the same cause so I didn't open yet another issue.

Please test those cases as well or provide an example then I will add them to the PR

gsakkis:
Thanks for the quick turnaround! As I suspected, unfortunately it doesn't work with nested settings.
MCVE:
```py
from pydantic import Field
from pydantic_settings import BaseSettings, SettingsConfigDict


class NestedSettings(BaseSettings):
    a: str = "a"
    b: str = "b"


class Settings(BaseSettings):
    model_config = SettingsConfigDict(env_prefix="xxx__", env_nested_delimiter="__", env_file="tmp.env")

    foo: str = ""
    bar_alias: str = Field("", alias="xxx__bar")
    nested_alias: NestedSettings = Field(default_factory=NestedSettings, alias="xxx__nested")


print(Settings().model_dump())
```

```console
$ cat tmp.env
[...]

Issue #219, opened 2024-01-25
ValidationError when combining dotenv file, env_prefix and fields with alias


gsakkis:
I have the following settings setup:
- a non empty env prefix
- one or more fields with alias that starts with the env prefix.

MCVE:
```py
from pydantic import Field
from pydantic_settings import BaseSettings, SettingsConfigDict


class Settings(BaseSettings):
    model_config = SettingsConfigDict(env_prefix="xxx__", env_file=".env")

    foo: str = ""
    bar_alias: str = Field("", alias="xxx__bar")


print(Settings().model_dump())
```

Populating the fields via env vars works:
```console
$ xxx__foo=1 xxx__bar=2 python testsettings.py
{'foo': '1', 'bar_alias': '2'}
```

Populating them via a dotenv file raises ValidationError:
```console
$ cat .env
xxx__foo=1
xxx__bar=2

$  python testsettings.py
Traceback (most recent call last):
...
pydantic_core._pydantic_core.ValidationError: 1 validation error for Settings
bar
  Extra inputs are not permitted [type=extra_forbidden, input_value='2', input_type=str]
    For further information visit https://errors.pydantic.dev/2.5/v/extra_forbidden
```

hramezani:
Thanks @gsakkis for reporting this.

dotenv setting source consider `extra` config. So

> Pydantic settings consider extra config in case of dotenv file. It means if you set the extra=forbid (default) on model_config and your dotenv file contains an entry for a field that is not defined in settings model, it will raise ValidationError in settings construction.

This is not the case in env source because it will ignore the extra input.

You can ignore this behavior by adding `extra='ignore'` to your config

gsakkis:
Thanks @hramezani for the quick reply! I'm not sure how the quoted docs are relevant here; `xxx__bar` does refer to a field defined in the model, it's not an `extra` I want to ignore.

hramezani:
Sorry @gsakkis for misunderstanding. Now, I got your point.

I will prepare a patch to fix this bug

hramezani:
I've created https://github.com/pydantic/pydantic-settings/pull/221 to fix it. it would be great to take a look or test it if you want

samuelcolvin:
@gsakkis any chance you could review @hramezani's PR.

Otherwise, we'll merge and release next week.

PR #221, opened 2024-01-26
Fix dotenv source extra values parsing provided in dotenv file

hramezani:
Fiexes https://github.com/pydantic/pydantic-settings/issues/219 and https://github.com/pydantic/pydantic-settings/issues/220


samuelcolvin (approved):
LGTM.

But ideally it would be great if @gsakkis could review this and confirm it fixes their problem?

Issue #223, opened 2024-01-26
How to build a cascading settings resolver


trondhindenes:
In our current config module we have the following logic:

- Determine some basic info about the app such as app name, and environment name (from environment variables)
- Then use these to optionally look up data from a settings store such as aws ssm parameter store or similar

This is done by invoking a series of config providers, from lowest-prioritoy up to highest. Each variable value can at any time be overwritten by the value from a higher-priority provider.

It seems to me that pydantic doesn't support this, because it seems to read top-down starting from the highest-priority provider. In my example I guess AWS SSM would be read first, and then any settings left unresolved would be left to the environment variable provider. That wouldn't work for us, since we need the env var provided-values to build the paths etc for aws ssm.

So I guess I'm asking:
Is it possible with pydantic-settings to build a hierarchy of "config sources" where one source relies on data provided by another?

mnbbrown:
I was exploring this earlier this week.. from my grep of the codebase it's not possible currently because `__call__` in `PydanticBaseSettingsSource` doesn't get the current state of the settings model as an argument. See https://github.com/pydantic/pydantic-settings/blob/main/pydantic_settings/main.py#L181

If the current model was passed in the source itself could:

a. retrieve the current value of attributes to help build new attributes
b. decided if it should overwrite or not.

[This PR](https://github.com/pydantic/pydantic-settings/pull/224) is an imperfect solution to this problem but goes some way to addressing it. Imperfect because the `current_state` is untyped and it's a breaking change.

hramezani:
Thanks @trondhindenes for reporting this issue.

Yes, as @mnbbrown described it is not possible right now.

We have to postpone it to V3 because it will introduce a breaking change.
I will mark it as a feature request and we can include this in V3

hramezani:
@mnbbrown @trondhindenes what do you think about https://github.com/pydantic/pydantic-settings/pull/326?

Does this PR solve your problems?

trondhindenes:
I believe it does, thank you!

PR #224, opened 2024-01-27
Pass current_state in when calling custom and built-in sources.

mnbbrown:
This will allow sources to dynamically decided if to provide a value for the attribute.
For example, a higher priority source can decided to leave the low-priority source "as is" if it's not None.

Note: I think this is considered a breaking change as the signature of `__call__` on the source abstract type is changing.

See https://github.com/pydantic/pydantic-settings/issues/223#issuecomment-1913196648 for context
Interested in the maintainers thoughts here :)

### DIFF
--- a/pydantic_settings/sources.py
+++ b/pydantic_settings/sources.py
@@ -107,7 +107,7 @@ class PydanticBaseSettingsSource(ABC):
         return json.loads(value)
 
     @abstractmethod
-    def __call__(self) -> dict[str, Any]:
+    def __call__(self, current_state: dict[Any, str] = {}) -> dict[str, Any]:
         pass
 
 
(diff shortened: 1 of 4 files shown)


pydantic_settings/sources.py
```diff
@@ -107,7 +107,7 @@ def decode_complex_value(self, field_name: str, field: FieldInfo, value: Any) ->
         return json.loads(value)
 
     @abstractmethod
-    def __call__(self) -> dict[str, Any]:
+    def __call__(self, current_state: dict[Any, str] = {}) -> dict[str, Any]:
```
hramezani, line 110:
```suggestion
    def __call__(self, current_state: dict[str, Any] = {}) -> dict[str, Any]:
```
Same for other places

hramezani:
Thanks @mnbbrown for this PR.

we need some tests for this and have to wait to merge it before `V3` because it is a breaking change.

hramezani:
Superseded by https://github.com/pydantic/pydantic-settings/pull/326

PR #229, opened 2024-02-12
Adding optional prefixes

kassett:
There is a common use-case where several services use the same env vars -- i.e. an ENVIRONMENT variable. Occasionally, you want to override this value per service, but do not want to enforce it for every single variable for every single service. I have added an optional argument ``prefix_optional``, wherein the prefixed version of an environment variable will be checked first, but will then fallback to the naked version.

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -184,6 +184,21 @@ class Settings(BaseSettings):
 
 1. `env_prefix` will be ignored and the value will be read from `FooAlias` environment variable.
 
+#### Optional Environment Variable Prefixes
+
+```py
+from pydantic_settings import BaseSettings, SettingsConfigDict
+
+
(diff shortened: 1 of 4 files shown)


kassett:
@PrettyWood Could you take a look at this?

PrettyWood:
Hello @kassett and thanks for opening a PR
We first need an issue to discuss the problem and agree on the expected API before opening a PR and reviewing the code.
For example what if I want a 3rd possible value for a given var? And prioritize one or another?
We could have a list of aliases and check in order the presence of a given env var.

kassett:
@PrettyWood I opened an issue here: https://github.com/pydantic/pydantic-settings/issues/231

hramezani:
closed based on discussion on issue

Issue #235, opened 2024-02-17
We should use tomli not tomlkit


samuelcolvin:
`tomli` is the default.

My mistake not picking this up when reviewing #211.

We can still fall back to tomlkit.

samuelcolvin:
Cc @hramezani.

PR #236, opened 2024-02-18
Replace tomlkit with tomli and remove toml_file_encoding config

hramezani:
Fixes https://github.com/pydantic/pydantic-settings/issues/235

- Replace `tomlkit` with `tomli`
- Remove `remove toml_file_encoding` config because file open in binary mode doesn't take an encoding argument


samuelcolvin (commented):
LGTM, only question is whether we should fall back to tomlkit if tomli is not installed?

hramezani:
> LGTM, only question is whether we should fall back to tomlkit if tomli is not installed?

I think we don't need to fall back on tomlkit

PR #255, opened 2024-03-13
add `PyprojectTomlConfigSettingsSource`

### LINKED ISSUE #253: `PyprojectTomlConfigSettingsSource` or extend `TomlConfigSettingsSource`
I have a need to accept user provided settings from either a file or environment variable. I personally like to limit the number of root-level files in my projects so I would like to provide the ability for my users to do that as well by enabling the use of a `[tool.*]` table for specifying settings. Looking through the current implementation of `TomlConfigSettingsSource`, there does not appear to be an easy way to do this.

I would like to propose one of two approaches to add this functionality as I am likely not the only one that would benefit from it:

1. Extend `TomlConfigSettingsSource` (and it's configuration) to enable specifying a table within the TOML file to load data from.
2. Create `PyprojectTomlConfigSettingsSource` (from `TomlConfigSettingsSource`) as it's own source with the functionality described above. Additionally, it could default to using the `pyproject.toml` file in the current working directory (or parent directory if not found) by default, if the source is enabled. Rational behind trying the parent directory relates to another tool I use, poetry. However, it will go beyond just the direct parent.

A case could also be made for both to be implemented - extending `TomlConfigSettingsSource` so that the table selection is more widely available and `PyprojectTomlConfigSettingsSource` for some file discovery be default.

----

I have a PoC of `PyprojectTomlConfigSettingsSource` that I am currently using. I'm open to contributing this if desired and the approach I took the way the maintainers would like to go. I'm just not starting there to conserve time.

```python
from __future__ import annotations

from pathlib import Path
from typing import TYPE_CHECKING

    BaseSettings,
    TomlConfigSettingsSource,
)
    SettingsConfigDict as _SettingsConfigDict,
)

if TYPE_CHECKING:
    from pydantic_settings import PydanticBaseSettingsSource


class SettingsConfigDict(_SettingsConfigDict, total=False):
    """Overrides ``SettingsConfigDict`` to add ``pyproject.toml`` settings."""

    pyproject_toml_path: Path | str
    """Explicit path to a ``pyproject.toml`` file.

    If not provided, will attempt to load the ``pyproject.toml`` file located in the
    current working directory. If not found, the parent directory will be tried.
    Only one level higher will be tried.

    """

    toml_table_path: tuple[str, ...]
    """Path to the table to load.

    .. rubric:: Example
    .. code-block:: toml
        :caption: pyproject.toml

        [tool.poetry]

    .. code-block:: python

        SettingsConfigDict(toml_table_path=("tool", "poetry"))

    """


class PyprojectTomlConfigSettingsSource(TomlConfigSettingsSource):
    """``pyproject.toml`` config settings source."""

    def __init__(
        self,
        toml_file: Path | None = None,
    ) -> None:
        """Instantiate class."""
        self.table_path: tuple[str, ...] = settings_cls.model_config.get(
            "toml_table_path", ()
        )
        self.toml_file_path = self._pick_pyproject_toml_file(
            toml_file or settings_cls.model_config.get("pyproject_toml_path")
        )
        self.toml_data = self._read_files(self.toml_file_path)
        for key in self.table_path:
            self.toml_data = self.toml_data.get(key, {})
        super(TomlConfigSettingsSource, self).__init__(settings_cls, self.toml_data)

    @staticmethod
    def _pick_pyproject_toml_file(provided: Path | None) -> Path:
        """Pick a pyproject.toml file path to use."""
        if provided:
            return provided.resolve()
        rv = Path.cwd() / "pyproject.toml"
        if not rv.is_file():
            rv = rv.parent.parent / "pyproject.toml"
        return rv


class Settings(BaseSettings):
    """Settings."""

    model_config = SettingsConfigDict(toml_table_path=("tool", "poetry"))

    name: str | None = None

    @classmethod
        cls,
        file_secret_settings: PydanticBaseSettingsSource,  # noqa: ARG003
        """Example override to enable ``PyprojectTomlConfigSettingsSource``."""
        return (
            env_settings,
            dotenv_settings,
            PyprojectTomlConfigSettingsSource(settings_cls),
            init_settings,
        )


if __name__ == "__main__":
    print(Settings.model_validate({}))
```

ITProKyle:
> PR template copied from https://github.com/pydantic/pydantic/blob/18d39febf00cb07954e19ebbdbd8168ef20e3df1/.github/PULL_REQUEST_TEMPLATE.md as this repo does not contain one.

## Change Summary

- added a `PyprojectTomlConfigSettingsSource` source class
  - finds and uses a `pyproject.toml` file to fill variables
  - optionally use the values from a specific table
 - added `SettingsConfigDict.pyproject_toml_depth` (only used by `PyprojectTomlConfigSettingsSource`)
   - specify a number of levels **up** the directory tree to look for a `pyproject.toml` if not found in the current working directory
   - if not provided, `0` is used
   - if no file is found, path in current working directory is used
- added `SettingsConfigDict.pyproject_toml_table_header` (only used by `PyprojectTomlConfigSettingsSource`)
  - used to optionally select a table to fill values from
  - if not provided the root table is used
  - _header_ was used over _path_ (mention in the feature request) after reviewing the terminology used in [TOML v1.0.0](https://toml.io/en/v1.0.0#table)

## Related issue number

resolves #253

## Checklist

* [x] The pull request title is a good summary of the changes - it will be used in the changelog
* [x] Unit tests for the changes exist
* [x] Tests pass on CI
* [x] Documentation reflects the changes where applicable
* [x] My PR is ready to review


hramezani:
Thanks @ITProKyle for this PR. I still haven't looked at it in detail.

> added SettingsConfigDict.pyproject_toml_depth (only used by PyprojectTomlConfigSettingsSource)
specify a number of levels up the directory tree to look for a pyproject.toml if not found in the current working directory
if not provided, 0 is used
if no file is found, path in current working directory is used

Isn't it better to have a `pyproject_toml_file` config to specify the path of `pyproject.toml` file. the default value could be the `pyproject.toml`.
I think the `pyproject_toml_depth` config is not needed as we can use the `pyproject_toml_file`

> added SettingsConfigDict.toml_table_header (only used by PyprojectTomlConfigSettingsSource)
used to optionally select a table to fill values from
if not provided the root table is used
header was used over path (mention in the feature request) after reviewing the terminology used in [TOML v1.0.0](https://toml.io/en/v1.0.0#table)

Let's have a default header. something like `tools.pydantic-settings`

hramezani:
> I changed toml_table_header to pyproject_toml_table_header for consistency of scoped configs. This would also make it possible to add support for toml_table_header to the TOML source if desired for loading global settings (e.g. from ~/.conf/pydantic-settings/config.toml) that does not need to follow the same nested table structure.

> pyproject_toml_table_header defaults to using the [tool.pydantic-settings] table.

👍 

> The rationale behind supporting discovery is to mirror the behavior of most tools (I have used) that use pyproject.toml for configuration - accept an explicit path and try to find the file if that path is not provided.

Can we have the same behavior here as well?(I still haven't checked the code. So, if you implemented like this it would be fine)

docs/index.md
```diff
@@ -592,6 +593,66 @@ foobar = "Hello"
...
+        settings_cls: Type[BaseSettings],
+        init_settings: PydanticBaseSettingsSource,
+        env_settings: PydanticBaseSettingsSource,
+        dotenv_settings: PydanticBaseSettingsSource,
+        file_secret_settings: PydanticBaseSettingsSource,
+    ) -> Tuple[PydanticBaseSettingsSource, ...]:
+        return (PyprojectTomlConfigSettingsSource(settings_cls),)
+
+
+class RootSettings(Settings):
```
hramezani, line 635: Why did you define the `RootSettings`?

docs/index.md
```diff
@@ -592,6 +593,66 @@ foobar = "Hello"
...
+    model_config = SettingsConfigDict(extra='ignore')
+```
+
+This will be able to read the following "pyproject.toml" file, located in your working directory, resulting in `Settings(field='some-table')` & `RootSettings(field='root')`:
```
hramezani, line 639: `RootSettings` also will be `RootSettings(field='some-table')`

docs/index.md
```diff
@@ -592,6 +593,66 @@ foobar = "Hello"
...
+[tool.some-table]
+field = "some-table"
+```
+
+By default, `PyprojectTomlConfigSettingsSource` will only look for a "pyproject.toml" in the your current working directory.
+However, there are two options to change this behavior.
+
+* `SettingsConfigDict(pyproject_toml_depth=<int>)` can be provided to check `<int>` number of directories **up** in the directory tree for a "pyproject.toml" if one is not found in the current working directory.
+  By default, no parent directories are checked.
+* An explicit file path can be provided to the source when it is instantiated (e.g. `PyprojectTomlConfigSettingsSource(Path('~/.config').resolve() /'pyproject.toml')`).
```
hramezani, line 653: It would be great to add a complete example here to show how to use a custom path for `pyproject.toml`

tests/test_sources.py
```diff
@@ -0,0 +1,99 @@
+"""Test pydantic_settings.sources."""
```
hramezani, line 1: I would suggest moving the important tests here to the `test_settings` file as a complete test case.
for example:

- Test for a custom `pyproject.toml` file
- Test for different depth

hramezani:
Thanks @ITProKyle for updating the PR. Overall LGTM. I left a couple of comments.

Please update

docs/index.md
ITProKyle: Looks like I forgot to update the example code after adding a default. This is to show how to load from different locations. Updating this in an incoming commit plus adding a third example to show another use case.

- `Settings` for using the default table
- `SomeTableSettings` replaces what `Settings` was doing before, providing a `pyproject_toml_table_header` config value with a different table header
- `RootSettings` updated to work as I had it before setting a default header, loading from the root of pyproject.toml

docs/index.md
ITProKyle: Same as previous comment, I overlooked updating this. A fix is prepped in an incoming commit.

docs/index.md
ITProKyle: added

tests/test_sources.py
ITProKyle: I added some tests to `test_settings.py` for additional coverage. However, I added these test to `test_sources.py` to make assertions at the source level that arn't possible (from what I can tell) at the `BaseSettings` subclass level (e.g. asserting that the correct file path is chosen). This can be inferred from the settings level by doing things like only creating a single file with the expected value but it's not quite as precise.

I'll leave this file in place for now with the additional tests added to `test_settings.py`. If you feel they are still not necessary, feel free to delete the file or LMK and I can do it alongside any additional changes requested.

hramezani (approved):

hramezani:
Thanks @ITProKyle

PR #269, opened 2024-04-12
Fix typing syntax for compatibility with python 3.8

NellyWhads:
Typing issues were brought to light when attempting to override `BaseSettings.settings_customise_sources()` as per [the documentation](https://docs.pydantic.dev/latest/concepts/pydantic_settings/#parsing-environment-variable-values:~:text=You%20may%20also%20populate%20a%20complex%20type%20by%20providing%20your%20own%20source%20class.).

On python 3.8, this raised the `TypeError: 'type' object is not subscriptable` exception from [this line](https://github.com/pydantic/pydantic-settings/blob/c2fd92ffa74b14ed38fd8fcb1f77d087ca4a960e/pydantic_settings/main.py#L125)

Is this the only way forward? Or am I missing something and hitting an edge case with python 3.8 and `type`?

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -18,7 +18,7 @@ jobs:
     - name: set up python
       uses: actions/setup-python@v4
       with:
-        python-version: '3.10'
+        python-version: '3.8'
 
     - run: pip install -r requirements/linting.txt -r requirements/pyproject.txt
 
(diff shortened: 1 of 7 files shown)


NellyWhads:
I can't seem to get the linter to respect python 3.8 rules - I don't know why. Hm.

NellyWhads:
I made a derp. Fixed the issue on my end instead.

If there's any interest in this, I'll keep it here, else it can be forgotten.

NellyWhads:
Closing due to lack of interest.

PR #272, opened 2024-04-21
Add Azure Key Vault settings source

### LINKED ISSUE #143: Microsoft KeyVault fetch support
### Initial Checks

- [X] I have searched Google & GitHub for similar requests and couldn't find anything
- [X] I have read and followed [the docs](https://docs.pydantic.dev) and still think this feature is missing

### Description

Hi 

Currently, we use the Pydantic with a wrapper that checks if the field exists (in addition to the env var) in the KeyVault.

This is not ideal as it is done in a separate place. 

A new feature of fetching data from KeyVault will be very helpful.

Thanks! 

### Affected Components

- [ ] [Compatibility between releases](https://docs.pydantic.dev/changelog/)
- [X] [Data validation/parsing](https://docs.pydantic.dev/usage/models/#basic-model-usage)
- [ ] [Data serialization](https://docs.pydantic.dev/usage/exporting_models/) - `.model_dump()` and `.model_dump_json()`
- [ ] [JSON Schema](https://docs.pydantic.dev/usage/schema/)
- [ ] [Dataclasses](https://docs.pydantic.dev/usage/dataclasses/)
- [ ] [Model Config](https://docs.pydantic.dev/usage/model_config/)
- [ ] [Field Types](https://docs.pydantic.dev/usage/types/) - adding or changing a particular data type
- [ ] [Function validation decorator](https://docs.pydantic.dev/usage/validation_decorator/)
- [ ] [Generic Models](https://docs.pydantic.dev/usage/models/#generic-models)
- [ ] [Other Model behaviour](https://docs.pydantic.dev/usage/models/) - `model_construct()`, pickling, private attributes, ORM mode
- [ ] [Plugins](https://docs.pydantic.dev/) and integration with other tools - mypy, FastAPI, python-devtools, Hypothesis, VS Code, PyCharm, etc.

Selected Assignee: @Kludex

Selected Assignee: @samuelcolvin

AndreuCodina:
Add Azure Key Vault settings source.

**Doubt:** Where should I add the required Python packages (`azure-keyvault-secrets==4.8.0` and `azure-identity==1.16.0`) to use `AzureKeyVaultSettingsSource`?

This is the infrastructure:

```terraform
resource "azurerm_resource_group" "pydantic" {
  name     = "pydantic"
  location = "northeurope"
}

resource "azurerm_key_vault" "pydantic" {
  name                          = "kv-pydantic"
  resource_group_name           = "pydantic"
  location                      = "northeurope"
  tenant_id                     = data.azuread_client_config.current.tenant_id
  sku_name                      = "standard"
  soft_delete_retention_days    = 7
  purge_protection_enabled      = false
  enable_rbac_authorization     = true
  public_network_access_enabled = true
}

resource "azurerm_key_vault_secret" "sql_server_password" {
  name         = "SqlServer--Password"
  value        = "SqlServerPassword"
  key_vault_id = azurerm_key_vault.pydantic.id
}

resource "azurerm_key_vault_secret" "my_password" {
  name         = "MyPassword"
  value        = "MyPassword"
  key_vault_id = azurerm_key_vault.pydantic.id
}
```

Closes #143


hramezani:
Thanks @AndreuCodina for this PR.

> Doubt: Where should I add the required Python packages (azure-keyvault-secrets==4.8.0 and azure-identity==1.16.0) to use AzureKeyVaultSettingsSource?

It has to be added to [optional dependencies section](https://github.com/pydantic/pydantic-settings/blob/c5c52e96b743fedecfcbb4f06520fd892f94419c/pyproject.toml#L48)

AndreuCodina:
> Thanks @AndreuCodina for this PR.
> 
> > Doubt: Where should I add the required Python packages (azure-keyvault-secrets==4.8.0 and azure-identity==1.16.0) to use AzureKeyVaultSettingsSource?
> 
> It has to be added to [optional dependencies section](https://github.com/pydantic/pydantic-settings/blob/c5c52e96b743fedecfcbb4f06520fd892f94419c/pyproject.toml#L48)
> 
> Also, I can't see any test added here. You need to add some tests

How do I make the imports in code? I've created the function `import_azure_key_vault`, but I'm getting `mypy pydantic_settings
pydantic_settings/sources.py:14: error: Cannot find implementation or library stub for module named "azure.core.credentials"  [import-not-found]`

hramezani:
You need to update the requirements by running `make refresh-lockfiles` command

AndreuCodina:
> You need to update the requirements by running `make refresh-lockfiles` command

```bash
$ make refresh-lockfiles
`make: *** No rule to make target 'refresh-lockfiles'.  Stop.`
```

hramezani:
Please refresh your fork. this command was added recently

requirements/linting.txt
```diff
@@ -1,10 +1,10 @@
 #
-# This file is autogenerated by pip-compile with Python 3.8
```
hramezani, line 2: Please run the make command with `python 3.8`

pydantic_settings/sources.py
```diff
@@ -877,6 +892,49 @@ def _read_file(self, file_path: Path) -> dict[str, Any]:
...
+    ) -> None:
+        import_azure_key_vault()
+        super().__init__(settings_cls)
+        self._secret_client = SecretClient(vault_url=url, credential=credential)  # type: ignore
+
+    def get_field_value(self, field: FieldInfo, field_name: str) -> tuple[Any, str, bool]:
+        field_value: Any | None = None
+
+        # It's not possible to use underscores in Azure Key Vault
+        secret_name = field_name.replace('_', '-')
```
hramezani, line 912: I think you need to use [_extract_field_info](https://github.com/pydantic/pydantic-settings/blob/09d10094d97425853829c2971051a561ccd4a2d2/pydantic_settings/sources.py#L199) to respect the field alias if defined.

pydantic_settings/sources.py
```diff
@@ -877,6 +892,49 @@ def _read_file(self, file_path: Path) -> dict[str, Any]:
...
+
+        try:
+            secret = self._secret_client.get_secret(secret_name)
+            field_value = secret.value
+        except ResourceNotFoundError:  # type: ignore
+            field_value = None
+
+        return field_value, field_name, False
+
+    def prepare_field_value(self, field_name: str, field: FieldInfo, value: Any, value_is_complex: bool) -> Any:
```
hramezani, line 922: Do you need to return the fetched value? what about the complex values(like json strings or nested models values).

Probably you don't need to override this method and use the parent one

requirements/pyproject.txt
```diff
@@ -1,23 +1,73 @@
 #
-# This file is autogenerated by pip-compile with Python 3.8
+# This file is autogenerated by pip-compile with Python 3.12
```
hramezani, line 2: same as above

requirements/testing.txt
```diff
@@ -1,17 +1,15 @@
 #
-# This file is autogenerated by pip-compile with Python 3.8
+# This file is autogenerated by pip-compile with Python 3.12
```
hramezani, line 2: same as above

tests/test_sources.py
```diff
@@ -97,3 +111,50 @@ def test___init___parent(self, mocker: MockerFixture, tmp_path: Path) -> None:
...
+        class AzureKeyVaultSettings(BaseSettings):
+            """AzureKeyVault settings."""
+
+        key_vault_secret = KeyVaultSecret(SecretProperties(), 'SecretValue')
+        mocker.patch(f'{MODULE}.SecretClient.get_secret', return_value=key_vault_secret)
+        obj = AzureKeyVaultSettingsSource(
+            AzureKeyVaultSettings, 'https://my-resource.vault.azure.net/', DefaultAzureCredential()
+        )
+
+        obj.get_field_value(field=FieldInfo(), field_name='sqlserverpassword')
```
hramezani, line 142: Don't we need any assertion here?

AndreuCodina: The method is just a mocked HTTP call to an API. I think I can't test anything more. The reliable test is the Jupyter notebook I'm executing against my Azure infrastructure.

pydantic_settings/sources.py
AndreuCodina: Overrided method removed.

pydantic_settings/sources.py
AndreuCodina: This is complex and I don't see the point in aliasing secret names. Can we iterate in the next version?

hramezani: This is the basic functionality that we already have in all of the source classes.

You need to move the `_extract_field_info` to `PydanticBaseSettingsSource` class and then call it here. it returns a list of tuples that contain `field_key`, `env_name` and `value_is_complex`. you can use the `env_name` to fetch the value from key vault . there are some use-cases of `_extract_field_info` in the code, you can check them as well.

tests/test_sources.py
hramezani: You can test the returned value from the mocked method.

I think it would be good to have a complete test like other tests. I mean having a settings model that uses the `AzureKeyVaultSettingsSource`.



```py
def test_azure_key_vault_settings_source():
        my_password: str
        sql_server__password: str
        @classmethod
            cls,
            return (
                AzureKeyVaultSettingsSource(
                    settings_cls, os.environ['KEY_VAULT__URL'], DefaultAzureCredential()
                ),
            )

    with mocker.patch(f'{MODULE}.SecretClient.get_secret', return_value=<return value>):
        settings = AzureKeyVaultSettings()
        assert settings.my_password == ...
        assert settings.sql_server__password == ...```

File: tests/test_sources.py [hunk 7]
```diff
@@ -97,3 +111,50 @@ def test___init___parent(self, mocker: MockerFixture, tmp_path: Path) -> None:
...
+            SQLSERVERPASSWORD: str
+            sql_server__password: str
+
+        key_vault_secret = KeyVaultSecret(SecretProperties(), 'SecretValue')
+        mocker.patch(f'{MODULE}.SecretClient.get_secret', return_value=key_vault_secret)
+        obj = AzureKeyVaultSettingsSource(
+            AzureKeyVaultSettings, 'https://my-resource.vault.azure.net/', DefaultAzureCredential()
+        )
+
+        obj()
```
hramezani, line 160: Here as well, you can test the mocked function return value

pydantic_settings/sources.py
hramezani: This is the basic functionality that we already have in all of the source classes.
Why is this complex? you need to call the

tests/test_sources.py
AndreuCodina: Done.

pydantic_settings/sources.py
AndreuCodina: I've added the alias support.

AndreuCodina:
@hramezani is everything ok?

hramezani:
@AndreuCodina thanks for the update.

I will review it later. My plan is to include it in the next minor release 2.4 if it is ready.

pydantic_settings/sources.py
```diff
@@ -1558,6 +1573,69 @@ def _read_file(self, file_path: Path) -> dict[str, Any]:
...
+        for field_name, field in self.settings_cls.model_fields.items():
+            field_value, field_key, value_is_complex = self.get_field_value(field, field_name)
+            field_value = self.prepare_field_value(field_name, field, field_value, value_is_complex)
+
+            if field_value is not None:
+                data[field_key] = field_value
+
+        return data
+
+    def _extract_field_info(self, field: FieldInfo, field_name: str) -> list[tuple[str, str, bool]]:
```
hramezani, line 1614: Why did you duplicate `_extract_field_info` here?

hramezani:
@AndreuCodina I checked the PR. we have some missed functionality like nested model support, case insensitive support. let me talk with the team and get back to you.

pydantic_settings/sources.py
AndreuCodina: It's not duplicated. I've had to modify it.

docs/index.md
```diff
@@ -1300,6 +1301,45 @@ class ExplicitFilePathSettings(BaseSettings):
...
 ```
 
+### Azure Key Vault
+
+You must set two parameters:
+
+- `url`: For example, `https://my-resource.vault.azure.net/`.
+- `credential`: If you use `DefaultAzureCredential`, in local you can execute `az login` to get your identity credentials. The identity must have a role assignement (the recommended one is `Key Vault Secrets User`), so you can have access to the secrets.
+
+```
```
samuelcolvin, line 1311:
``````suggestion
```py
``````

File: docs/index.md [hunk 10]
```diff
@@ -1300,6 +1301,45 @@ class ExplicitFilePathSettings(BaseSettings):
         )
 ```
 
+### Azure Key Vault
+
+You must set two parameters:
```
samuelcolvin, line 1306: we need to be clear about the limitations relative to other sources, if we can't get all KV pairs easily.

hramezani: @AndreuCodina if we can find a way to fetch all the key/value first and build a dict, we can easily use the logic that we have in `EnvSettingsSource`. otherwise, we have to be clear and document what exactly we support and what we not support relative to other sources.

AndreuCodina:
> @AndreuCodina I checked the PR. we have some missed functionality like nested model support, case insensitive support. let me talk with the team and get back to you.

What do I have to do then? I've checked EnvSettingsSource and if I have to copy (and maybe modify some line), this will not be maintainable. I've already duplicated complex code I don't understand.

I can get all secrets from Azure Key Vault (names and values) and replace `--` (these are the symbols used for nested/hierarchy data) with `__`, but if you require me to add a lot of code just to get secrets from Key Vault, I'd need to use shared code provided by the library.

After this source, I plan to add another source for Azure App Configuration, but I need more guidance here to re-use all code I can.

hramezani:
> I can get all available secrets from Azure Key Vault (names and values) and replace -- (these are the symbols used for nested/hierarchy data) with __, but if you require me to add a lot of code just to get secrets from Key Vault, I'd need to use shared code provided by the library.

No, we don't want to add the code from the library.

Right now, the `AzureKeyVaultSettingsSource` is doing a very simple thing. looping over model fields and fetching the values. but in other sources like `EnvSettingsSource` we can load nested model values. for example, by the following model:

```py
    class SubSubModel(BaseSettings):
        dvals: Dict

    class SubModel(BaseSettings):
        vals: List[str]
        sub_sub_model: SubSubModel

    class Cfg(BaseSettings):
        sub_model: SubModel

        model_config = SettingsConfigDict(env_prefix='cfg_', env_nested_delimiter='__')
``` 
You can set the envs like `cfg_sub_model__vals='["one", "two"]'` and `cfg_sub_model__sub_sub_model__dvals'='{"three": 4}'` and `pydantic-settings` load the values for the model.

I think we don't have the above functionality in `AzureKeyVaultSettingsSource`. I also believe we don't have `env_prefix` support there.

 The easy way to have all the functionality is to load all the possible key/value and inherit from `EnvSettingsSource`. like `DotEnvSettingsSource`. but it seems there is no easy way to load all the key/values.

So, I would suggest implementing the current functionality that we have in other sources as much as possible and documenting the functionalities that you don't support in `AzureKeyVaultSettingsSource`.

AndreuCodina:
> > I can get all available secrets from Azure Key Vault (names and values) and replace -- (these are the symbols used for nested/hierarchy data) with __, but if you require me to add a lot of code just to get secrets from Key Vault, I'd need to use shared code provided by the library.
> 
> No, we don't want to add the code from the library.
> 
> Right now, the `AzureKeyVaultSettingsSource` is doing a very simple thing. looping over model fields and fetching the values. but in other sources like `EnvSettingsSource` we can load nested model values. for example, by the following model:
> 
> ```python
>     class SubSubModel(BaseSettings):
>         dvals: Dict
> 
>     class SubModel(BaseSettings):
>         vals: List[str]
>         sub_sub_model: SubSubModel
> 
>     class Cfg(BaseSettings):
>         sub_model: SubModel
> 
>         model_config = SettingsConfigDict(env_prefix='cfg_', env_nested_delimiter='__')
> ```
> 
> You can set the envs like `cfg_sub_model__vals='["one", "two"]'` and `cfg_sub_model__sub_sub_model__dvals'='{"three": 4}'` and `pydantic-settings` load the values for the model.
> 
> I think we don't have the above functionality in `AzureKeyVaultSettingsSource`. I also believe we don't have `env_prefix` support there.
> 
> The easy way to have all the functionality is to load all the possible key/value and inherit from `EnvSettingsSource`. like `DotEnvSettingsSource`. but it seems there is no easy way to load all the key/values.
> 
> So, I would suggest implementing the current functionality that we have in other sources as much as possible and documenting the functionalities that you don't support in `AzureKeyVaultSettingsSource`.

Ok, I think it's all already done.

docs/index.md
AndreuCodina: Done.

docs/index.md
AndreuCodina: I can't do this because there's no Key Vault resource to test against.

pydantic_settings/sources.py
```diff
@@ -1679,6 +1692,48 @@ def _read_file(self, file_path: Path) -> dict[str, Any]:
...
+
+    def _load_env_vars(self) -> Mapping[str, str | None]:
+        self._secret_client = SecretClient(vault_url=self._url, credential=self._credential)  # type: ignore
+        secret_names: list[str] = [secret.name for secret in self._secret_client.list_properties_of_secrets()]
+        env_vars: dict[str, str | None] = {}
+
+        for secret_name in secret_names:
+            secret = self._secret_client.get_secret(secret_name)
+            secret_value = secret.value
+            env_vars[secret_name] = secret_value
```
hramezani, line 1729: It loops over all the secrets and loads their values. I haven't tested it yet. do you know how long does it take if we have e.g 100 secrets?

AndreuCodina: Less than 7 seconds.

It's the recommended option, and if you want a more advanced scenario or many more secrets, the normal is a) configure a refresh interval or b) use Azure App Configuration.

In other words, you instance your `BaseSettings` as a singleton when the backend is started.

pydantic_settings/sources.py
```diff
@@ -1725,6 +1738,47 @@ def _read_file(self, file_path: Path) -> dict[str, Any]:
...
+            settings_cls,
+            case_sensitive=True,
+            env_prefix='',
+            env_nested_delimiter='--',
+            env_ignore_empty=False,
+            env_parse_none_str=env_parse_none_str,
+            env_parse_enums=env_parse_enums,
+        )
+
+    def _load_env_vars(self) -> Mapping[str, str | None]:
```
samuelcolvin, line 1767: Couldn't we use a lazy mapping, something like

```py
from collections import Mapping


class AzureMapping(Mapping):
    def __init__(self, secret_client):
        self._secret_client = secret_client
        self._keys = self._secret_client.get_names()

    def __getitem__(self, __key):
        return self._secret_client.get_secret(__key)

    def __len__(self):
        return len(self._keys)

    def __iter__(self):
        return iter(self._keys)
```

hramezani: why it doesn't work. the whole point of lazy mapping is prevent looping over all keys in keyvalult and only fetch the one that we need.

AndreuCodina: @hramezani When I execute the code and instance `ApplicationSettings`, `_load_env_vars` is called once per key, so all of them are fetched.

AndreuCodina: It seems "lazy mapping" means only fetching the keys we need from Key Vault, not preventing the fetch action (lazy loading).

Ok, the new code works well, but in this example I get an error: `SettingsError: error getting value for field "sql_server" from source "AzureKeyVaultSettingsSource"`.  Specifically, the error happens because `__getitem__`
 is called with the key `SqlServer` instead of `SqlServer--Password.`

hramezani: I think it is working as expected. what is the of key `SqlServer` in key-vault? the below error: 

`SettingsError: error getting value for field "sql_server" from source "AzureKeyVaultSettingsSource"`

You defined `sql_server: SqlServer = Field(..., alias="SqlServer")` which is a nested field. so , you have to define a valid json string(because `pydantic-settings` considers this field as a complex field and tries to json parse it) for it in key-vault. the key has to be `SqlServer` and the value has to be e.g `'{"Password": "key"}'`

AndreuCodina: In Key Vault, I have the key `SqlServer--Password`. In this Azure service, the nesting is always done with the `--` separator. Key Vault secrets store secrets/passwords (their keys can only have alphanumeric characters and dashes), can be referenced in several ways (App Configuration references, App Configuration SDK, Key Vault references, Key Vault SDK, ASP.NET Core integration, pydantic-settings, etc.), and are shared among applications, so they mustn't contain a JSON attached to a Pydantic model, they must contain the raw secret. And we should be able to map groups of secrets into a Pydantic model (SqlServer--User, SqlServer--Password, SqlServer--...) without modifying the password.

Using the branch code without `AzureKeyVaultMapping`, this Azure service works in its proper way.

hramezani: you probably have value for `SqlServer` in your key-vault. then `pydantic-settings` fetched that one instead of `SqlServer--Password`. that's why I suggested changing the value of `SqlServer` to a valid json string.

So, I think the problem is because you have both `SqlServer` and `SqlServer--Password` defined in key-vault. then pydantic-settings tries to get the value first from `SqlServer` and fails.

This is not related to the key vault settings source, if you have both `SqlServer` and `SqlServer--Password` env variable, you will get the same error.

This is not a problem. you can fix it by:

- Remove the `SqlServer` key or convert it to a valid json string
- Or, change the alias for `sql_server`

AndreuCodina: Having only the secret `SqlServer--Password` in Key Vault, `__getitem__` is called twice: one with `SqlServer` and another one with `SqlServer--Password`.

AndreuCodina: Having only `SqlServer--Password` in Key Vault

image

it throws an exception because `__getitem__` is called twice: once with `SqlServer` and another one with `SqlServer--Password`. I've had to add a try...except.

I've pushed the changes if you want to try it. The tests are failing too.

hramezani: > it throws an exception because __getitem__ is called twice: once with SqlServer and another one with SqlServer--Password. I've had to add a try...except.

This is the correct behavior. I think you have to replace the `return None` in `except` block with a `raise KeyError` to make it similar to a real dict.

> I've pushed the changes if you want to try it. The tests are failing too.

The tests are failing because you mocked the `get_secret` to return `SecretValue` for all the keys. so, it returns `SecretValue` for `SqlServer` and this is not a valid json string. that's why you get a parsing error. you have to change the test in a way to only return `SecretValue` for secret keys that we want and `ResourceNotFoundError` for others

I will try it later

AndreuCodina: Done!

docs/index.md
```diff
@@ -1173,6 +1173,58 @@ Last, run your application inside a Docker container and supply your newly creat
...
+
+- `url`: For example, `https://my-resource.vault.azure.net/`.
+- `credential`: If you use `DefaultAzureCredential`, in local you can execute `az login` to get your identity credentials. The identity must have a role assignement (the recommended one is `Key Vault Secrets User`), so you can have access to the secrets.
+
+You must have the same naming convention in the field name as in the Key Vault secret name. For example, if the secret is named `SqlServerPassword`, the field name must be the same. You can use an alias too.
+
+In Key Vault, nested models are supported with the `--` separator. For example, `SqlServer--Password`.
+
+JSON objects, Key Vault arrays (e.g. `MySecret--0`, `MySecret--1`) and prefixes are not supported.
```
hramezani, line 1187: Are you sure that we can't use prefixes? I think the new implementation supports prefixes as well.

pydantic_settings/sources.py
```diff
@@ -1725,6 +1742,69 @@ def _read_file(self, file_path: Path) -> dict[str, Any]:
...
+        self._loaded_secrets = {}
+        self._secret_client = secret_client
+        self._secret_names: list[str] = [secret.name for secret in self._secret_client.list_properties_of_secrets()]
+
+    def __getitem__(self, key: str) -> str | None:
+        if key not in self._loaded_secrets:
+            try:
+                self._loaded_secrets[key] = self._secret_client.get_secret(key).value
+            except ResourceNotFoundError:  # type: ignore
+                raise KeyError
```
hramezani, line 1763:
```suggestion
                raise KeyError(key)
```

docs/index.md
AndreuCodina: Yes, it does. Updated.

docs/index.md
```diff
@@ -1173,6 +1173,58 @@ Last, run your application inside a Docker container and supply your newly creat
 docker service create --name pydantic-with-secrets --secret my_secret_data pydantic-app:latest
 ```
 
+## Azure Key Vault
+
+You must set two parameters:
+
+- `url`: For example, `https://my-resource.vault.azure.net/`.
+- `credential`: If you use `DefaultAzureCredential`, in local you can execute `az login` to get your identity credentials. The identity must have a role assignement (the recommended one is `Key Vault Secrets User`), so you can have access to the secrets.
```
hramezani, line 1181:
```suggestion
- `credential`: If you use `DefaultAzureCredential`, in local you can execute `az login` to get your identity credentials. The identity must have a role assignment (the recommended one is `Key Vault Secrets User`), so you can access the secrets.
```

docs/index.md
```diff
@@ -1173,6 +1173,58 @@ Last, run your application inside a Docker container and supply your newly creat
...
+        env_settings: PydanticBaseSettingsSource,
+        dotenv_settings: PydanticBaseSettingsSource,
+        file_secret_settings: PydanticBaseSettingsSource,
+    ) -> tuple[PydanticBaseSettingsSource, ...]:
+        return(
+            AzureKeyVaultSettingsSource(
+                settings_cls,
+                os.environ['AZURE_KEY_VAULT_URL'],
+                DefaultAzureCredential()
+        )
```
hramezani, line 1225:
```suggestion
import os
from azure.identity import DefaultAzureCredential
from pydantic import BaseModel
from pydantic_settings import (
    BaseSettings,
    PydanticBaseSettingsSource
)
from pydantic_settings.sources import AzureKeyVaultSettingsSource


class SubModel(BaseModel):
    a: str


class AzureKeyVaultSettings(BaseSettings):
    foo: str
    bar: int
    sub: SubModel

    @classmethod
    def settings_customise_sources(
        cls,
        settings_cls: type[BaseSettings],
        init_settings: PydanticBaseSettingsSource,
        env_settings: PydanticBaseSettingsSource,
        dotenv_settings: PydanticBaseSettingsSource,
        file_secret_settings: PydanticBaseSettingsSource,
    ) -> tuple[PydanticBaseSettingsSource, ...]:
        az_key_vault_settings = AzureKeyVaultSettingsSource(
            settings_cls,
            os.environ['AZURE_KEY_VAULT_URL'],
            DefaultAzureCredential(),
        )
        return init_settings, env_settings, dotenv_settings, file_secret_settings, az_key_vault_settings
```

hramezani, line 1225:
```suggestion
import os

from azure.identity import DefaultAzureCredential
from pydantic import BaseModel

from pydantic_settings import BaseSettings, PydanticBaseSettingsSource
from pydantic_settings.sources import AzureKeyVaultSettingsSource


class SubModel(BaseModel):
    a: str


class AzureKeyVaultSettings(BaseSettings):
    foo: str
    bar: int
    sub: SubModel

    @classmethod
    def settings_customise_sources(
        cls,
        settings_cls: type[BaseSettings],
        init_settings: PydanticBaseSettingsSource,
        env_settings: PydanticBaseSettingsSource,
        dotenv_settings: PydanticBaseSettingsSource,
        file_secret_settings: PydanticBaseSettingsSource,
    ) -> tuple[PydanticBaseSettingsSource, ...]:
        az_key_vault_settings = AzureKeyVaultSettingsSource(
            settings_cls,
            os.environ['AZURE_KEY_VAULT_URL'],
            DefaultAzureCredential(),
        )
        return (
            init_settings,
            env_settings,
            dotenv_settings,
            file_secret_settings,
            az_key_vault_settings,
        )
```

docs/index.md
```diff
@@ -1173,6 +1173,61 @@ Last, run your application inside a Docker container and supply your newly creat
...
+- `credential`: If you use `DefaultAzureCredential`, in local you can execute `az login` to get your identity credentials. The identity must have a role assignment (the recommended one is `Key Vault Secrets User`), so you can access the secrets.
+
+You must have the same naming convention in the field name as in the Key Vault secret name. For example, if the secret is named `SqlServerPassword`, the field name must be the same. You can use an alias too.
+
+In Key Vault, nested models are supported with the `--` separator. For example, `SqlServer--Password`.
+
+Key Vault arrays (e.g. `MySecret--0`, `MySecret--1`) are not supported.
+
+```python
+import os
```
hramezani, line 1190:
```suggestion
import os
from typing import Tuple, Type
```

docs/index.md
```diff
@@ -1173,6 +1173,61 @@ Last, run your application inside a Docker container and supply your newly creat
...
+            os.environ['AZURE_KEY_VAULT_URL'],
+            DefaultAzureCredential(),
+        )
+        return (
+            init_settings,
+            env_settings,
+            dotenv_settings,
+            file_secret_settings,
+            az_key_vault_settings,
+        )
```
hramezani, line 1228:
```suggestion
import os
from typing import Tuple, Type

from azure.identity import DefaultAzureCredential
from pydantic import BaseModel

from pydantic_settings import BaseSettings, PydanticBaseSettingsSource
from pydantic_settings.sources import AzureKeyVaultSettingsSource


class SubModel(BaseModel):
    a: str


class AzureKeyVaultSettings(BaseSettings):
    foo: str
    bar: int
    sub: SubModel

    @classmethod
    def settings_customise_sources(
        cls,
        settings_cls: Type[BaseSettings],
        init_settings: PydanticBaseSettingsSource,
        env_settings: PydanticBaseSettingsSource,
        dotenv_settings: PydanticBaseSettingsSource,
        file_secret_settings: PydanticBaseSettingsSource,
    ) -> Tuple[PydanticBaseSettingsSource, ...]:
        az_key_vault_settings = AzureKeyVaultSettingsSource(
            settings_cls,
            os.environ['AZURE_KEY_VAULT_URL'],
            DefaultAzureCredential(),
        )
        return (
            init_settings,
            env_settings,
            dotenv_settings,
            file_secret_settings,
            az_key_vault_settings,
        )
```

hramezani:
@AndreuCodina I changed the example in the doc and made it simpler.

Please add `AzureKeyVaultSettingsSource` to the [__init__](https://github.com/pydantic/pydantic-settings/blob/6ffd6bd8950b56a1d277b24e5872d42638296898/pydantic_settings/__init__.py#L2) and then change the doc example to import `AzureKeyVaultSettingsSource` from `pydantic_settings` instead of `pydantic_settings.sources`

AndreuCodina:
> @AndreuCodina I changed the example in the doc and made it simpler.
> 
> Please add `AzureKeyVaultSettingsSource` to the [**init**](https://github.com/pydantic/pydantic-settings/blob/6ffd6bd8950b56a1d277b24e5872d42638296898/pydantic_settings/__init__.py#L2) and then change the doc example to import `AzureKeyVaultSettingsSource` from `pydantic_settings` instead of `pydantic_settings.sources`

Done. I've removed "```python" in the markdown because if not, the tests fail.

docs/index.md
```diff
@@ -1173,6 +1173,62 @@ Last, run your application inside a Docker container and supply your newly creat
...
+
+```
+import os
+from typing import Tuple, Type
+
+from azure.identity import DefaultAzureCredential
+from pydantic import BaseModel
+
+from pydantic_settings import BaseSettings, PydanticBaseSettingsSource
+from pydantic_settings import AzureKeyVaultSettingsSource
```
hramezani, line 1197:
```suggestion
```py
import os

from azure.identity import DefaultAzureCredential

    AzureKeyVaultSettingsSource,
    BaseSettings,
    PydanticBaseSettingsSource,
)
```

hramezani:
> > @AndreuCodina I changed the example in the doc and made it simpler.
> > Please add `AzureKeyVaultSettingsSource` to the [**init**](https://github.com/pydantic/pydantic-settings/blob/6ffd6bd8950b56a1d277b24e5872d42638296898/pydantic_settings/__init__.py#L2) and then change the doc example to import `AzureKeyVaultSettingsSource` from `pydantic_settings` instead of `pydantic_settings.sources`
> 
> Done. I've removed "```python" in the markdown because if not, the tests fail.

it fails because of imports in the code. I've fixed it and returned the "```python"

hramezani (approved):

hramezani:
Thanks @AndreuCodina

PR #280, opened 2024-04-23
Use macos-13 for Python 3.8 and 3.9

hramezani:


.github/workflows/ci.yml
```diff
@@ -31,8 +31,20 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: [ubuntu, macos, windows]
+        os: [ubuntu, macos-13, macos, windows]
```
davidhewitt, line 34:
```suggestion
        os: [ubuntu-latest, macos-13, macos-latest, windows-latest]
```

and also change line 51.

But I wonder, why is `macos-13` necessary? `lastest` is `macos-12`, which is `x86_64`. `macos-14` is coming soon, which is M1.

hramezani: MacOs `latest` is `14.4.1` and python 3.8 and 3.9 are not available in `14.4.1`
That's why I changed the CI to use `macos-13` for them.

davidhewitt (approved):

Issue #308, opened 2024-06-07
Error when using validator in nested model with case_sensitive=True and using upper case


nstonic:
Example:

```py
class DjangoSettings(BaseModel):
    ALLOWED_HOSTS: list[str] = ["127.0.0.1", "localhost"]

    @field_validator("ALLOWED_HOSTS", mode="before")
    def parse_comma_separated(cls, value):
        if not isinstance(value, str):
            return value
        return [item.strip() for item in value.split(",") if item]


class EnvSettings(BaseSettings):
    DJ: DjangoSettings

    model_config = SettingsConfigDict(
        env_nested_delimiter="__",
        case_sensitive=True,
        extra="forbid",
    )

```

In envs: `DJ__ALLOWED_HOSTS: '127.0.0.1, localhost'`

Crashes with an error:

```
Traceback (most recent call last):
  File "/opt/app/src/manage.py", line 22, in <module>
    main()
  File "/opt/app/src/manage.py", line 18, in main
    execute_from_command_line(sys.argv)
  File "/usr/local/lib/python3.11/site-packages/django/core/management/__init__.py", line 442, in execute_from_command_line
    utility.execute()
  File "/usr/local/lib/python3.11/site-packages/django/core/management/__init__.py", line 382, in execute
    settings.INSTALLED_APPS
  File "/usr/local/lib/python3.11/site-packages/django/conf/__init__.py", line 102, in __getattr__
    self._setup(name)
  File "/usr/local/lib/python3.11/site-packages/django/conf/__init__.py", line 89, in _setup
    self._wrapped = Settings(settings_module)
                    ^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/site-packages/django/conf/__init__.py", line 217, in __init__
    mod = importlib.import_module(self.SETTINGS_MODULE)
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/importlib/__init__.py", line 126, in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "<frozen importlib._bootstrap>", line 1206, in _gcd_import
  File "<frozen importlib._bootstrap>", line 1178, in _find_and_load
[...]

PR #309, opened 2024-06-10
Fix issue with nested model uppercase field name in case insensitive mode

hramezani:
Fixes https://github.com/pydantic/pydantic-settings/issues/308

### DIFF
--- a/pydantic_settings/sources.py
+++ b/pydantic_settings/sources.py
@@ -597,8 +597,11 @@ class EnvSettingsSource(PydanticBaseEnvSettingsSource):
 
         return True, allow_parse_failure
 
+    # Default value of `case_sensitive` is `False`, because we don't want to break existing behavior.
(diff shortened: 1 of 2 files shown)


sydney-runkle (approved):
Nice works, thanks for the documentation!

pydantic_settings/sources.py
```diff
@@ -597,8 +597,11 @@ def _field_is_complex(self, field: FieldInfo) -> tuple[bool, bool]:
 
         return True, allow_parse_failure
 
+    # Default value of `case_sensitive` is `False`, because we don't want to break existing behavior.
```
hramezani, line 600:
```suggestion
    # Default value of `case_sensitive` is `None`, because we don't want to break existing behavior.
```
```suggestion
    # Default value of `case_sensitive` is `False`, because we don't want to break existing behavior.
```

Issue #311, opened 2024-06-12
Settings parsing breaks with complex type


ducminh-phan:
Minimal example

`main.py`:

```python
from typing import Annotated, Literal

from pydantic import BaseModel, Field, Json
from pydantic_settings import BaseSettings, SettingsConfigDict


class AWSS3(BaseModel):
    type: Literal["aws"]


class AzureBlobStorage(BaseModel):
    type: Literal["azure"]


Storage = Annotated[
    AWSS3 | AzureBlobStorage,
    Field(discriminator="type"),
]


class Settings(BaseSettings):
    storage: Json[Storage] | None = None

    model_config = SettingsConfigDict(
        env_file=".env",
    )


settings = Settings()
print(settings)
```

`.env`:

```env
STORAGE='{
    "type": "aws"
}'
```

With `pydantic@2.7.3`, `pydantic-settings@2.2.1`, the script `main.py` produces output

```
storage=AWSS3(type='aws')
```

With `pydantic@2.7.3`, `pydantic-settings@2.3.2`, the script `main.py` produces the following error

```
Traceback (most recent call last):
  File "/.../venv/lib/python3.10/site-packages/pydantic_settings/sources.py", line 368, in __call__
    field_value, field_key, value_is_complex = self.get_field_value(field, field_name)
  File "/.../venv/lib/python3.10/site-packages/pydantic_settings/sources.py", line 534, in get_field_value
    for field_key, env_name, value_is_complex in self._extract_field_info(field, field_name):
  File "/.../venv/lib/python3.10/site-packages/pydantic_settings/sources.py", line 273, in _extract_field_info
    elif origin_is_union(get_origin(field.annotation)) and _union_is_complex(field.annotation, field.metadata):
  File "/.../venv/lib/python3.10/site-packages/pydantic_settings/sources.py", line 1621, in _union_is_complex
    return any(_annotation_is_complex(arg, metadata) for arg in get_args(annotation))
  File "/.../venv/lib/python3.10/site-packages/pydantic_settings/sources.py", line 1621, in <genexpr>
    return any(_annotation_is_complex(arg, metadata) for arg in get_args(annotation))
  File "/.../venv/lib/python3.10/site-packages/pydantic_settings/sources.py", line 1600, in _annotation_is_complex
    inner, meta = get_args(annotation)
ValueError: too many values to unpack (expected 2)

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
[...]

PR #312, opened 2024-06-12
Fix an intriduced bug in parsing json field with discriminated union

hramezani:
Fixes https://github.com/pydantic/pydantic-settings/issues/311


sydney-runkle (approved):
Nice work!

PR #318, opened 2024-06-20
Deprecate `read_env_file` and move it to `DotEnvSettingsSource`

### LINKED ISSUE #296: Can't read Pydantic Settings from stdin
### Requirement

I need to read Settings from stdin. This is not our primary scenario but something we need for tooling, for security reasons (secrets we don't want on disk). Our tooling reads settings to process them, and ideally they should be piped to our app to leave no traces behind, even temporarily.

### What we tried

Settings pass filenames only into the public API. And using the stdin filename '/dev/stdin' doesn't work (on Mac) as the reads don't block until data is available (it simply returns an empty file). 

Because we couldn't find a clean way to do this, I think it is an issue that should be addressed. Perhaps there is a different way to do it that we missed, please let me know.

### Possible Fixes 

The best fix would be to simply support streams everywhere in the API, instead of just filenames.

It would also be fine to support special filenames '-' or '/dev/stdin' as special cases that are internally replaced by `stream=sys.stdin`. Or to allow a SettingsSource to provide a stream instead of file names. 

In our case we hacked a third solution deriving a class from `DotEnvSettingsSource` because we need most of that functionality (just need to replace the source in some cases).  But because [`read_env_file`](https://github.com/pydantic/pydantic-settings/blob/6d25cee4bb7a6db592ca0da123c53f3d775cd1e1/pydantic_settings/sources.py#L901) is a global function instead of a method, we cannot simply override that and had to copy a few other methods from the implementation. Moving that into the DotEnvSettingsSource class would allow a clean override of just `read_env_file` that could handle stdin reading. This is probably the minimal solution that cleanly resolves this requirement, without changing anything for the mainstream usage of Pydantic settings.

WarpedPixel:
Fixes #296
Adds a method _read_env_file to DotEnvSettingsSource that can be easily overridden.
This fix keeps a global function so it can continue to be tested, but adds an underscore to the name.
The plain read_env_file is marked as deprecated.

### DIFF
--- a/pydantic_settings/sources.py
+++ b/pydantic_settings/sources.py
@@ -753,11 +753,23 @@ class DotEnvSettingsSource(EnvSettingsSource):
     def _load_env_vars(self) -> Mapping[str, str | None]:
         return self._read_env_files()
 
+    @staticmethod
+    def _static_read_env_file(
+        file_path: Path,
+        *,
+        encoding: str | None = None,
+        case_sensitive: bool = False,
+        ignore_empty: bool = False,
+        parse_none_str: str | None = None,
+    ) -> Mapping[str, str | None]:
+        file_vars: dict[str, str | None] = dotenv_values(file_path, encoding=encoding or 'utf8')
+        return parse_env_vars(file_vars, case_sensitive, ignore_empty, parse_none_str)
+
     def _read_env_file(
         self,
         file_path: Path,
     ) -> Mapping[str, str | None]:
-        return _read_env_file(
+        return self._static_read_env_file(
             file_path,
             encoding=self.env_file_encoding,
             case_sensitive=self.case_sensitive,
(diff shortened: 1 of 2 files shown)


WarpedPixel:
please review

pydantic_settings/sources.py
```diff
@@ -753,6 +753,30 @@ def __init__(
...
+        parse_none_str: str | None = None,
+    ) -> Mapping[str, str | None]:
+        file_vars: dict[str, str | None] = dotenv_values(file_path, encoding=encoding or 'utf8')
+        return parse_env_vars(file_vars, case_sensitive, ignore_empty, parse_none_str)
+
+    def _read_env_file(
+        self,
+        file_path: Path,
+    ) -> Mapping[str, str | None]:
+        return self._static_read_env_file(
```
hramezani, line 772: can't we use the global `read_env_file` here? then we don't need to deprecate `read_env_file`.
I mean we can only add `_read_env_file` to `DotEnvSettingsSource` and use it in `_read_env_files`.
Then you can override `DotEnvSettingsSource._read_env_file` for your usecase.

Am I missing something?

WarpedPixel: Yes, we could. My assumption is that we did not want to continue to export unqualified global names because that creates a legacy we have a hard time changing. Which was your original concern with this change, the potential to break people. This gives us a path to delete the deprecated function and in the next version simply no longer have that legacy.
If we would rather keep that legacy and not rock the boat, it is easy enough to keep the global function.

pydantic_settings/sources.py
```diff
@@ -1708,8 +1724,17 @@ def read_env_file(
     ignore_empty: bool = False,
     parse_none_str: str | None = None,
 ) -> Mapping[str, str | None]:
-    file_vars: dict[str, str | None] = dotenv_values(file_path, encoding=encoding or 'utf8')
-    return parse_env_vars(file_vars, case_sensitive, ignore_empty, parse_none_str)
+    warnings.warn(
+        'read_env_file will be removed in the next version, use DotEnvSettingsSource._static_read_env_file if you must',
```
hramezani, line 1728: Please add a test for this warning

WarpedPixel: added test_read_env_file_deprecation

pydantic_settings/sources.py
WarpedPixel: Assuming we are ok with the current solution, if not happy to make to change

hramezani:
Thanks @WarpedPixel

PR #324, opened 2024-06-24
Add Azure App Configuration settings source

AndreuCodina:
Azure App Configuration enable having configurations and secrets and change them in several microservices without re-deploying them.

Azure Key Vault stores secrets. Azure App Configuration provides configurations, secrets, several secret stores, external references, auto-reloading, feature flags, and a lot of complex usages.

This settings source copies the minimal configuration of the ASP.NET Core settings source.


AndreuCodina:
@hramezani Code from `main` is not passing the lint action. Can you check it, please?

AndreuCodina:
@hramezani By the way, this source uses the builder pattern, like in ASP.NET Core. I prefer it, but by implementing all options without it, we can end up with certain complexity:

- Over 30 parameters in the source constructor (different types of connections, permissions, timeouts, key vault, snapshots, refreshing time, feature flags, keys, trimmings, external sources). Each configuration has its own set of available settings.
- Importing not needed types, e.g. `keys=[AzureAppConfigurationSeletor("my_key__*", label="Development")]` instead of `.select_key("my_key__*", label="Development")`

Is it ok for you?

The Key Vault source was simpler, but we have a lot of microservices using App Configuration. Can you release it at the same time as the another one?

hramezani:
> Code from main is not passing the lint action. Can you check it, please?

It is because you've updated the `ruff` version in your PR. so you need to address the required fixes to make ruff happy.

I need to talk with the team about the new source. right now I am not sure that we want to support it.

Also, the implementation is like the old implementation of key vault(looping over keys to fetching values). probably we need a lazy mapping or something like that.
BTW, please wait for my feedback from team

hramezani:
@AndreuCodina 
I talked with the team and we decided not to accept the Azure App Configuration settings source. here are some reasons for this decision:

- The implementation looks complex and it is really hard to maintain such a complex code.
- We haven't received a lot of requests to support this source.
- It is possible for people like you to implement it in their own codebase.

Thanks for your effort here and sorry for this.

AndreuCodina:
If anyone needs it, I have published a Python package for this: https://github.com/AndreuCodina/pydantic-settings-azure-app-configuration

hramezani:
Thanks @AndreuCodina for put it on a package

PR #326, opened 2024-06-27
feat: Enable access to the current state in settings sources

### LINKED ISSUE #223: How to build a cascading settings resolver

VictorColomb:
## Why this PR?

This PR aims at enabling custom settings sources to access the output of the previous sources, through the `.previous_state` property.

The ultimate use-case is the same as #224, however this PR does not introduce a breaking change.

Fixes #223.

## Use-case

My real-life use-case: retrieving settings from AWS Secrets Manager but setting the secret name in environment variables.

## Discussion

_There are a number of things to be discussed about this PR, and I would very much appreciate the input / guidance of the pydantic-settings maintainers._

1. I have reversed the order in which the settings sources are discovered ([here](https://github.com/VictorColomb/pydantic-settings/blob/main/pydantic_settings/main.py#L312-L315)) in order for the output of the sources with the most priority to be available to the ones of lesser importance. This seems like the most intuitive behavior to me, but this is rather subjective. Any preference?
2. Currently, settings sources are being handed the raw output from the previous ones. In other words, nothing has yet been validated by Pydantic. Would it be desired behavior to instead pass a validated instance of `self.settings_cls`? I have not gone this way because imho it adds to much complexity for not that much benefit.
3. I have defined the `__previous_state` setter method as "private", in that Python will perform name mangling on the method. This way, we are almost certain to introduce no breaking changes at all (someone could have defined their own `set_previous_state` method). However, that means calling `_PydanticBaseSettingsSource__set_previous_state` in the sources discovery code, which is neither elegant nor pythonic. My question therefore is should I remove the double-underscore for cleaner code at the marginal risk of introducing a breaking change for some?
4. Should I add a bit of documentation, for example as a subsection of the [Adding sources](https://docs.pydantic.dev/latest/concepts/pydantic_settings/#adding-sources) section?

### DIFF
--- a/pydantic_settings/sources.py
+++ b/pydantic_settings/sources.py
@@ -126,36 +126,36 @@ class PydanticBaseSettingsSource(ABC):
     def __init__(self, settings_cls: type[BaseSettings]):
         self.settings_cls = settings_cls
         self.config = settings_cls.model_config
-        self._previous_state: dict[str, Any] = {}
-        self._previous_states: dict[str, dict[str, Any]] = {}
+        self._current_state: dict[str, Any] = {}
+        self._settings_sources_data: dict[str, dict[str, Any]] = {}
 
-    def _set_previous_state(self, state: dict[str, Any]) -> None:
+    def _set_current_state(self, state: dict[str, Any]) -> None:
         """
         Record the state of settings from the previous settings sources. This should
         be called right before __call__.
         """
-        self._previous_state = state
+        self._current_state = state
 
-    def _set_previous_states(self, states: dict[str, dict[str, Any]]) -> None:
+    def _set_settings_sources_data(self, states: dict[str, dict[str, Any]]) -> None:
         """
         Record the state of settings from all previous settings sources. This should
         be called right before __call__.
         """
-        self._previous_states = states
+        self._settings_sources_data = states
 
     @property
-    def previous_states(self) -> dict[str, dict[str, Any]]:
-        """
-        The state of all previous settings sources.
-        """
-        return self._previous_states
-
-    @property
-    def previous_state(self) -> dict[str, Any]:
+    def current_state(self) -> dict[str, Any]:
         """
         The current state of the settings, populated by the previous settings sources.
         """
-        return self._previous_state
+        return self._current_state
+
+    @property
+    def settings_sources_data(self) -> dict[str, dict[str, Any]]:
+        """
+        The state of all previous settings sources.
+        """
+        return self._settings_sources_data
 
     @abstractmethod
     def get_field_value(self, field: FieldInfo, field_name: str) -> tuple[Any, str, bool]:
--- a/pydantic_settings/sources.py
+++ b/pydantic_settings/sources.py
@@ -126,21 +126,21 @@ class PydanticBaseSettingsSource(ABC):
     def __init__(self, settings_cls: type[BaseSettings]):
         self.settings_cls = settings_cls
         self.config = settings_cls.model_config
-        self.__previous_state: dict[str, Any] = {}
+        self._previous_state: dict[str, Any] = {}
 
-    def __set_previous_state(self, state: dict[str, Any]) -> None:
+    def _set_previous_state(self, state: dict[str, Any]) -> None:
         """
         Record the state of settings from the previous settings sources. This should
         be called right before __call__.
         """
-        self.__previous_state = state
+        self._previous_state = state
 
     @property
     def previous_state(self) -> dict[str, Any]:
         """
         The current state of the settings, populated by the previous settings sources.
         """
-        return self.__previous_state
+        return self._previous_state
 
     @abstractmethod
     def get_field_value(self, field: FieldInfo, field_name: str) -> tuple[Any, str, bool]:
--- a/pydantic_settings/sources.py
+++ b/pydantic_settings/sources.py
@@ -12,7 +12,6 @@ from argparse import SUPPRESS, ArgumentParser, HelpFormatter, Namespace, _SubPar
 from collections import deque
 from dataclasses import is_dataclass
 from enum import Enum
-from functools import reduce
 from pathlib import Path
 from types import FunctionType
 from typing import (
@@ -127,7 +126,15 @@ class PydanticBaseSettingsSource(ABC):
     def __init__(self, settings_cls: type[BaseSettings]):
         self.settings_cls = settings_cls
         self.config = settings_cls.model_config
+        self._previous_state: dict[str, Any] = {}
         self._previous_states: dict[str, dict[str, Any]] = {}
+
+    def _set_previous_state(self, state: dict[str, Any]) -> None:
+        """
+        Record the state of settings from the previous settings sources. This should
+        be called right before __call__.
+        """
+        self._previous_state = state
 
     def _set_previous_states(self, states: dict[str, dict[str, Any]]) -> None:
         """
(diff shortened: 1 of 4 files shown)


hramezani:
Thanks @VictorColomb for this PR and sorry for the late response.

> I have reversed the order in which the settings sources are discovered ([here](https://github.com/VictorColomb/pydantic-settings/blob/main/pydantic_settings/main.py#L312-L315)) in order for the output of the sources with the most priority to be available to the ones of lesser importance. This seems like the most intuitive behavior to me, but this is rather subjective. Any preference?

This is good

> Currently, settings sources are being handed the raw output from the previous ones. In other words, nothing has yet been validated by Pydantic. Would it be desired behavior to instead pass a validated instance of self.settings_cls? I have not gone this way because imho it adds to much complexity for not that much benefit.

Your approach is fine. don't need to validate beforehand. 

> I have defined the __previous_state setter method as "private", in that Python will perform name mangling on the method. This way, we are almost certain to introduce no breaking changes at all (someone could have defined their own set_previous_state method). However, that means calling _PydanticBaseSettingsSource__set_previous_state in the sources discovery code, which is neither elegant nor pythonic. My question therefore is should I remove the double-underscore for cleaner code at the marginal risk of introducing a breaking change for some?

Yes, let's have it without double underscore. let's have the `previous_states` as a property and `_previous_states` and the setter as a public method.


> Should I add a bit of documentation, for example as a subsection of the [Adding sources](https://docs.pydantic.dev/latest/concepts/pydantic_settings/#adding-sources) section?

sure.


I have a request:

- The current implementation passes only the previous source state to the next one, I would suggest to pass all the previous states to the next one. so, the states should be something like

```
{'source_key': {source_state}}
```

Please let me know when you are ready for the next round of review.

VictorColomb:
Hi @hramezani. Thanks for your feedback!

I have implemented your suggestions.

However, I do see a potential pitfall in recording all previous states: if two settings sources have the same name (e.g. two instances of the same class, but with different init parameters), only the latest one would be recorded in `previous_states`.

pydantic_settings/sources.py
```diff
@@ -126,6 +127,28 @@ class PydanticBaseSettingsSource(ABC):
...
+
+    @property
+    def previous_states(self) -> dict[str, dict[str, Any]]:
+        """
+        The state of all previous settings sources.
+        """
+        return self._previous_states
+
+    @property
+    def previous_state(self) -> dict[str, Any]:
```
hramezani, line 147: Do we really need this property here?

hramezani:
Thanks @VictorColomb for updating the PR.

- Please add some tests related to `previous_states`

pydantic_settings/sources.py
VictorColomb: In my opinion, this property covers the majority of use-cases and avoids end-users the trouble of having to scan through all entries of the `previous_states` dictionary to find the key they are looking for.

As an example, I have opened this PR because I would like to read a "secret name" from a previous settings source and use it to retrieve a JSON document from AWS Secrets Manager. It does not matter to me whether the secret name comes from env vars, a dotenv file or the cli arguments. Therefore, the `previous_state` property is the one I would like to use.

VictorColomb:
@hramezani test added! I also updated my branch from main.

pydantic_settings/sources.py
hramezani: Ok, that makes sense.

hramezani: Ok, that makes sense. As we have the `state` in `main.py` loop, isn't it better to set the state there? I mean we already have `deep_update` there.

VictorColomb: I'm not sure what you mean. Something like [this diff](https://topaz.github.io/paste/#XQAAAQB+BwAAAAAAAAAyGkkJwvwRfVDeCk9H2J4aD4LFWYdfouH8Zl+cjlbJpqoN1QDlrZOpnDV0gB3D+Kwc3qYU+JNV10stFOHO+K0AXsQWyEXAZyffe+a9Gw2xJBVc15TqbudD7c8WIHu6zatZsXk26uI4+aNT3bEyfs4KSOTjXXWY+HdcMrUhWULnr8i2DM+yhJkTCtJwkBoPkvUaKmzi+PMEbP6j+hl/Hp3iEl227e8dgaMJ0uycxTc5J/N2LRGH2aV2dBTBJAnho0zvJ2JMt5f1OQtgI3g5O+yG1nYWtox6rC9Jkmf5g4t/wXOxqRwemC42PBuKRnP1YMnB26eRkrUzK1JXl1jkooPyE0C7TlPrPBzPgdM57zWGXK//pa4fc9FHmRaVesjnOEZRSi9k7MXGuyVCTZ7dn5guQkJijTkuSV8U51ivqVASps/f2nrXNMCynPcrR99VoOr8+b0y0zI7U5Nd66YvON10NYEFBHN8xMqYovPS5GxToHXIc223kgVhNtKXaSOdmCbLo5MYvTbUAL6DtI9X5RM5D/RNEbGvjcx1tVVRf6H8lPCC+rDFB/0OGqT+QbyLVuMGm1uQgE/EKkliZjvr+1ZikDAJdDZjYjo9YQgIZF203WKbzuU4rWtTZXPGmlh9Brx+s6xrQcY3Wy2cmyu/Qg85nF0VpudqDZ6eg6uRrEer181PbNyLo3YsABrxklznUNyGuSStCU+c3B1nSP3z6W6/KdYIaSZG9kRkIHdz7JuQHuPUXnH7u+XbLixOl8rA9vvlwhWAMMdKrLmYEHnb74cyYfoJyzQaO6BQDQ31KRCyYQ9iTpWlJg3p26BNn4ronjAQvgO5iDx/emq4g/A8lTKr3vv3sK+J5DCE+u1Ewt7t3Pgz/0thXWR4OkHvZ76HYP/ofxek)?

hramezani: Exactly!

VictorColomb: Done!

pydantic_settings/sources.py
```diff
@@ -126,6 +126,36 @@ class PydanticBaseSettingsSource(ABC):
...
+
+    def _set_previous_states(self, states: dict[str, dict[str, Any]]) -> None:
+        """
+        Record the state of settings from all previous settings sources. This should
+        be called right before __call__.
+        """
+        self._previous_states = states
+
+    @property
+    def previous_states(self) -> dict[str, dict[str, Any]]:
```
hramezani, line 147: I think `previous_states` and `previous_state` may make user confused. what about replacing  `previous_states` with `settings_sources_data` and `previous_state` with `current_state`?

If you have any suggestions please let me know.

pydantic_settings/sources.py
hramezani: Thanks!

pydantic_settings/sources.py
VictorColomb: I have `current_settings_state` in mind, but it's uselessly verbose.

I don't find `settings_sources_data` particularly explicit, but I cannot find a better name. Maybe `previous_settings_sources_output`, or `previous_sources_output` (a little shorter)?

In any case, I've implemented your suggestions. Let me know what you think!

hramezani (approved):

hramezani:
Thanks @VictorColomb

PR #328, opened 2024-06-30
Fix regex flags accidentally passed as count

musicinmybrain:
The `flags` argument to `re.sub()` was accidentally passed positionally in the place of the `count` argument in `tests/test_settings.py`. Fix this by passing it as a keyword argument, which additionally fixes a `DeprecationWarning` in Python 3.13+ due to passing `flags` and/or `count` as positional arguments. See: https://docs.python.org/3.13/library/re.html#re.sub


hramezani:
Thanks @musicinmybrain

PR #338, opened 2024-07-10
Add EnvYamlConfigSettingsSource with env substitution support

KochankovID:
In some cases, an additional level of mapping is required between the names of the environment variables and the attributes in yaml file.

For example, in my company, they want to distinct responsibility between the devops team that names environment variables and the developers team that uses these variables

I added support for this using a third-party envyaml library and wrote tests. I also added information about this to the documentation

Review pls 🙏


hramezani:
Thanks @KochankovID for this PR.

We are not going to support this new source because you can easily create your custom settings source(like the one that you created in this PR) and use it in your code.

KochankovID:
@hramezani I got your point, but the same thing we can tell about yaml source, or toml. Why are they supported?

hramezani:
> @hramezani I got your point, but the same thing we can tell about yaml source, or toml. Why are they supported?

These are the most common sources and people request them a lot. but your user-case is a valid one but not very common. if we want to accept these less common sources, we will end up with many sources that are hard to maintain.

Thanks for your effort and sorry that we can't accept it.

KochankovID:
@hramezani Thank you for the response ❤️ Accepted the answer

hramezani:
> @hramezani Thank you for the response ❤️ Accepted the answer

Thanks for your understanding!

PR #341, opened 2024-07-11
Add get_subcommand function.

### LINKED ISSUE #335: No arguments to CLI subparser fails to raise an error
With `pydantic-settings` 2.3.4,  when no arguments are provided to a subparser style CLI, there is no exception raised. 

I believe a user should not have to explicitly catch for these `None` cases on the "Root" `Settings` model.

## Acceptance criteria:

- An error is raised when no arguments are provided to a subparser
- Add explicit test case for no arguments provided
- Updated docs to include an example and suggested patterns on how to run a program (or function) using subcommands.

To demonstrate this issue:

```python
import sys
from pydantic_settings import (BaseSettings, CliSubCommand)


class Alpha(BaseModel):
    """Alpha Utils"""
    name: str


class Beta(BaseModel):
    """Beta Utils"""
    age: int


class Root(BaseSettings, cli_parse_args=True, cli_prog_name="example"):
    alpha: CliSubCommand[Alpha]
    beta: CliSubCommand[Beta]



def runner_01(r: Root) -> int:
    print(f"Running {r}")
    return 0


if __name__ == '__main__':
    sys.exit(runner_01(Root()))

```

Running would be:

```bash
> python example.py                    
Running alpha=None beta=None
> echo $?
0
```

It's possible to workaround the current design by mapping these `Optional` fields of `Root` to a tuple, then case matching (alternatively, writing with `if/elif`). 

```python
def runner_02(r: Root) -> int:
    match (r.alpha, r.beta):
        case (None, None): raise ValueError(f"Invalid {r}")
        case (None, beta): print(f"Running beta {beta=}")
        case (alpha, None): print(f"Running {alpha=}")
        case (_, _): raise ValueError(f"Invalid {r}")
    return 0
```

However, it's a bit clunky and not particularly obvious that you need to handle these `None` cases. I believe the structural validation/checking should be occurring at the pydantic-settings level.

## Misc Feedback

From a type standpoint, the current design is a bit peculiar because it leans on `None`. 

```python
class Root(BaseSettings, cli_parse_args=True, cli_prog_name="example"):
    alpha: CliSubCommand[Alpha] # sugar for Optional[_CliSubCommand]
    beta: CliSubCommand[Beta]
```

I believe the it would be useful to use a union type in the design. 

```python
class Root(BaseSettings, cli_parse_args=True, cli_prog_name="example"):
    command: Alpha | Beta
```

Calling a function/main would be:

```python
def runner(m: Model):

    match m.command:
        case Alpha(): print(f"Running alpha={m.command}")
        case Beta(): print(f"Running beta={m.command}")
```

In general, it seems like there's a core part of the API that is missing the plumbing layer of calling your function with the initialized and validated settings. 

There's also bit of friction with using `mypy`.

```python


class Root(BaseSettings, cli_parse_args=True):
    name: str


if __name__ == "__main__":
    print(Root())
```

Yields

```bash
> mypy example.py         
example.py:9: error: Missing named argument "name" for "Root"  [call-arg]
Found 1 error in 1 file (checked 1 source file)
```

Perhaps this is general issue with `pydantic-settings` and should be filed as a separate ticket.

kschwab:
Fixes #335


kschwab:
Added support for unioned CLI subcommands and positional args. Users can now mix and match subcommand styles as they see fit, with results populated into the relevant fields and other set to `None`:

```python
import sys
from pydantic_settings import BaseSettings, CliSubCommand, get_subcommand

class alpha(BaseModel):
    """Alpha Utils"""
    name: str

    def sub_main(self) -> int:
        print(f"Running alpha {self}")
        return 0

class beta(BaseModel):
    """Beta Utils"""
    age: int

    def sub_main(self) -> int:
        print(f"Running beta {self}")
        return 0

class gamma(BaseModel):
    """Gamma Utils"""
    height: int

    def sub_main(self) -> int:
        print(f"Running gamme {self}")
        return 0

class RootA(BaseSettings, cli_parse_args=True, cli_use_class_docs_for_groups=True):
    subcommand: CliSubCommand[alpha | beta | gamma]

    def main(self) -> int:
        print(f"Running root A {self}")
        return self.subcommand.sub_main() if self.subcommand is not None else 1

class RootB(BaseSettings, cli_parse_args=True, cli_use_class_docs_for_groups=True):
    subcommands: CliSubCommand[alpha | beta]
    gamma: CliSubCommand[gamma]

    def main(self) -> int:
        print(f"Running root B {self}")
        return get_subcommand(self).sub_main()

class RootC(BaseSettings, cli_parse_args=True, cli_use_class_docs_for_groups=True):
    alpha: CliSubCommand[alpha]
    beta: CliSubCommand[beta]
    gamma: CliSubCommand[gamma]

    def main(self) -> int:
        print(f"Running root C {self}")
        return get_subcommand(self).sub_main()

if __name__ == '__main__':
    sys.exit(RootA().main())
```

Still need to add relevant tests and doc updates.

kschwab:
@hramezani this PR is ready. Updated docs are [here](https://github.com/pydantic/pydantic-settings/blob/4a63f8a904b87f637a0829cb0a365ca20055581e/docs/index.md#subcommands-and-positional-arguments).

hramezani:
@kschwab Thanks. I will take a look later this week.

hramezani:
@kschwab Thanks for creating this PR.

I checked it a little bit and didn't like the idea of adding `def main` or `def sub_main`. isn't it possible to somehow remove these methods? can't the CLI settings source handle it(I mean calling the sub_command method)?

I got confused of the way that we should use the `sub_command` method. it is probably because:

- I didn't read the issue carefully
- It is not documented well

Also, I am afraid that this introduces a breaking change and this is the thing that we hardly try to avoid

kschwab:
@hramezani, I'll get back to these soon. No worries though on this PR with respect to breaking changes etc. It's purely a utility function. I'll update the docs to make it clearer. All it does is grab a subcommand if set, and raise an error if required:

```python
class Git(BaseSettings, cli_parse_args=True, cli_prog_name='git'):
    """git - The stupid content tracker"""

    clone: CliSubCommand[Clone] = Field(description='Clone a repo ...')
    plugins: CliSubCommand[Plugins] = Field(description='Fake GIT plugins')


sys.argv = ['example.py', 'clone', 'repo', 'dest']
git = Git()
print(git.model_dump())
#> {'clone': {'repository': 'repo', 'directory': 'dest', 'local': False}, 'plugins': None}

print(get_subcommand(git).model_dump())
#> {'repository': 'repo', 'directory': 'dest', 'local': False}

sys.argv = ['example.py']
git = Git()
print(git.model_dump())
#> {'clone': None, 'plugins': None}

print(get_subcommand(git).model_dump())
#> Error: CLI subcomand is required {clone, plugins}

print(get_subcommand(git, is_required=False).model_dump())
#> None
```

The docs currently just show how it is most likely to be applied.

hramezani:
@kschwab It is still unclear why we need the `get_subcommand` utility. why can't we use the `.model_dump`?

kschwab:
> @kschwab It is still unclear why we need the get_subcommand utility. why can't we use the .model_dump?

It's for convenience. Users can use `model_dump` etc., but it's atypical from a CLI application point of view for users to handle this manually.

e.g., if we look at `docker` subcommands:

```text
$ docker help

Usage:  docker [OPTIONS] COMMAND

A self-sufficient runtime for containers

Common Commands:
  run         Create and run a new container from an image
  exec        Execute a command in a running container
  ps          List containers
  build       Build an image from a Dockerfile
  pull        Download an image from a registry
  push        Upload an image to a registry
  images      List images
  login       Log in to a registry
  logout      Log out from a registry
  search      Search Docker Hub for images
  version     Show the Docker version information
  info        Display system-wide information

Management Commands:
  builder     Manage builds
  container   Manage containers
  context     Manage contexts
  image       Manage images
  manifest    Manage Docker image manifests and manifest lists
  network     Manage networks
  plugin      Manage plugins
  system      Manage Docker
  trust       Manage trust on Docker images
  volume      Manage volumes

Swarm Commands:
  swarm       Manage Swarm

Commands:
  attach      Attach local standard input, output, and error streams to a running container
  commit      Create a new image from a container's changes
  cp          Copy files/folders between a container and the local filesystem
  create      Create a new container
  diff        Inspect changes to files or directories on a container's filesystem
  events      Get real time events from the server
  export      Export a container's filesystem as a tar archive
  history     Show the history of an image
  import      Import the contents from a tarball to create a filesystem image
  inspect     Return low-level information on Docker objects
  kill        Kill one or more running containers
  load        Load an image from a tar archive or STDIN
  logs        Fetch the logs of a container
  pause       Pause all processes within one or more containers
  port        List port mappings or a specific mapping for the container
  rename      Rename a container
  restart     Restart one or more containers
  rm          Remove one or more containers
  rmi         Remove one or more images
  save        Save one or more images to a tar archive (streamed to STDOUT by default)
  start       Start one or more stopped containers
  stats       Display a live stream of container(s) resource usage statistics
  stop        Stop one or more running containers
  tag         Create a tag TARGET_IMAGE that refers to SOURCE_IMAGE
  top         Display the running processes of a container
  unpause     Unpause all processes within one or more containers
  update      Update configuration of one or more containers
  wait        Block until one or more containers stop, then print their exit codes
```

Without `get_subcommand`, it means each user will have to implement their own function (using `model_dump` or some other means) to find the subcommand that was invoked. This is what was raised in #335 as an issue. The `get_subcommand` utility simply finds and returns the active subcommand so the user doesn't have to.

As an aside, outside of this PR, there are only two other features (union/alias support for `CliSubCommand` and `CliApp`) I believe are necessary for bringing the `CliSettingsSource` up to a typical CLI implementation. So not much more, but these last few I do think are important as they are relatively "standard" features for CLI applications.

hramezani (approved):

hramezani:
Thanks @kschwab for the explanation and for working on this.

Please resolve the conflict and then it will be ready to go

kschwab:
Thanks @hramezani, good to go.

Issue #342, opened 2024-07-11
RootModel[str] cannot be parsed/requires double quotation.


user1584:
I am trying to create an extension of standard str class. My idea was to just create a RootModel based on str and then add some custom logic. However, I encountered strange problems when using the resulting class as part of the settings.
My RootModel looks like this (without custom logic):
```
import pydantic

class CustomStr(pydantic.RootModel[str]):
    root: str
```

Initializing with `CustomStr("hello world")` works as expected. The next step was to use it as part of the settings:
```
import pydantic_settings
import os

os.environ["custom_str"] = 'hello world'

class Settings(pydantic_settings.BaseSettings):
    custom_str: CustomStr

Settings()
```

This results in 'SettingsError: error parsing value for field "custom_str" from source "EnvSettingsSource"' (see complete traceback below).
If I switch the type of custom_str from CustomStr to standard str, the example works. During lots of debugging, I noticed that it works if I wrap the string in the environment variable in double quotes like `os.environ["custom_str"] = '"hello world"'` since this can be parsed by json.loads. Note: The order of single and double quotes matters!
So, the question is: Is there any way to let pydantic/pydantic_settings handle the custom model like regular str class?

I am using pydantic version '2.7.1' and pydantic_settings version '2.2.1'.

Here's the complete traceback:
```
---------------------------------------------------------------------------
JSONDecodeError                           Traceback (most recent call last)
File /home/work/.venv/lib/python3.9/site-packages/pydantic_settings/sources.py:323, in PydanticBaseEnvSettingsSource.__call__(self)
    322 try:
--> 323     field_value = self.prepare_field_value(field_name, field, field_value, value_is_complex)
    324 except ValueError as e:

File /home/work/.venv/lib/python3.9/site-packages/pydantic_settings/sources.py:513, in EnvSettingsSource.prepare_field_value(self, field_name, field, value, value_is_complex)
    512     if not allow_parse_failure:
--> 513         raise e
    515 if isinstance(value, dict):

File /home/work/.venv/lib/python3.9/site-packages/pydantic_settings/sources.py:510, in EnvSettingsSource.prepare_field_value(self, field_name, field, value, value_is_complex)
    509 try:
--> 510     value = self.decode_complex_value(field_name, field, value)
    511 except ValueError as e:

File /home/work/.venv/lib/python3.9/site-packages/pydantic_settings/sources.py:147, in PydanticBaseSettingsSource.decode_complex_value(self, field_name, field, value)
[...]

PR #344, opened 2024-07-13
Determine RootModel complexity from root type

user1584:
As described in #342, on initialization of RootModels based on non-complex types such as str, the json.loads function is called to parse the environment variable contents. Thus, even though str-based variables can be passed as plain strings, a str-based RootModel requires the string to be wrapped in double quotes to make it json-compliant. The proposed code change works around this behaviour by determining the complexity of the field base on the root attribute's type. Thus, RootModels with non-complex root types should be handled as if the root type was used without being wrapped in a RootModel.


hramezani:
Thanks @user1584 for this PR.

The Python 3.8 and 3.9 failing is because `annotation.__annotations__.get('root', None)` returns `RootModelRootType` for `RootModel[List[Foo]]` in python 3.8 and 3.9 but it returns `None` for the rest of python version.

So, we can probably ignore this like:

```py
    if isinstance(annotation, type) and issubclass(annotation, RootModel):
        root_annotation = annotation.__annotations__.get('root', None)
        if root_annotation is not None and root_annotation != 'RootModelRootType':
            annotation = root_annotation
```

Also, I would suggest moving the whole logic to `def _annotation_is_complex`:

user1584:
@hramezani, thank you for your suggestion! I adapted the PR accordingly and the tests are working now.

hramezani (approved):

hramezani:
Thanks @user1584

PR #348, opened 2024-07-22
Add `nested_model_default_partial_update` flag and `DefaultSettingsSource`

### LINKED ISSUE #203: How to Override Deeply Nested Settings using Environment Variables?
Hello,

Is it possible to override a deeply nested setting without having to redefine the entirety of the model?

Below is a modified example based off of [Parsing environment variable values](https://docs.pydantic.dev/latest/concepts/pydantic_settings/#parsing-environment-variable-values):

```python
import os


class DeepSubModel(BaseModel):
    v4: str


class SubModel(BaseModel):
    v1: str
    v2: bytes
    v3: int
    deep: DeepSubModel


class Settings(BaseSettings):
    model_config = SettingsConfigDict(env_nested_delimiter='__')

    v0: str
    sub_model: SubModel

    @classmethod
        cls,
        settings_cls,
        init_settings,
        env_settings,
        dotenv_settings,
        file_secret_settings):
        return env_settings, init_settings, file_secret_settings


# Ideal scenario would be a simple point modification
os.environ['SUB_MODEL__DEEP__V4'] = 'override-v4'

try:
    print(Settings(v0='0', sub_model=SubModel(v1='init-v1', v2=b'init-v2', v3=3, 
                   deep=DeepSubModel(v4='init-v4'))).model_dump())
except ValidationError as e:
    print(e)
    """
    pydantic_core._pydantic_core.ValidationError: 3 validation errors for Settings
    sub_model.v1
      Field required [type=missing, input_value={'deep': {'v4': 'override-v4'}}, input_type=dict]
        For further information visit https://errors.pydantic.dev/2.5/v/missing
    sub_model.v2
      Field required [type=missing, input_value={'deep': {'v4': 'override-v4'}}, input_type=dict]
        For further information visit https://errors.pydantic.dev/2.5/v/missing
    sub_model.v3
      Field required [type=missing, input_value={'deep': {'v4': 'override-v4'}}, input_type=dict]
        For further information visit https://errors.pydantic.dev/2.5/v/missing
    """

# Current scenario seems to require entire definition of nested modes etc.
os.environ['SUB_MODEL'] = '{"v1": "reinit-v1", "v2": "reinit-v2"}'
os.environ['SUB_MODEL__V3'] = '33'

print(Settings(v0='0', sub_model=SubModel(v1='init-v1', v2=b'init-v2', v3=3, 
               deep=DeepSubModel(v4='init-v4'))).model_dump())
"""
{'v0': '0', 'sub_model': {'v1': 'reinit-v1', 'v2': b'reinit-v2', 'v3': 33, 'deep': {'v4': 'override-v4'}}}
"""
```

The difference here is `Settings` is defined through instantiation instead of environment variables. Ideally, the below concept would still apply to `Settings` with respect to nested precedence, allowing for point modifications of nested variables:
> Nested environment variables take precedence over the top-level environment variable JSON...

### LINKED ISSUE #345: Nested Settings with instances of models as defaults
When I have the next Models:

``` 


class DatabaseSettings(BaseModel):
    name: str = "test"
    env: str = "test"


class Settings(BaseSettings):
    db1: DatabaseSettings = DatabaseSettings(name="database1")
    db2: DatabaseSettings = DatabaseSettings(name="database2")


print(Settings(db1={"env": "prd"}))
``` 

I will get:

```
db1=DatabaseSettings(name='test', env='prd') db2=DatabaseSettings(name='database2', env='test')
``` 

However, I would expect:

```
db1=DatabaseSettings(name='database1', env='prd') db2=DatabaseSettings(name='database2', env='test')
``` 

I have written a small PydanticBaseSettingsSource fixes this behavior:

```
class FromClsSettings(pydantic_settings.PydanticBaseSettingsSource):
    def get_field_value(self, **kwargs) -> None:  # type: ignore[override]  # noqa: ANN003
        pass

    def __call__(self) -> dict[str, Any]:
        data = {}
        for field, field_info in self.settings_cls.model_fields.items():
            if isinstance(field_info.default, pydantic.BaseModel):
                data[field] = field_info.default.model_dump()
        return data
```

Is the issue the expected behavior? Is the proposed solution the correct one? Shall I make a PR?

### LINKED ISSUE #347: Environment Parsing with Default Nested Class
@hramezani, is the below case expected or a bug? I would have thought them to be equivalent. I came across this when updating the CLI handling of defaults. ~~I'll work to resolve it if it's a bug but wanted to confirm first before proceeding forward.~~

I added a potential fix if it is an issue.

```python

import os
from pydantic import BaseModel, ValidationError

class NestedA(BaseModel):
    v0: bool
    v1: bool

class NestedB(BaseModel):
    v0: bool = False
    v1: bool = True

class SettingsGood(BaseSettings, env_nested_delimiter='__'):
    nested: NestedB = NestedB()

class SettingsBad(BaseSettings, env_nested_delimiter='__'):
    nested: NestedA = NestedA(v0=False, v1=True)

os.environ['NESTED__V0'] = 'True'
print(SettingsGood().model_dump())
"""
{'nested': {'v0': True, 'v1': True}}
"""

try:
    print(SettingsBad().model_dump())
except ValidationError as e:
    print(e)
"""
1 validation error for SettingsBad
nested.v1
  Field required [type=missing, input_value={'v0': 'True'}, input_type=dict]
"""
```

### LINKED ISSUE #363: partial specify the sub model value does not work
```
class SubModel(BaseModel):
    v1: str = "default"
    v2: bytes = b"hello"
    v3: int = 3

class Settings(BaseSettings):
        cli_prog_name="prog1",
        cli_parse_args=True
    )

    v0: str = "ok"
    sub_model: SubModel = SubModel(v1="top default", v3=33)

sys.argv = [
    'example.py',
    "--sub_model.v1='cli'"
]
print(Settings())
# > v0='from env' sub_model=SubModel(v1="'cli'", v2=b'hello', v3=3)
```

expected value:
```
# > v0='from env' sub_model=SubModel(v1="'cli'", v2=b'hello', v3=33)
```

similar cause of the problem: if v3 does not have default in SubModel, it raise an error:
```
sub_model.v3
  Field required [type=missing, input_value={'v1': "'cli'"}, input_type=dict]
```

I think we need to enable partial update with nested sub model, otherwise, the BaseSettings does not really support nested model field as cli arg or environment variable. If we decide not to support partial update, I guess it is better to drop the env_nested and --field.subfield idea completely. Because this behaviour, especially the silent fallback to SubModel's default will completely surprise the user.

kschwab:
Fixes #347
Fixes #203
Fixes #345 
Fixes #363

### DIFF
--- a/pydantic_settings/sources.py
+++ b/pydantic_settings/sources.py
@@ -10,7 +10,7 @@ import warnings
 from abc import ABC, abstractmethod
 from argparse import SUPPRESS, ArgumentParser, HelpFormatter, Namespace, _SubParsersAction
 from collections import deque
-from dataclasses import asdict, is_dataclass
+from dataclasses import is_dataclass
 from enum import Enum
 from pathlib import Path
 from types import FunctionType
@@ -265,12 +265,7 @@ class DefaultSettingsSource(PydanticBaseSettingsSource):
                 if field_info.validate_default is not False:
                     resolved_name = self._get_resolved_name(field_name, field_info)
                     if field_info.default not in (PydanticUndefined, None):
-                        if is_model_class(field_info.annotation):
-                            defaults[resolved_name] = field_info.default.model_dump()
-                        elif is_dataclass(field_info.annotation):
-                            defaults[resolved_name] = asdict(field_info.default)
-                        else:
-                            defaults[resolved_name] = field_info.default
+                        defaults[resolved_name] = field_info.default
                     elif field_info.default_factory is not None:
                         defaults[resolved_name] = field_info.default_factory
         return defaults
@@ -311,7 +306,7 @@ class DefaultSettingsSource(PydanticBaseSettingsSource):
         return self.defaults
 
     def __repr__(self) -> str:
-        return f'DefaultSettingsSource()'
+        return 'DefaultSettingsSource()'
 
 
 class InitSettingsSource(PydanticBaseSettingsSource):
--- a/pydantic_settings/sources.py
+++ b/pydantic_settings/sources.py
@@ -254,15 +254,15 @@ class DefaultSettingsSource(PydanticBaseSettingsSource):
     Source class for loading default object values.
     """
 
-    def __init__(self, settings_cls: type[BaseSettings], default_objects_copy_by_value: bool | None = None):
+    def __init__(self, settings_cls: type[BaseSettings], nested_model_partial_update: bool | None = None):
         super().__init__(settings_cls)
         self.defaults: dict[str, Any] = {}
-        self.default_objects_copy_by_value = (
-            default_objects_copy_by_value
-            if default_objects_copy_by_value is not None
-            else self.config.get('default_objects_copy_by_value', False)
+        self.nested_model_partial_update = (
+            nested_model_partial_update
+            if nested_model_partial_update is not None
+            else self.config.get('nested_model_partial_update', False)
         )
-        if self.default_objects_copy_by_value:
+        if self.nested_model_partial_update:
             for field_name, field_info in settings_cls.model_fields.items():
                 if is_dataclass(type(field_info.default)):
                     self.defaults[_field_name_for_signature(field_name, field_info)] = asdict(field_info.default)
@@ -277,7 +277,7 @@ class DefaultSettingsSource(PydanticBaseSettingsSource):
         return self.defaults
 
     def __repr__(self) -> str:
-        return f'DefaultSettingsSource(default_objects_copy_by_value={self.default_objects_copy_by_value})'
+        return f'DefaultSettingsSource(nested_model_partial_update={self.nested_model_partial_update})'
 
 
 class InitSettingsSource(PydanticBaseSettingsSource):
@@ -289,14 +289,14 @@ class InitSettingsSource(PydanticBaseSettingsSource):
         self,
         settings_cls: type[BaseSettings],
         init_kwargs: dict[str, Any],
-        default_objects_copy_by_value: bool | None = None,
+        nested_model_partial_update: bool | None = None,
     ):
         self.init_kwargs = init_kwargs
         super().__init__(settings_cls)
-        self.default_objects_copy_by_value = (
-            default_objects_copy_by_value
-            if default_objects_copy_by_value is not None
-            else self.config.get('default_objects_copy_by_value', False)
+        self.nested_model_partial_update = (
+            nested_model_partial_update
+            if nested_model_partial_update is not None
+            else self.config.get('nested_model_partial_update', False)
         )
 
     def get_field_value(self, field: FieldInfo, field_name: str) -> tuple[Any, str, bool]:
@@ -306,7 +306,7 @@ class InitSettingsSource(PydanticBaseSettingsSource):
     def __call__(self) -> dict[str, Any]:
         return (
             TypeAdapter(Dict[str, Any]).dump_python(self.init_kwargs)
-            if self.default_objects_copy_by_value
+            if self.nested_model_partial_update
             else self.init_kwargs
         )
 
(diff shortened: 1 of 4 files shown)


kschwab:
Still need to update docs.

kschwab:
@hramezani please review the [updated docs](https://github.com/pydantic/pydantic-settings/blob/f812cb6e058b31a8702cd0fd512d07bc4873fc79/docs/index.md#parsing-default-objects) to make sure they meet your expectations. PR is ready to go otherwise 👍🏾

pydantic_settings/sources.py
```diff
@@ -247,24 +249,74 @@ def __call__(self) -> dict[str, Any]:
...
+        self.defaults: dict[str, Any] = {}
+        self.default_objects_copy_by_value = (
+            default_objects_copy_by_value
+            if default_objects_copy_by_value is not None
+            else self.config.get('default_objects_copy_by_value', False)
+        )
+        for field_name, field_info in settings_cls.model_fields.items():
+            if not self.default_objects_copy_by_value:
+                if is_dataclass(type(field_info.default)) or is_model_class(type(field_info.default)):
+                    self.defaults[_field_name_for_signature(field_name, field_info)] = field_info.default
```
kschwab, line 268: @hramezani also consider whether this case should be kept or removed (lines 267-268)

Currently, without these lines, the below behavior is observed:

```python
class SubModel(BaseModel):
    val: int = 0
    flag: bool = False

ORIGINAL_OBJECT = SubModel()

class SettingsCopyByReference(BaseSettings):
    model_config = SettingsConfigDict(env_nested_delimiter='__')

    default_object: SubModel = ORIGINAL_OBJECT


by_ref = SettingsCopyByReference(default_object=ORIGINAL_OBJECT)
assert by_ref.default_object is ORIGINAL_OBJECT # Passes for default object in kwargs

try:
    by_ref = SettingsCopyByReference()
    assert by_ref.default_object is ORIGINAL_OBJECT, 'Fails for default object in class'
except AssertionError as err:
    print(err)
    #> Fails for default object in class
```

My assumption is the behavior should be the same. Therefore, if we keep these lines it **_fixes_** the above such that behavior is the same for both class and kwarg default objects.

If not, we can remove these lines and keep current behavior.

pydantic_settings/sources.py
```diff
@@ -247,24 +249,74 @@ def __call__(self) -> dict[str, Any]:
...
+            TypeAdapter(Dict[str, Any]).dump_python(self.init_kwargs)
+            if self.default_objects_copy_by_value
+            else self.init_kwargs
+        )
 
     def __repr__(self) -> str:
-        return f'InitSettingsSource(init_kwargs={self.init_kwargs!r})'
+        return (
+            f'InitSettingsSource(init_kwargs={self.init_kwargs!r}, '
+            f'default_objects_copy_by_value={self.default_objects_copy_by_value})'
```
hramezani, line 318: Let's remove this line here. it will introduce a breaking change

pydantic_settings/sources.py
kschwab: Agreed. Also, `BaseModel` does not return the same instance for class default objects.

Should we change the flag name to something else more appropriate? I'm not sure what to call it, do you have any thoughts?

hramezani: The flag name should refer to partial updates and nested objects. like `nested_model_partial_update`.

I am open to your suggestions as well.

kschwab:
@hramezani I have no more changes here. The only thing we might consider is changing it from `nested_model_partial_update` to `nested_default_model_partial_update` if you think users may get confused.

hramezani:
> @hramezani I have no more changes here. The only thing we might consider is changing it from `nested_model_partial_update` to `nested_default_model_partial_update` if you think users may get confused.

@kschwab `nested_model_default_partial_update` seems better to me

kschwab:
Thanks @hramezani. Good to go.

hramezani (approved):

hramezani:
@kschwab Thanks for the PR. it is really a great addition.

It would be great to mention this fix in the issues mentioned in PR description and ask them for a test. Could you post a message on issues?

kschwab:
@hramezani I've asked for confirmation in issues #345 and #363. The other two are already covered in the current tests as they were filed by me 😄

Issue #346, opened 2024-07-19
More ``__init__`` Overwrite Support


acederberg:
Hello everybody!

My apologies if this is a bit long.

## The problem

Support for passing additional overwriting keyword arguments for additional sources to ``__init__`` is confusing. I would like to overwrite some values in one of my sources, but it looks like I'd have to create my source in the ``settings_customize_sources`` classmethod to get it into ``_settings_build_values``, which means I'd have to overwrite the ``_settings_build_values`` method somehow. This is less than desirable, since I have no means to inject those sources using ``super``.

## My Use Case

I have a bit of an odd use case for [a library I'm writing](https://github.com/acederberg/pydantic-settings-yaml). This library supports loading ``YAML`` from many sources optionally with reloading of the files or not when creating new instances. I know that there is a ``YAML`` source defined here, but my use case is different enough that I want to write my own source.

Essentially, I want to inject sources on an instance level when overwrites (``_yaml_files`` and ``_yaml_settings``) are provided, otherwise it uses the source created with the subclass hook:

```python
class BaseYamlSettings(BaseSettings):
    __yaml_settings_cls__: ClassVar[CreateYamlSettings]
    __yaml_settings_self__: CreateYamlSettings | None
    __yaml_exclude__: bool

    if TYPE_CHECKING:
        model_config: ClassVar[loader.YamlSettingsConfigDict]

    def __init_subclass__(cls, **kwargs):
        """Create the default ``CreateYamlSettings`` instance."""
        super().__init_subclass__(**kwargs)
        cls.__yaml_settings_cls__ = CreateYamlSettings(cls)

    def __init__(
        self,
        _yaml_files: loader.YamlSettingsFilesInput | None = None,
        _yaml_reload: bool | None = None,
        _yaml_exclude: bool = False,
        **kwargs,
    ):
        # NOTE: If any overwrites are added, then do not use ``__yaml_settings_cls__``.
        #       This is a pain because the other option is to overwrite
        #       ``_settings_build_values``, which likely results in having
        #       to overwrite ``settings_customise_sources``.
        if _yaml_files is not None:
            yaml_settings = self.model_config.copy()
            yaml_settings.update(
                yaml_files=_yaml_files,
                yaml_reload=_yaml_reload,
            )
            self.__yaml_settings_self__ = CreateYamlSettings(self.__class__)
        else:
            self.__yaml_settings_self__ = self.__yaml_settings_cls__

[...]

PR #350, opened 2024-07-23
Feature/instance level source customization

acederberg:
I have made this pull request to close [my own issue](https://github.com/pydantic/pydantic-settings/issues/346), #346.

I essentially wanted instance level source customization, therefore I added the callback to ``_settings_customize_instance``. 

This change should also help users when testing since they will not have to add ``settings_customize_sources`` to their class, and will only have to pass ``_settings_customize_instance`` to the constructor.

This also includes a small modification to ``pyproject.toml`` to define test dependencies to improve the developer experience.

Sorry that ``black`` went wild on ``settings.py``, there are only ``3`` lines actually added in there. I am assuming that actions will probably fix this, if not, I will fix it.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -46,7 +46,7 @@ dependencies = [
 dynamic = ['version']
 
 [project.optional-dependencies]
-test = ["pytest", "pytest-mock", "pytest-examples"]
+test = ["pytest>=8.3.0", "pytest-mock>=3.14.0", "pytest-examples>=0.0.12"]
 yaml = ["pyyaml>=6.0.1"]
 toml = ["tomli>=2.0.1"]
 azure-key-vault = ["azure-keyvault-secrets>=4.8.0", "azure-identity>=1.16.0"]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -46,6 +46,7 @@ dependencies = [
 dynamic = ['version']
 
 [project.optional-dependencies]
+test = ["pytest", "pytest-mock", "pytest-examples"]
 yaml = ["pyyaml>=6.0.1"]
 toml = ["tomli>=2.0.1"]
 azure-key-vault = ["azure-keyvault-secrets>=4.8.0", "azure-identity>=1.16.0"]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -46,7 +46,6 @@ dependencies = [
 dynamic = ['version']
 
 [project.optional-dependencies]
-test = ["pytest>=8.3.0", "pytest-mock>=3.14.0", "pytest-examples>=0.0.12"]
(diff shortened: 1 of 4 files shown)


hramezani:
Thanks @acederberg for this PR. please:

- Revert unrelated code formatting changes to make lint happy and also make the PR easier to review
- Add some tests for the change
- Document how to use the new feature in https://docs.pydantic.dev/latest/concepts/pydantic_settings/#customise-settings-sources

pyproject.toml
```diff
@@ -46,6 +46,7 @@ dependencies = [
 dynamic = ['version']
 
 [project.optional-dependencies]
+test = ["pytest>=8.3.0", "pytest-mock>=3.14.0", "pytest-examples>=0.0.12"]
```
hramezani, line 49: I think we don't need this change

acederberg: Could I add these in another pull request? I think it will make it easier to get things up and running for other contributors.

hramezani: probably need to improve the readme and include a contribution doc.
we have some useful commands in [Makefile](https://github.com/pydantic/pydantic-settings/blob/main/Makefile) that makes development easy like:

- `make install`
- `make format`
- `make lint`
- ...

acederberg: Oh okay! Funny enough I looked for one earlier. I'll revert this change.

acederberg:
@hramezani I added tests and docs, it looks like everything is good to go. I notice that the tests fail for ``3.8`` because of the use of ``tuple`` type hints - yet such type hints are included in the source. I will fix this so that tests pass but am concerned about this inconsistency.

acederberg:
@hramezani This should work. Is there anything else I ought to do?

hramezani:
Thanks @acederberg for the update.

I need to check it more with the team. not sure this is the best way to support it.

hramezani:
@acederberg Unfortunately, we are going not to accept this PR because:

- Supporting instance-level source customization needs some structural work that I prefer to do it in V3
- This is a specific use case and adding a new param does not make sense here. I know that `pydantic-settings` basically doesn't support it, but it doesn't make sense to add a new param now and probably remove it in V3

Thanks for your effort here and sorry for this

acederberg:
@hramezani No worries, I understand - these are good reasons.

Is there any way that I could help with v3?

hramezani:
> @hramezani No worries, I understand - these are good reasons.
> 
> Is there any way that I could help with v3?

Thanks for your understanding ❤️ 

Right now, we don't have any specific plan for V3. but we can keep the issue open and V3 tag to it to consider it when we are going to work on V3.

I am going to close the PR.

Thanks for your effort

Issue #356, opened 2024-07-31
Nested dataclass settings class gives `AttributeError:  __pydantic_fields__`


bschoenmaeckers:
I have a dataclass that I don't control as a nested object (wrapped by a BaseSettings class). This used to work great but after v2.3.0 (#214) it throws a AttributeError. See the following testcase that passes on `<=v2.2.1` but fails on `>=v2.3.0`.
```python
def test_nested_dataclass_setting(env):
    @dataclasses.dataclass
    class DataClass:
        value: str

    class SubSettings(BaseSettings, DataClass):
        model_config = SettingsConfigDict(alias_generator=str.lower)

    class Cfg(BaseSettings):
        model_config = SettingsConfigDict(env_nested_delimiter='__')

        sub: SubSettings

    env.set('SUB__VALUE', 'something')
    cfg = Cfg()
```

<details>
  <summary>The exception</summary>
  
  ```python
pydantic_settings\main.py:141: in __init__
    **__pydantic_self__._settings_build_values(
pydantic_settings\main.py:260: in _settings_build_values
    CliSettingsSource(
pydantic_settings\sources.py:902: in __init__
    self._connect_root_parser(
pydantic_settings\sources.py:1236: in _connect_root_parser
    self._add_parser_args(
pydantic_settings\sources.py:1328: in _add_parser_args
    self._add_parser_args(
pydantic_settings\sources.py:1255: in _add_parser_args
    for field_name, resolved_name, field_info in self._sort_arg_fields(model):
pydantic_settings\sources.py:1151: in _sort_arg_fields
    fields = model.__pydantic_fields__ if is_pydantic_dataclass(model) else model.model_fields
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

[...]

PR #357, opened 2024-08-01
Fix a bug in nested vanila dataclass

hramezani:
Fixes https://github.com/pydantic/pydantic-settings/issues/356


kschwab:
@hramezani it looks good. You'll want to add same logic to line [1282](https://github.com/pydantic/pydantic-settings/blob/f68e51a4d18e19b4be91438e7c3d922f1fbc5a66/pydantic_settings/sources.py#L1282) as well.

hramezani:
> @hramezani it looks good. You'll want to add same logic to line [1282](https://github.com/pydantic/pydantic-settings/blob/f68e51a4d18e19b4be91438e7c3d922f1fbc5a66/pydantic_settings/sources.py#L1282) as well.

Thanks @kschwab

PR #365, opened 2024-08-16
Add CLI bool flags

### LINKED ISSUE #361: Feature request: better support for argumentless CLI boolean flags
Currently, you can declare a model like this:

```python
class Settings(BaseSettings):
    dry_run: bool = False
        cli_parse_args=True,
        alias_generator=AliasGenerator(
            validation_alias=lambda s: AliasChoices(s, to_snake(s).replace("_", "-"))
        ),
    )
```

and end up with an application you can run in dry-run mode by running `myapp --dry-run=true`. But this is not exactly what is usually expected of CLI applications; usually, boolean flags are argumentless (`--dry-run`, `--force`, etc.) and their presence implies the associated value is true. Sometimes they have an inverse like `--no-dry-run` which implies the opposite.

I don't think pydantic-settings currently supports arguments like this. It would be very useful if it did :)

kschwab:
Fixes #361

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -871,17 +871,16 @@ options:
 #### CLI Boolean Flags
 
 Change whether boolean fields should be explicit or implicit by default using the `cli_implicit_flags` setting. By
-default, boolean fields are "explicit", meaning a boolean value must be explicitly provided to the flag, e.g.
+default, boolean fields are "explicit", meaning a boolean value must be explicitly provided on the CLI, e.g.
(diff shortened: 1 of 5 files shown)


kschwab:
@hramezani this is ready for review 👍🏾

hramezani:
Thanks @kschwab for implementing this feature.

@sersorrel could you please check it?

hramezani (approved):

PR #375, opened 2024-08-31
Cli prefix validation alias fix

kschwab:
Fixes CLI issue related to #25.

`cli_prefix` will add prefix to all CLI parser args. These parsed args (with prefixes) then get directly passed down to the env var layer. These were getting missed for non-complex cases as discussed (and expected) in #25. 

However, CLI must make this conversion since the prefix is reserved with the CLI parser. This fix simply maps the CLI prefixed arg to the expected non-prefixed env var.

Test case shows expected case.


hramezani (approved):

hramezani:
Thanks @kschwab

PR #377, opened 2024-09-02
Feat/support complex list definition on env vars [Closes #376]

### LINKED ISSUE #376: feat: Add Support for complex list defined from env vars
Currently , only list of complex types can only be defined completely from env vars as Json values.  The idea is to also allow
to define using nested delimiters, where index is explicitly defined in env vars:

```python
class SubType(BaseSettings):
    v1 :str 

class Config(Settings):
   top: list[SubType]
```

then, something like this should work...
```sh
CONFIG__0__TOP="foo"
```

inean:
Proposal of implementation.  It works but no sanity check on indexes are done 😅

### DIFF
--- a/pydantic_settings/sources.py
+++ b/pydantic_settings/sources.py
@@ -43,7 +43,7 @@ from pydantic import AliasChoices, AliasPath, BaseModel, Json, RootModel, TypeAd
 from pydantic._internal._repr import Representation
 from pydantic._internal._signature import _field_name_for_signature
 from pydantic._internal._typing_extra import WithArgsTypes, origin_is_union, typing_base
-from pydantic._internal._utils import deep_update, is_model_class, lenient_issubclass
+from pydantic._internal._utils import KeyType, deep_update, is_model_class, lenient_issubclass
 from pydantic.dataclasses import is_pydantic_dataclass
 from pydantic.fields import FieldInfo
 from pydantic_core import PydanticUndefined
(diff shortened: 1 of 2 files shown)


pydantic_settings/sources.py
```diff
@@ -804,6 +809,15 @@ class Cfg(BaseSettings):
             return None
 
         annotation = field.annotation if isinstance(field, FieldInfo) else field
+        if get_origin(annotation) is list:
+            try:
+                # check if key is an integer. If so, it's an index. we fake a field info with the list type
+                # so future calls can continue to traverse the model and set proper types for leaf nodes
+                int(key)
```
hramezani, line 816: we need a test case with not int index

pydantic_settings/sources.py
```diff
@@ -874,7 +891,13 @@ def explode_env_vars(self, field_name: str, field: FieldInfo, env_vars: Mapping[
             if isinstance(env_var, dict):
                 if last_key not in env_var or not isinstance(env_val, EnvNoneType) or env_var[last_key] == {}:
                     env_var[last_key] = env_val
-
+        if is_list:
+            # if field is list based, ensure keys are sequential and return an ordered list
+            result, values = [], result
+            for i in (str(i) for i in range(len(values))):
+                if i not in values:
+                    raise ValueError(f'Expected entry with index {i} for {field_name}')
```
hramezani, line 899: a test case for this line as well

hramezani:
Thanks @inean for this PR. please:

- Fix lint error
- Add requested test cases
- Add a new example in the docs to show the new feature

hramezani:
@inean is it going to support simple cases like:
```py
class Config(BaseSettings):
   top: List[str]

config = Config()

```
by `TOP__0=foo`?

inean:
No, only complex cases ie: List[List|Dict|Submodel]. I already refactoring to move code to deep_copy so it can handle those cases.

inean:
@hramezani Just one question:

given this code...

```python

class TopModel(BaseModel):
    v1: str | None = None
    v2: int 
 
class Settings(BaseSettings):
    top:  List[List[TopModel]]

...

env.set(TOP='[[{'v2': 'xx'}]]')
env.set(TOP__0__0__v2 = '1')
env.set(TOP__1__0__v2 = '2')
env.set(TOP__3__0__v2='3')

cfg = Settings()
```
Should `TOP__3__0__v2` be ignored or raise a `ValueError`? We can append new entries at the end of the list (`TOP__1` in this case), but  `TOP__3` requires a default value as `TOP__2`which we can't provide. Python is this case (when using `[].index` on a list with `index > len()` ) simply appends at the end, but I think that it's better to just ignore that insert or raise error

hramezani:
@inean Thanks for the update and sorry for being late.

After discussing with the team about supporting this syntax, we decided not to support it because:

- The code looks complex and pydantic-settings is already complex. It is not worth adding this complexity to support a feature partially.
- Also, the feature doesn't work for the [simple case that I mentioned above](https://github.com/pydantic/pydantic-settings/pull/377#issuecomment-2327159068) and I am sure people want it in the future if we introduce the new syntax

PR #379, opened 2024-09-03
CLI ignore external parser list fix

kschwab:
Fixes an issue with external parsers that use non-string type list args. They were getting parsed when they should be ignored. Internally, our parser always parses everything into type string, so if pre-parsed args are not of type string they should be ignored.

Simple example of issue:

```python
from argparse import ArgumentParser
from pydantic_settings import BaseSettings, CliSettingsSource

class Cfg(BaseSettings):
    var: str = 'hi'

parser = ArgumentParser()

# External parser using type=int is issue resolved for this fix
parser.add_argument('--num-list', action='append', type=int)

cli_settings = CliSettingsSource(Cfg, root_parser=parser)

parsed_args = parser.parse_args(['--num-list', '1', '--num-list', '2', '--num-list', '3'])
print(parsed_args.num_list)
#> [1, 2, 3]

assert Cfg(_cli_settings_source=cli_settings(parsed_args=parsed_args)).model_dump() == {'var': 'hi'}
```

Also added support for `SimpleNamespace`, which is related to but distinct from `argparse.Namespace`.

### DIFF
--- a/pydantic_settings/sources.py
+++ b/pydantic_settings/sources.py
@@ -18,6 +18,7 @@ from dataclasses import asdict, is_dataclass
 from enum import Enum
 from pathlib import Path
 from textwrap import dedent
+from types import SimpleNamespace
 from typing import (
     TYPE_CHECKING,
     Any,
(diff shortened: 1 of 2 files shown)


hramezani (approved):

hramezani:
Thanks @kschwab

Issue #336, opened 2024-07-09
Remove SystemExit usage 


mpkocher:
`SystemExit` inherits from `BaseException` and should be used very sparingly. 

https://docs.python.org/3/library/exceptions.html#SystemExit

Currently, the docs and examples are using it in several places. 

https://docs.pydantic.dev/latest/concepts/pydantic_settings/#subcommands-and-positional-arguments

I suspect the core issue is the leaking of an argparse implementation detail (which  internally calls `sys.exit`).

This is also perhaps the source of inconsistencies in exit codes when an error occurs. Sometimes it's 1 and sometimes it's 2. 

## Acceptance Criteria

1. Remove usage of SystemExit in the docs
2. CLI Examples should adhere to the standard style for Pydantic error catching (e.g., `ValidationError`)

## Possible Solutions

1. In Python 3.9, `argparse` added `exit_on_error` flag  [to help solve](https://docs.python.org/3/library/argparse.html#exit-on-error) this friction point with `exit` being called. 
2. Subclass `argparse.ArgumentParser` and override `.exit()` or `error()` to raise the appropriate exception (from `status != 0`) instead of calling `sys.exit`.

hramezani:
Thanks @mpkocher for reporting this issue.

@kschwab Could you please take a look at this issue?

kschwab:
Yes, I agree with what @mpkocher raised. I'll work to push a fix in the next few days.

kschwab:
@mpkocher, I've modified the acceptance criteria one to:

> Remove usage of SystemExit in the docs **_for non-zero exit codes_**

Otherwise, any application that calls `--help` would raise an exception. i.e., this issue only removes exit on error in favor of raising a `SettingsError` exception.

kschwab:
Actually, the more I think on this the more I see it as a feature request to add a `cli_exit_on_error` flag. Hardcoding `exit_on_error=True` isn't ideal either. Therefore, we can:

1. Standardize CLI parsing errors to raise `SettingsError` exceptions.
2. Provide a `cli_exit_on_error` config that defaults to `True`, just like argparse.

mpkocher:
> @mpkocher, I've modified the acceptance criteria one to:
> 
> > Remove usage of SystemExit in the docs **_for non-zero exit codes_**
> 
> Otherwise, any application that calls `--help` would raise an exception. i.e., this issue only removes exit on error in favor of raising a `SettingsError` exception.

[...]

PR #389, opened 2024-09-10
Add CLI App Support

kschwab:
Adds utility functions for running CLI applications.

See [documentation](https://github.com/kschwab/pydantic-settings/blob/b47e1647fc067ca51f8a50abdd41568f72fc6282/docs/index.md#creating-cli-applications) for details.

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -808,11 +808,11 @@ assert get_subcommand(cmd).model_dump() == {
 ### Creating CLI Applications
 
 The `CliApp` class provides two utility methods, `CliApp.run` and `CliApp.run_subcommand`, that can be used to run a
-Pydantic `BaseSettings`, `BaseModel`, or `pydantic.dataclasses.dataclass` as a CLI application. Primarily, they provide
-structure for running `cli_cmd` methods associated with models.
+Pydantic `BaseSettings`, `BaseModel`, or `pydantic.dataclasses.dataclass` as a CLI application. Primarily, the methods
+provide structure for running `cli_cmd` methods associated with models.
 
-`CliApp.run` can be used to directly provide `cli_args` for parsing and will run the model `cli_cmd` method (if defined)
-after instantiation:
+`CliApp.run` can be used in directly providing the `cli_args` to be parsed, and will run the model `cli_cmd` method (if
+defined) after instantiation:
 
 ```py
 from pydantic_settings import BaseSettings, CliApp
@@ -822,9 +822,11 @@ class Settings(BaseSettings):
     this_foo: str
 
     def cli_cmd(self) -> None:
+        # Print the parsed data
         print(self.model_dump())
         #> {'this_foo': 'is such a foo'}
 
+        # Update the parsed data showing cli_cmd ran
         self.this_foo = 'ran the foo cli cmd'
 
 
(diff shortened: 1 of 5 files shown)


docs/index.md
```diff
@@ -560,19 +559,7 @@ print(Settings().model_dump())
...
-
-```py
-from pydantic_settings import BaseSettings
-
-
-class Settings(BaseSettings):
-    this_foo: str
-
-
-print(Settings(_cli_parse_args=['--this_foo', 'is such a foo']).model_dump())
```
kschwab, line 573: Fixed suggestion of using private param `_cli_parse_args`. The "formal" equivalent is now:

```python
CliApp.run(Settings, cli_args=['--this_foo', 'is such a foo']).model_dump()
```

docs/index.md
```diff
@@ -1211,13 +1287,15 @@ cli_settings = CliSettingsSource(Settings, root_parser=parser)
 
 # Parse and load CLI settings from the command line into the settings source.
 sys.argv = ['example.py', '--food', 'kiwi', '--name', 'waldo']
-print(Settings(_cli_settings_source=cli_settings(args=True)).model_dump())
```
kschwab, line 1214: Fixed suggestions of using private param `_cli_settings_source` for its formal equivalent.

pydantic_settings/main.py
```diff
@@ -33,7 +40,6 @@ class SettingsConfigDict(ConfigDict, total=False):
     env_parse_enums: bool | None
     cli_prog_name: str | None
     cli_parse_args: bool | list[str] | tuple[str, ...] | None
-    cli_settings_source: CliSettingsSource[Any] | None
```
kschwab, line 36: This was a bug. It's not possible to provide `cli_settings_source` as a configuration setting as the `settings_cls` cannot be defined (circular dependency).

hramezani:
Thanks @kschwab for this PR.

I have some concerns about the cli settings source, it has a lot of code and I think maintaining this amount of code would be hard in the future.

Specially about `CLI App`, Is it requested by somebody?

kschwab:
@hramezani, yes, this was requested and discussed in #335 and #336. It is related to discussion we had on [get_subcomand](https://github.com/pydantic/pydantic-settings/pull/341#issuecomment-2310615348), where this functionality for CLI is somewhat standard. I think drawing the line here for feature support makes sense.

Also, I understand the LOC concern for `CliSettingsSource`, however the root of this is in `sources.py`. With everything now in and understood, it can be simplified. I was planning on addressing this separately.

hramezani:
@kschwab Don't we need more tests for this patch? I can't see any test for `CliApp.run` in particular.  did I miss something?

kschwab:
@hramezani it's mostly a superficial commit. The doc tests cover happy paths, and I added tests in `test_settings.py` for exception paths. Would you like me to copy the doc tests into `test_settings.py`?

kschwab:
@hramezani, I added more tests specific to `CliApp.run` and updated all other test references using `_cli_parse_args=...` to instead use `CliApp.run(Settings, cli_args=...)` where relevant.

hramezani:
Thanks @kschwab.

I will check it next week

hramezani (approved):

hramezani:
Thanks @kschwab

Issue #259, opened 2024-03-21
Override default file in Yaml/Toml/Json ConfigSettingsSource at runtime.


theelderbeever:
I would like to override the defaulted config file path used in a file settings source. In the example below the default yaml file that is read in is `/etc/config.yaml` however, oftentimes I like to be able to pass the config file location in at application startup via a cli arg. Is there an ergonomic way to update this at runtime?

```python
import os
from typing import Tuple, Type

from pydantic import BaseModel, Field

from pydantic_settings import (
    BaseSettings,
    PydanticBaseSettingsSource,
    YamlConfigSettingsSource,
    EnvSettingsSource,
)


class Nested(BaseModel):
    nested_field: str = Field(default=...)


class Settings(BaseSettings):
    foobar: str = Field(default=...)
    nested: Nested = Field(default=...)

    @classmethod
    def settings_customise_sources(
        cls, settings_cls: Type[BaseSettings], **kwargs
    ) -> Tuple[PydanticBaseSettingsSource, ...]:
        return (
            EnvSettingsSource(
                settings_cls,
                env_prefix="APP__",
                env_nested_delimiter="__",
                case_sensitive=False,
            ),
            YamlConfigSettingsSource(settings_cls, yaml_file="/etc/config.yaml"),
        )


print(Settings().model_dump())
```

hramezani:
Thanks @theelderbeever for reporting this.

Right now there is no clear way to pass the file. We have to make it possible to pass the file like `Settings(_yaml_file='<file_name>')` but it will be a breaking change. So, we have to postpone it until `V3`.

BTW, you can do it in a hacky way like:
```py
import os
import sys
from typing import Tuple, Type

from pydantic import BaseModel, Field

from pydantic_settings import (
    BaseSettings,
    PydanticBaseSettingsSource,
    YamlConfigSettingsSource,
    EnvSettingsSource,
)

YAML_FILE_PATH = None


class Nested(BaseModel):
    nested_field: str = Field(default=...)


class Settings(BaseSettings):
    foobar: str = Field(default=...)
    nested: Nested = Field(default=...)

    @classmethod
    def settings_customise_sources(
        cls, settings_cls: Type[BaseSettings], **kwargs
    ) -> Tuple[PydanticBaseSettingsSource, ...]:
        global YAML_FILE_PATH

        return (
            EnvSettingsSource(
                settings_cls,
                env_prefix="APP__",
                env_nested_delimiter="__",
[...]

PR #402, opened 2024-09-13
Fix PathType typing in case of sequence

### LINKED ISSUE #401: Mypy reports error when passing a list to YamlConfigSettingsSource through yaml_file
**pydantic version:** 2.8.2
**pydantic-settings version:** 2.5.2
**OS:** elementary

When I try to use multiple settings sources, I get the following typechecking error : 
```
error: Argument "yaml_file" to "YamlConfigSettingsSource" has incompatible type "list[str]"; expected "Path | str | list[Path | str] | tuple[Path | str, ...] | None"  [arg-type]
```
This is because `list` is invariant so `list[str]` cannot be narrowed to `list[Path | str]` (as per [this issue](https://github.com/python/mypy/issues/3351)).
I'll post a PR to replace `list[Path | str] | tuple[Path | str]`, but if we specifically need to check for `list | tuple`, we might distribute list and tuple over the unions.

just-maiyak:
Fixes #401


pydantic_settings/sources.py
```diff
@@ -108,8 +108,8 @@ def import_azure_key_vault() -> None:
         ) from e
 
 
-DotenvType = Union[Path, str, List[Union[Path, str]], Tuple[Union[Path, str], ...]]
-PathType = Union[Path, str, List[Union[Path, str]], Tuple[Union[Path, str], ...]]
+DotenvType = Union[Path, str, Sequence[Union[Path, str]] ...]]
+PathType = Union[Path, str, Sequence[Union[Path, str]] ...]]
```
hramezani, line 112:
```suggestion
DotenvType = Union[Path, str, Sequence[Union[Path, str]] ...]
PathType = Union[Path, str, Sequence[Union[Path, str]] ...]
```

hramezani:
I just fixed the syntax error to let the CI run

pydantic_settings/sources.py
```diff
@@ -108,8 +108,8 @@ def import_azure_key_vault() -> None:
         ) from e
 
 
-DotenvType = Union[Path, str, List[Union[Path, str]], Tuple[Union[Path, str], ...]]
-PathType = Union[Path, str, List[Union[Path, str]], Tuple[Union[Path, str], ...]]
+DotenvType = Union[Path, str, Sequence[Union[Path, str]] ...]
+PathType = Union[Path, str, Sequence[Union[Path, str]] ...]
```
hramezani, line 112:
```suggestion
DotenvType = Union[Path, str, Sequence[Union[Path, str]], ...]
PathType = Union[Path, str, Sequence[Union[Path, str]], ...]
```

just-maiyak:
Perfect thank you so much, I amended without staging the changes and after that I didn't have my laptop on hand 😅

pydantic_settings/sources.py
```diff
@@ -108,8 +108,8 @@ def import_azure_key_vault() -> None:
         ) from e
 
 
-DotenvType = Union[Path, str, List[Union[Path, str]], Tuple[Union[Path, str], ...]]
-PathType = Union[Path, str, List[Union[Path, str]], Tuple[Union[Path, str], ...]]
+DotenvType = Union[Path, str, Sequence[Union[Path, str]], ...]
+PathType = Union[Path, str, Sequence[Union[Path, str]], ...]
```
hramezani, line 112:
```suggestion
DotenvType = Union[Path, str, Sequence[Union[Path, str]]]
PathType = Union[Path, str, Sequence[Union[Path, str]]]
```

just-maiyak:
I fixed the CI, there were still a couple of unused imports. It should be good to merge :)

hramezani:
Thanks @just-maiyak for updating this PR.

Could you provide a simple example that has mypy error? I am going to test the error and the fix?

Thanks!

just-maiyak:
Sure, if you run mypy on this file  :

hramezani:
Thanks @just-maiyak for providing the example.

The model that you provided looks weird to me. Could you explain why you are using `YamlConfigSettingsSource` in the `model_config`?

just-maiyak:
I'm trying to read settings from a list of yaml files (just like we would from a .env file). I'm following the docs right here : [Other Settings Source](https://docs.pydantic.dev/latest/concepts/pydantic_settings/#other-settings-source).

hramezani:
But this should be like the first example in the page:

```py

    BaseSettings,
    PydanticBaseSettingsSource,
    SettingsConfigDict,
    TomlConfigSettingsSource,
)


class Nested(BaseModel):
    nested_field: str


class Settings(BaseSettings):
    foobar: str
    nested: Nested
    model_config = SettingsConfigDict(toml_file='config.toml')

    @classmethod
        cls,
        settings_cls: Type[BaseSettings],
    ) -> Tuple[PydanticBaseSettingsSource, ...]:
        return (TomlConfigSettingsSource(settings_cls),)
```

You only need to change the `TomlConfigSettingsSource` with `YamlConfigSettingsSource `

just-maiyak:
Alright, it's not very clear from the docs that you have to keep `settings_customize_sources`. It's also not so clear how `SettingsConfigDict` works. Is the latter built from calling `settings_customize_sources` ?

hramezani:
> Alright, it's not very clear from the docs that you have to keep settings_customize_sources

This is because yaml settings source is not part of the default settings sources and you need to somehow override the default behavior to enable this source. we decided to use this approach to prevent breaking change in V2.
Here is the [doc](https://docs.pydantic.dev/latest/concepts/pydantic_settings/#customise-settings-sources).

BTW, any improvement on docs is welcome. feel free to create a PR if you want. You don't need to create a new issue.

Thanks!

just-maiyak:
I'll work on a docs contribution :) Thanks for helping out !

just-maiyak:
Sorry for being so persistent ... I'm reopening this again because I still thing the issue is happening in the case I start from the sample you provide and parametrize the `yaml_file` like so : 

```python

    BaseSettings,
    PydanticBaseSettingsSource,
    SettingsConfigDict,
    TomlConfigSettingsSource,
)


class Nested(BaseModel):
    nested_field: str


class Settings(BaseSettings):
    foobar: str
    nested: Nested

    config_files: list[str] = ["config.yaml"] # This might be the return of a funtion or initialized in __init__
    model_config = SettingsConfigDict(toml_file=config_files)

    @classmethod
        cls,
        settings_cls: Type[BaseSettings],
    ) -> Tuple[PydanticBaseSettingsSource, ...]:
        return (TomlConfigSettingsSource(settings_cls),)
```
Should this be done in another way ?

hramezani:
I think the `config_files: list[str] = ["config.yaml"]` is not the correct way. You probably did it like this because the config file name is not specified and you want to pass it when you initialize the model.

There is an issue for this problem https://github.com/pydantic/pydantic-settings/issues/259
I am going to prepare a PR for this issue but before that probably you can use one of the workaround suggested in the issue.

just-maiyak:
That's my case indeed, want to provide additional files at runtime. Unfortunately, I still have the same typing problem even with the workaround as soon as I wrap the settings in a `list`.

I still think the `Sequence` type is more appropriate if write operations to that variable are not possible (which I think is the case because it seems we only construct settings once as soon as we read that `yaml_config` parameter).

What also works is to explicitly type whatever we're passing as `list[Path | str]` but that introduces unnecessary noise imo.

hramezani (approved):

PR #405, opened 2024-09-14
Add `cli_ignore_unknown_args` config option.

### LINKED ISSUE #404: Ignore unknown sys.argv
I really like the ability to override args from my config from the cli, but it breaks a lot of other things that also rely on sys.argv like pytest and uvicorn. I have tried setting cli_exit_on_error=False, but it doesn't seem to stop it from blowing up with args that are not declared in the model if I have cli_parse_args=True.

How can I have it just ignore args it does not know so I can use it with pytest, uvicorn, etc...?

kschwab:
Fixes #404.

Adds a config that will ignore unknown CLI args, equivalent to [argparse.parse_known_args](https://docs.python.org/3/library/argparse.html#argparse.ArgumentParser.parse_known_args).

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -979,7 +979,7 @@ class ImplicitSettings(BaseSettings, cli_parse_args=True, cli_implicit_flags=Tru
     """
 ```
 
-#### Ignore Unknown Args
+#### Ignore Unknown Arguments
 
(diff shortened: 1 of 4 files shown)


hramezani (approved):

hramezani:
Thanks @kschwab

PR #418, opened 2024-09-21
Add `cli_flag_prefix_char` config option.

### LINKED ISSUE #414: Feature: Make argument flag configurable
Hi, would it be possible to make the argument flag (the `--` prefix before argument names) configurable? I saw that it is hardcoded here:
https://github.com/pydantic/pydantic-settings/blob/7e7ccdbc7b27d3f3d292d35a3b5033ef740fd6f7/pydantic_settings/sources.py#L1624

Our use case is that we are migrating some tools from omegaconf to pydantic and would like to use pydantic-settings to parse CLI arguments. My understanding is that pydantic-settings always adds a `--` prefix before argument names whereas omegaconf doesn't add this prefix. It would be great to have an option to disable the prefix in pydantic-settings to make the transition backwards compatible.

I also saw that `--` is used in another place but am not sure if this would be relevant for this feature:
https://github.com/pydantic/pydantic-settings/blob/7e7ccdbc7b27d3f3d292d35a3b5033ef740fd6f7/pydantic_settings/sources.py#L1506

kschwab:
Fixes #414.

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -1207,6 +1207,31 @@ sub_model options:
 """
 ```
 
+#### Change the CLI Flag Prefix Character
+
+Change The CLI flag prefix character used in CLI optional arguments by settings `cli_flag_prefix_char`.
+
+```py
+import sys
+
+from pydantic import AliasChoices, Field
+
+from pydantic_settings import BaseSettings
+
+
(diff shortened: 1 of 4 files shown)


hramezani (approved):

hramezani:
Thanks @kschwab 🙏

PR #432, opened 2024-10-01
Use SettingsErrors when file can't be used as source

l00ptr:
Some file formats (at least yaml) allow to represent non dictionnary values. In such situation we can't add the values read from those files. Instead of raising a ValueError, we now raise a SettingsError and indicate we can't parse the related config file.


hramezani:
Thanks @l00ptr for this PR. I think this will introduce a breaking change because some people code might be depend on the old error

l00ptr:
ok, i think i can try to help on that topic. I can start writing a simple test case for invalid yaml file and handle json decoding error.

dlax:
> I think this will introduce a breaking change because some people's code might depend on the old error.

It seems to me that the change would not break users code because `SettingsError` is a subclass of `ValueError`: https://github.com/pydantic/pydantic-settings/blob/0d605d076553452f67141f00e5dc873aa06c5b01/pydantic_settings/sources.py#L122
so anybody doing:
```python
try:
   s = Settings()
except ValueError:
   # handle exception here
   ...
```
would still have their code working. This could be checked in tests added here.

Or did you have something else in mind @hramezani ?

pydantic_settings/sources.py
```diff
@@ -1912,7 +1912,10 @@ def _read_files(self, files: PathType | None) -> dict[str, Any]:
         for file in files:
             file_path = Path(file).expanduser()
             if file_path.is_file():
-                vars.update(self._read_file(file_path))
+                try:
+                    vars.update(self._read_file(file_path))
+                except ValueError as e:
+                    raise SettingsError(f'Parsing error encountered for {file_path}: {e}')
```
dlax, line 1918: I think we could improve the user experience by explicitly checking that the result of file read is a dict because otherwise, (e.g. in added tests), the `ValueError`, *which is about `dict.update()`*, is not very informative:
```python
>>> d
{}
>>> d.update("abc")
  File "<stdin>", line 1, in <module>
ValueError: dictionary update sequence element #0 has length 1; 2 is required
```

Also, I realize that calling `dict.update()` with an iterable is also valid and so loading a YAML like:
```yaml
- - x
  - y
- - a
  - b
```
would work; not sure if it's expected? But maybe it's another issue.

hramezani:
> > I think this will introduce a breaking change because some people's code might depend on the old error.
> 
> It seems to me that the change would not break users code because `SettingsError` is a subclass of `ValueError`:
> 
> https://github.com/pydantic/pydantic-settings/blob/0d605d076553452f67141f00e5dc873aa06c5b01/pydantic_settings/sources.py#L122
> 
> 
> so anybody doing:
> ```python
> try:
>    s = Settings()
> except ValueError:
>    # handle exception here
>    ...
> ```
> 
> would still have their code working. This could be checked in tests added here.
> 
> Or did you have something else in mind @hramezani ?

I still think it will be a breaking change. some people code may 

> > I think this will introduce a breaking change because some people's code might depend on the old error.
> 
> It seems to me that the change would not break users code because `SettingsError` is a subclass of `ValueError`:
> 
> https://github.com/pydantic/pydantic-settings/blob/0d605d076553452f67141f00e5dc873aa06c5b01/pydantic_settings/sources.py#L122
> 
> 
> so anybody doing:
> ```python
> try:
>    s = Settings()
> except ValueError:
>    # handle exception here
>    ...
> ```
> 
> would still have their code working. This could be checked in tests added here.
> 
> Or did you have something else in mind @hramezani ?

Thanks @dlax for the explanation. However, it is still a breaking change because some people's code might depend on the error type `ValueError`(rare case). So, better not to do it now and postpone the decision before `V3`.

`pydantic-settings` had a lot of changes from V1 to V2 and we decided not to introduce breaking changes even something like this.

pydantic_settings/sources.py
l00ptr: we can probably check if the object returned by `self._read_file` is a dict or not and handle that specific case.

There are also a lot of cases where reading from (toml, yaml,...) file will raise ValueError.  Eg:

``` python
 import tomli

tomli.loads("demo =\ndemo2 =")
```

Will throw a tomli.TOMLDecodeError :/ finally i think we should simply keep the `ValueError`.

pydantic_settings/sources.py
```diff
@@ -1912,7 +1912,15 @@ def _read_files(self, files: PathType | None) -> dict[str, Any]:
...
-                vars.update(self._read_file(file_path))
+                try:
+                    settings = self._read_file(file_path)
+                except ValueError as e:
+                    raise SettingsError(f'Failed to parse settings from {file_path}, {e}')
+                if not isinstance(settings, dict):
+                    raise SettingsError(
+                        f'Failed to parse settings from {file_path}, expecting an object (valid dictionnary)'
+                    )
+                vars.update(settings)
```
hramezani, line 1923: @l00ptr Thanks for updating this.
As I mentioned before we can't merge this PR because it will introduce a breaking change. So, we have 2 options here:

1- Keep only the tests that test the current behavior(ValueError) and revert to raising `SettingsError`; then, we can merge the tests only.
2- Keep the PR open until `V3` and merge it before `V3`.

I would prefer option 1.

l00ptr: @hramezani i splitted those changes on 3  commits, let me know if i should create a specific MR for the 2 last commits (about SettingsError) and only keep the first one here.

hramezani: @l00ptr no need for a new PR.

Then we have to wait to merge it on `V3`. I will add a `v3` label to the PR to not merge it until `v3`

PR #436, opened 2024-10-06
Add support for suppressing fields from CLI help.

### LINKED ISSUE #431: Support for suppressing fields from the CLI --help message
In stdlib argparse you are able to set the help message to `argparse.SUPPRESS` to hide the entry from the generated `--help` output.

Would it be possible to support this in pydantic-settings as well?
It seems to work, if you add a 
```python
if help==SUPPRESS:
  pass
elif ...
```
here https://github.com/pydantic/pydantic-settings/blob/main/pydantic_settings/sources.py#L1885
and setting the field description to `argparse.SUPPRESS`

kschwab:
Fixes #431.


hramezani (approved):

hramezani:
Thanks @kschwab

PR #465, opened 2024-11-04
Use the class name in the `__repr__` implementations

dlax:
Previously, the `__repr__()` method of those settings source classes was inherited from InitSettingsSource's thus returning a misleading 'InitSettingsSource(init_kwargs={})'. We here define a specific `__repr__()` method implementation for classes using a config file (json, toml, yaml).

### DIFF
--- a/pydantic_settings/sources.py
+++ b/pydantic_settings/sources.py
@@ -349,7 +349,9 @@ class DefaultSettingsSource(PydanticBaseSettingsSource):
         return self.defaults
 
     def __repr__(self) -> str:
-        return f'DefaultSettingsSource(nested_model_default_partial_update={self.nested_model_default_partial_update})'
+        return (
+            f'{self.__class__.__name__}(nested_model_default_partial_update={self.nested_model_default_partial_update})'
+        )
 
 
 class InitSettingsSource(PydanticBaseSettingsSource):
@@ -383,7 +385,7 @@ class InitSettingsSource(PydanticBaseSettingsSource):
         )
 
     def __repr__(self) -> str:
-        return f'InitSettingsSource(init_kwargs={self.init_kwargs!r})'
+        return f'{self.__class__.__name__}(init_kwargs={self.init_kwargs!r})'
 
 
 class PydanticBaseEnvSettingsSource(PydanticBaseSettingsSource):
@@ -673,7 +675,7 @@ class SecretsSettingsSource(PydanticBaseEnvSettingsSource):
         return None, field_key, value_is_complex
 
     def __repr__(self) -> str:
-        return f'SecretsSettingsSource(secrets_dir={self.secrets_dir!r})'
+        return f'{self.__class__.__name__}(secrets_dir={self.secrets_dir!r})'
 
 
 class EnvSettingsSource(PydanticBaseEnvSettingsSource):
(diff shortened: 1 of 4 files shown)


tests/test_settings.py
```diff
@@ -1808,6 +1811,21 @@ def test_builtins_settings_source_repr():
...
+        (YamlConfigSettingsSource, 'YamlConfigSettingsSource(yaml_file=config.yaml)'),
+    ),
+)
+def test_configfile_settings_source_repr(cls: Type[PydanticBaseSettingsSource], expected: str) -> None:
+    ext = cls.__name__[:4].lower()
+    path = Path(f'config.{ext}')
+    source = cls(BaseSettings(), path)
+    assert repr(source) == expected
+
+
```
hramezani, line 1828: Please separate the tests and move them into 

- test_source_json.py
- test_source_toml.py
- test_source_yaml.py

hramezani:
Thanks @dlax for the PR. I left a comment

Viicos (changes requested):
Thanks for the contribution. Instead, let's change the implementation of `InitSettingsSource.__repr__` to:

```python
    def __repr__(self) -> str:
        return f'{self.__class__.__name__}(init_kwargs={self.init_kwargs!r})'
```

hramezani:
> ```python
> init_kwargs
> ```

Then we will have `JsonConfigSettingsSource(init_kwargs={})` for json config source. we don't have `init_kwargs` in this source.

Viicos:
Ah yes, well in this case probably best to keep it as is, but use `self.__class__.__name__` for each `__repr__` implementation to keep the class name and repr in sync

hramezani:
@dlax Please can you change all the `__repr__` to and use `self.__class__.__name__`?

tests/test_settings.py
```diff
@@ -1808,6 +1811,21 @@ def test_builtins_settings_source_repr():
     )
 
 
+def test_repr_json_config_settings_source() -> None:
```
hramezani, line 1814: as I mentioned in the previous comment, we have a separate test file for json config settings source `test_source_json.py`. please move the test to this file.

Also do the same for other sources tests

tests/test_settings.py
```diff
@@ -1808,6 +1811,21 @@ def test_builtins_settings_source_repr():
...
+    source = JsonConfigSettingsSource(BaseSettings(), Path('config.json'))
+    assert repr(source) == 'JsonConfigSettingsSource(json_file=config.json)'
+
+
+def test_repr_toml_config_settings_source() -> None:
```
hramezani, line 1819: This test has to be moved to `test_source_toml.py`

tests/test_settings.py
```diff
@@ -1808,6 +1811,21 @@ def test_builtins_settings_source_repr():
...
+    source = TomlConfigSettingsSource(BaseSettings(), Path('config.toml'))
+    assert repr(source) == 'TomlConfigSettingsSource(toml_file=config.toml)'
+
+
+def test_repr_yaml_config_settings_source() -> None:
```
hramezani, line 1824: This test has to be moved to `test_source_yaml.py`

tests/test_settings.py
dlax: Done.

hramezani (approved):

hramezani:
Thanks @dlax

PR #470, opened 2024-11-06
Fix rendering of annotations in code example

Viicos:
A space was missing and seems to be required when using blank lines in annotations.

The joys of mkd-material's syntax :smiling_face_with_tear: 

On main:

image


sydney-runkle (approved):
Thanks 👍

PR #481, opened 2024-11-26
Fix alias resolution to use preferred key.

kschwab:
Fixes alias resolution order for `EnvSettingsSource` and `SecretsSettingsSource`.

Issue was first observed in #469. See [comment](https://github.com/pydantic/pydantic-settings/issues/469#issuecomment-2491494347) for details.

PR #507, opened 2024-12-18
Move preferred alias resolution to private method

### LINKED ISSUE #504: EnvSettingsSource get_field_value returns field_name instead of the env name.
In 2.7.0 get_field_value returns the field_name instead of the env name, we depended on this logic, but now it makes it really hard to figure out which environment variable was used.

Can we have the env_name added back to EnvSettingsSource to look up the value?

I can see why the change was made, and it makes sense, but we are loosing data with it. We were using it to give feedback for users to track each configuration's value's source. eg.:   conf.value1 -> $MY_ENVVAR,  conf.value2 -> .env/VALUE2, etc...

Code block in question:
```
        for field_key, env_name, value_is_complex in field_infos:
            env_val = self.env_vars.get(env_name)
            if env_val is not None:
                if value_is_complex or (self.config.get('populate_by_name', False) and (field_key == field_name)):
                    preferred_key = field_key
                break

        return env_val, preferred_key, value_is_complex
```
 We are also fine if env_name is saved somewhere inside the class to lookup later.

kschwab:
Fixes #504 

Reverts #481 and instead moves preferred alias resolution into [private method](https://github.com/pydantic/pydantic-settings/pull/507/files#diff-b4b68ace3cb0cf2820d1d882735748b675a379c39463ccc89700a9618a80a2a2R565).

If desired, the private method could be made public in V3.

### DIFF
--- a/pydantic_settings/sources.py
+++ b/pydantic_settings/sources.py
@@ -268,7 +268,9 @@ class PydanticBaseSettingsSource(ABC):
         return self._settings_sources_data
 
     @abstractmethod
-    def get_field_value(self, field: FieldInfo, field_name: str) -> tuple[Any, str, bool]:
+    def get_field_value(
+        self, field: FieldInfo, field_name: str, use_preferred_alias: bool = False
+    ) -> tuple[Any, str, bool]:
         """
         Gets the value, the key for model creation, and a flag to determine whether value is complex.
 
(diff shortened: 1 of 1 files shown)


hramezani:
Thanks @kschwab for updating the PR.

Please add a comment somewhere and suggest what we should do in `V3`.

kschwab:
Sure @hramezani, added V3 docstring comment 👍🏾

hramezani (approved):

hramezani:
Thanks @kschwab

PR #519, opened 2025-01-10
CLI support for optional and variadic positional args

### LINKED ISSUE #517: Support default values for `CliPositionalArg` when parsing from the CLI
Currently, there doesn't appear to be a way to have an optional positional argument when parsing settings from the CLI. This is a blocker for me, trying to migrate over from `typer`.

Running the following will result in this error: `pydantic_settings.sources.SettingsError: positional argument BaseMainCommand.value has a default value`

```python3
from pydantic_settings import BaseSettings, CliPositionalArg, SettingsConfigDict


class Main(BaseSettings):
        cli_parse_args=True,
        cli_enforce_required=True,
    )

    value: CliPositionalArg[int | None] = None


parsed = Main() # Error is raised here.
print(parsed.value)
```

### LINKED ISSUE #518: Allow variadic positional args when parsing from CLI
Currently, it appears there's no way to take an arbitrary number of positional arguments when parsing settings from the command line. This is a blocker for me, migrating over from `typer`.

The closest you can get it by doing the following:

```python3
# test.py

from pydantic_settings import BaseSettings, CliPositionalArg, SettingsConfigDict


class Main(BaseSettings):
        cli_parse_args=True,
        cli_enforce_required=True,
    )

    values: CliPositionalArg[list[str]]


parsed = Main()
print(parsed.values)
```

However, this requires you to run it like `python test.py a,b,c` or `python test.py '["a", "b", "c"]'` rather than simply `python test.py a b c`, which often makes more sense if the command only takes in a single list of values.

It would be nice to have something like a `CliPositionalArgs` type to support this:

```python3
# test.py

from pydantic_settings import BaseSettings, CliPositionalArgs, SettingsConfigDict


class Main(BaseSettings):
        cli_parse_args=True,
        cli_enforce_required=True,
    )

    values: CliPositionalArgs[list[str]]


parsed = Main()
print(parsed.values)
```

kschwab:
Fixes #517
Fixes #518

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -1285,6 +1285,9 @@ However, if your use case [aligns more with #2](#command-line-support), using Py
 likely want required fields to be _strictly required at the CLI_. We can enable this behavior by using
 `cli_enforce_required`.
 
+!!! note
+    A required `CliPositionalArg` field is always strictly required (enforced) at the CLI.
+
 ```py
 import os
 import sys
--- a/docs/index.md
+++ b/docs/index.md
@@ -842,9 +842,10 @@ print(User().model_dump())
 
 ### Subcommands and Positional Arguments
 
-Subcommands and positional arguments are expressed using the `CliSubCommand` and `CliPositionalArg` annotations. These
-annotations can only be applied to required fields (i.e. fields that do not have a default value). Furthermore,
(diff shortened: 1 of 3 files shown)


hramezani (approved):

hramezani:
Thanks @kschwab

PR #529, opened 2025-01-24
Avoid using `Any` in `BaseSettings` signature to avoid mypy errors

Viicos:
When `--disallow-any-explicit`, this raises a mypy error when using the plugin (which synthesizes the `__init__` method).

Also use PEP 570 positional only argument for `self`.

Fixes https://github.com/pydantic/pydantic/issues/11313.


hramezani (approved):

hramezani:
Thanks @Viicos

Issue #51, opened 2023-05-10
Failing single underscore nested delimiter in BaseSettings with submodules


petroslamb:
### Initial Checks

- [X] I have searched GitHub for a duplicate issue and I'm sure this is something new
- [X] I have searched Google & StackOverflow for a solution and couldn't find anything
- [X] I have read and followed [the docs](https://docs.pydantic.dev) and still think this is a bug
- [X] I am confident that the issue is with pydantic (not my code, or another library in the ecosystem like [FastAPI](https://fastapi.tiangolo.com) or [mypy](https://mypy.readthedocs.io/en/stable))

### Description

Hi,

I am following the [example](https://docs.pydantic.dev/latest/usage/settings/#parsing-environment-variable-values) for nesting a `BaseSettings` class with `BaseModels` and I am using `env_nested_delimiter = '_'`. 

Running the example code I get:

```
pydantic.error_wrappers.ValidationError: 4 validation errors for Settings
sub_model -> v1
  field required (type=value_error.missing)
sub_model -> v2
  field required (type=value_error.missing)
sub_model -> v3
  field required (type=value_error.missing)
sub_model -> deep
  field required (type=value_error.missing)
```
While it looks possible to use the single underscore delimiter, the values are not picked up and I am not sure which ones the settings object is looking for in this example.

### Example Code

```Python
import os
from unittest import TestCase

from pydantic import BaseModel, BaseSettings


class DeepSubModel(BaseModel):
    v4: str


class SubModel(BaseModel):
    v1: str
    v2: bytes
    v3: int
    deep: DeepSubModel


class Settings(BaseSettings):
    v0: str
    sub_model: SubModel

    class Config:
        env_nested_delimiter = '_'
        env_prefix = 'TEST_'


class TestConfig(TestCase):

    def setUp(self) -> None:
        self.original_env = os.environ.copy()

    def tearDown(self) -> None:
        os.environ.clear()
        os.environ.update(self.original_env)

    def test_nested_delimiter(self):
   
        os.environ['TEST_V0'] = 'v0'
        os.environ['TEST_SUB_MODEL_V1'] = 'v1'
        os.environ['TEST_SUB_MODEL_V2'] = 'v2'
        os.environ['TEST_SUB_MODEL_V3'] = '3'
        os.environ['TEST_SUB_MODEL_DEEP_V4'] = 'v4'

        config = Settings()

        self.assertEqual(config.v0, 'v0')
[...]

PR #534, opened 2025-02-08
Add env_nested_max_split setting

gsakkis:
New setting to limit the parsing of nested env vars split by `env_nested_delimiter` up to a given nesting depth. 

The main use case is to allow `env_nested_delimiter='_'` to work with fields that contain underscores, as proposed [here](https://github.com/pydantic/pydantic-settings/issues/51#issuecomment-2642862688).

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -324,6 +324,58 @@ print(Settings().model_dump())
 `env_nested_delimiter` can be configured via the `model_config` as shown above, or via the
 `_env_nested_delimiter` keyword argument on instantiation.
 
+By default environment variables are split by `env_nested_delimiter` into arbitrarily deep nested fields. You can limit
+the depth of the nested fields with the `env_nested_depth` config setting. A common use case this is particularly useful
+is for two-level deep settings, where the `env_nested_delimiter` (usually a single `_`) may be a substring of model
+field names. For example:
+
+```bash
+# your environment
+export GENERATION_LLM_PROVIDER='anthropic'
+export GENERATION_LLM_API_KEY='your-api-key'
+export GENERATION_LLM_API_VERSION='2024-03-15'
+```
+
+You could load them into the following settings model:
+
+```py
+from pydantic import BaseModel
+
+from pydantic_settings import BaseSettings, SettingsConfigDict
+
+
+class LLMConfig(BaseModel):
+    provider: str = 'openai'
+    api_key: str
+    api_type: str = 'azure'
+    api_version: str = '2023-03-15-preview'
+
+
+class GenerationConfig(BaseSettings):
+    model_config = SettingsConfigDict(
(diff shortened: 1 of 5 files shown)


docs/index.md
```diff
@@ -324,6 +324,58 @@ print(Settings().model_dump())
 `env_nested_delimiter` can be configured via the `model_config` as shown above, or via the
 `_env_nested_delimiter` keyword argument on instantiation.
 
+By default environment variables are split by `env_nested_delimiter` into arbitrarily deep nested fields. You can limit
+the depth of the nested fields with the `env_nested_depth` config setting. A common use case this is particularly useful
+is for two-level deep settings, where the `env_nested_delimiter` (usually a single `_`) may be a substring of model
```
hramezani, line 329: what about `env_nested_max_split` or `env_max_split` as a new config name?

pydantic_settings/main.py
```diff
@@ -412,6 +422,7 @@ def _settings_build_values(
         env_file_encoding=None,
         env_ignore_empty=False,
         env_nested_delimiter=None,
+        env_nested_depth=-1,
```
hramezani, line 425: Can't we user default value of `None`?

pydantic_settings/sources.py
```diff
@@ -745,6 +746,10 @@ def __init__(
         self.env_nested_delimiter = (
             env_nested_delimiter if env_nested_delimiter is not None else self.config.get('env_nested_delimiter')
         )
+        self.env_nested_depth = (
+            env_nested_depth if env_nested_depth is not None else self.config.get('env_nested_depth', -1)
```
hramezani, line 750: better to use `None` as default value

pydantic_settings/sources.py
```diff
@@ -1063,7 +1072,8 @@ def __call__(self) -> dict[str, Any]:
     def __repr__(self) -> str:
         return (
             f'{self.__class__.__name__}(env_file={self.env_file!r}, env_file_encoding={self.env_file_encoding!r}, '
-            f'env_nested_delimiter={self.env_nested_delimiter!r}, env_prefix_len={self.env_prefix_len!r})'
+            f'env_nested_delimiter={self.env_nested_delimiter!r}, env_nested_depth={self.env_nested_depth}, '
```
hramezani, line 1075: we don't need to include the new config name here because it will introduce a breaking change.

hramezani:
Thanks @gsakkis for this PR.

I left some comments.

pydantic_settings/main.py
gsakkis: -1 is the default value of `maxsplit` in `str.split`. We can use `None` but we have to map it to -1 before passing it to `split`.

hramezani (approved):

Issue #535, opened 2025-02-11
Don't spit env vars if env_nested_delimiter=None


gsakkis:
By default `env_nested_delimiter=None` but this value is not special-cased, leading to surprising (if not ourtight buggy) behavior with `case_sensitive=True`:

```python
    class Subsettings(BaseSettings):
        foo: str

    class Settings(BaseSettings):
        subsettings: Subsettings

        model_config = SettingsConfigDict(case_sensitive=True)

    env.set('subsettingsNonefoo', 'abc')
    s = Settings()
    assert s.subsettings.foo == 'abc'
```

hramezani:
Thanks @gsakkis for reporting this issue.

The above example code that you provided returns `pydantic_settings.sources.SettingsError` before [your last commit](https://github.com/pydantic/pydantic-settings/commit/ccf99b2d78803f001930ef11e7ebefb1ec30d173). that behavior was wrong but could be fixed by putting the below code at the beginning of function:

```py
        if not self.env_nested_delimiter:
            return {}
```

This block of code can also fix the problem after you commit but I am afraid your change breaks people's code.
It would be great if you can take a look and provide a fix otherwise I will make a fix later.

PR #540, opened 2025-02-18
Don't explode env vars if env_nested_delimiter is empty

gsakkis:
Fixes https://github.com/pydantic/pydantic-settings/issues/535


hramezani:
Thanks @gsakkis

Issue #538, opened 2025-02-17
Can we provide better error message for user?


braindevices:
when there are multiple configuration source enabled the error message usually quite confusing.

I wonder if it is possible to expose some debug flag (maybe from ENV?) that allow user to easily see the dict from each source according to the priority?

hramezani:
Thanks @braindevices for this feature. Yeah, I think having such a flag for debugging would be great!

PR #544, opened 2025-02-19
[feature] Optionally Include settings source in ValidationError

AlbertDeFusco:
When working with settings configured for multiple sources it can be hard to pin down the exact source of a validation error. See #538 

In this PR I've added a configuration parameter `validate_each_source=True` that performs the validation while constructing the merged state of all configured sources. All validation errors are collected and raised after all sources have been parsed. The settings source is prepended to the `loc` field. Further, `type=missing` line errors are checked using the fully merged state and added to the raised ValidationError exception.

Here is an example using the Init and Env sources.

```python
import os

class MySettings(BaseSettings, validate_each_source=True, env_prefix='settings_'):
    foo: str
    default: int = 0
    extra: str | None = None

os.environ['settings_default'] = 'a'
settings = MySettings(extra=0)
```

The following error message is reported

```
  File "/Users/adefusco/Development/pydantic/pydantic-settings/testy.py", line 10, in <module>
    settings = MySettings(extra=0)
               ^^^^^^^^^^^^^^^^^^^
  File "/Users/adefusco/Development/pydantic/pydantic-settings/pydantic_settings/main.py", line 182, in __init__
    **self._settings_build_values(
      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/adefusco/Development/pydantic/pydantic-settings/pydantic_settings/main.py", line 451, in _settings_build_values
    raise ValidationError.from_exception_data(
pydantic_core._pydantic_core.ValidationError: 3 validation errors for MySettings
InitSettingsSource.extra
  Input should be a valid string [type=string_type, input_value=0, input_type=int]
    For further information visit https://errors.pydantic.dev/2.10/v/string_type
EnvSettingsSource.default
  Input should be a valid integer, unable to parse string as an integer [type=int_parsing, input_value='a', input_type=str]
    For further information visit https://errors.pydantic.dev/2.10/v/int_parsing
foo
  Field required [type=missing, input_value={'default': 'a', 'extra': 0}, input_type=dict]
    For further information visit https://errors.pydantic.dev/2.10/v/missing
```

### DIFF
--- a/pydantic_settings/main.py
+++ b/pydantic_settings/main.py
@@ -14,7 +14,7 @@ from pydantic._internal._signature import _field_name_for_signature
 from pydantic._internal._utils import deep_update, is_model_class
 from pydantic.dataclasses import is_pydantic_dataclass
 from pydantic.main import BaseModel
-from pydantic_core import ValidationError, InitErrorDetails
+from pydantic_core import InitErrorDetails, ValidationError
 
 from .sources import (
     ENV_FILE_SENTINEL,
@@ -417,15 +417,13 @@ class BaseSettings(BaseModel):
                 states[source_name] = source_state
                 state = deep_update(source_state, state)
 
-                if validate_each_source:
-                    if not source_state:
-                        continue
+                if source_state and validate_each_source:
                     try:
                         _ = super().__init__(**source_state)
                     except ValidationError as e:
                         line_errors = json.loads(e.json())
                         for line in line_errors:
-                            if line.get("type", "") == 'missing':
+                            if line.get('type', '') == 'missing':
                                 continue
                             line['loc'] = [source_name] + line['loc']
                             ctx = line.get('ctx', {})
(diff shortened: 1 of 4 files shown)


hramezani:
Thanks @AlbertDeFusco for this PR.

>‌In this PR I've added a configuration parameter validate_each_source=True that performs the validation while constructing the merged state of all configured sources. 

I think we decided to have a flag to print out collected values for each source.

AlbertDeFusco:
I can add that. Maybe an env var like `PYDANTIC_SETTINGS_DEBUG_SOURCE_INPUTS=1` to log the parsed source states?

hramezani:
> I can add that. Maybe an env var like `PYDANTIC_SETTINGS_DEBUG_SOURCE_INPUTS=1` to log the parsed source states?

Looks good. please make another PR for it.

BTW, I am not sure we want to add the `validate_each_source` configuration. I can't see any value in validating every source data separately.

AlbertDeFusco:
I can agree that validating each source is excessive. I'll work on some alternate implementations to provide source information in the ValidationError

hramezani:
> I can agree that validating each source is excessive. I'll work on some alternate implementations to provide source information in the ValidationError

JFYI, At the point of validation, we don't know which data comes from which sources because  `pydantic-settings` merges values from all the sources.

PR #547, opened 2025-02-20
Introduce `uv` for Project Management

KanchiShimono:
This Pull Request introduces [`uv`](https://docs.astral.sh/uv/) to improve project management. The key reasons for adopting `uv` are as follows:

1. **Unified Environment Management**  
   `uv` allows managing Python versions, virtual environments, and dependencies with a single tool, reducing the overhead of environment setup.

2. **Strict Dependency Management**  
   `uv` precisely manages dependencies based on OS, CPU architecture, and Python versions. In contrast, the current approach using `pip-compile` generates lock files tied to the developer's Python version, which can lead to unexpected issues (e.g., failing to resolve dependencies for libraries that have dropped support for older Python versions).

3. **High Performance**  
   With optimized dependency resolution and caching mechanisms, `uv` provides a fast and efficient package management experience.

4. **Consistency Across the Organization**  
   Since [Pydantic has already adopted `uv`](https://github.com/pydantic/pydantic/pull/10727), standardizing project management tools within the same organization reduces cognitive load for developers.


KanchiShimono:
The test failures are caused by [changes introduced in Pydantic 2.10](https://github.com/pydantic/pydantic/pull/10725) and are unrelated to the changes in this PR.

However, it was found that in Python 3.9 or earlier combined with Pydantic 2.10, `CliSettingsSource` does not correctly generate metavar.
https://github.com/pydantic/pydantic-settings/blob/4b6fd3d0968bfabffb8dd7c413b4e13ac35bd3d9/pydantic_settings/sources.py#L2001-L2005

As a workaround, applying the following change allows the tests to pass, but I am unsure if this is the appropriate solution.

```py
        elif isinstance(obj, WithArgsTypes) and sys.version_info >= (3, 9):
```

I believe it would be best to handle the test failures separately from this PR. What do you think?

hramezani:
> The test failures are caused by [changes introduced in Pydantic 2.10](https://github.com/pydantic/pydantic/pull/10725) and are unrelated to the changes in this PR.
> 
> However, it was found that in Python 3.9 or earlier combined with Pydantic 2.10, `CliSettingsSource` does not correctly generate metavar.
> 
> https://github.com/pydantic/pydantic-settings/blob/4b6fd3d0968bfabffb8dd7c413b4e13ac35bd3d9/pydantic_settings/sources.py#L2001-L2005
> 
> As a workaround, applying the following change allows the tests to pass, but I am unsure if this is the appropriate solution.
> 
> ```python
>         elif isinstance(obj, WithArgsTypes) and sys.version_info >= (3, 9):
> ```
> 
> I believe it would be best to handle the test failures separately from this PR. What do you think?

Thanks @KanchiShimono for this PR.

Yes, it is better to handle the test failure in separate PR.

I think @kschwab can fix the problem because he wrote most of the code there and he is the CLI expert

KanchiShimono:
@hramezani @kschwab 
Thank you for your reply.

1. I have created a new issue #551 regarding the incorrect generation of metavar.
2. As a workaround, I added a constraint on the Pydantic version used during development with Python 3.9 or earlier. This constraint is applied only when managing the development environment with uv and does not affect installations from PyPI as a library.
3. Tests for Python 3.9 and earlier use an older version of Pydantic, causing some print tests using `pytest-examples` to fail. To address this, I modified the tests to skip test_docs.py when an older version of Pydantic is detected.

hramezani:
@KanchiShimono you need to rebase.

we removed support for Python 3.8 and fixed the bug

KanchiShimono:
@hramezani 

I have rebased the branch to incorporate the latest changes from main.
All tests are now passing.

.github/workflows/ci.yml
```diff
@@ -98,21 +105,18 @@ jobs:
...
-
       - name: check GITHUB_REF matches package version
-        uses: samuelcolvin/check-python-version@v3.1
+        uses: samuelcolvin/check-python-version@v4.1
         with:
           version_file_path: pydantic_settings/version.py
 
       - name: build
-        run: python -m build
+        run: uv build
```
hramezani, line 119: Let's not change the `release` part. as we are using the same thing in pydantic.

hramezani:
@KanchiShimono  I requested a small change, other than that LGTM.

KanchiShimono:
@hramezani 
Thank you for your review.

I have borrowed the release part from the latest version of Pydantic. Additionally, I have upgraded the default Python version used in GitHub Actions to 3.12.

hramezani:
Thanks @KanchiShimono

PR #553, opened 2025-02-25
CLI metavar fix

### LINKED ISSUE #551: Incorrect metavar generation in `CliSettingsSource` with Python 3.9 and Pydantic 2.10
When using Python 3.9 or earlier with Pydantic 2.10, metavar in `CliSettingsSource` is not generated correctly.

This issue is caused by [changes introduced in Pydantic 2.10 related to `WithArgsTypes`](https://github.com/pydantic/pydantic/pull/10725). Since `CliSettingsSource` relies on `WithArgsTypes` for metavar generation, its behavior has changed unexpectedly.

https://github.com/pydantic/pydantic-settings/blob/4b6fd3d0968bfabffb8dd7c413b4e13ac35bd3d9/pydantic_settings/sources.py#L2001-L2005

This issue does not occur in Python 3.10 or later.

## Reproduction Code

```py
import sys
from typing import List

class Settings(BaseSettings):

    val: List[str]


sys.argv = ['main.py', '-h']
Settings()
```

When running this code, the output looks like this. The type of val is displayed as `typing.List[str][str]`, but it should be `List[str]`.

```
usage: main.py [-h] [--val typing.List[str][str]]

optional arguments:
  --val typing.List[str][str]
                        (required)
```

kschwab:
Fixes #551.

References `_repr` and `_typing_extra` changes from pydantic/pydantic#10725.

### DIFF
--- a/pydantic_settings/sources.py
+++ b/pydantic_settings/sources.py
@@ -39,7 +39,7 @@ import typing_extensions
 from dotenv import dotenv_values
 from pydantic import AliasChoices, AliasPath, BaseModel, Json, RootModel, Secret, TypeAdapter
 from pydantic._internal._repr import Representation
-from pydantic._internal._typing_extra import WithArgsTypes, is_type_alias_type, origin_is_union, typing_base
+from pydantic._internal._typing_extra import WithArgsTypes, origin_is_union, typing_base
 from pydantic._internal._utils import deep_update, is_model_class, lenient_issubclass
 from pydantic.dataclasses import is_pydantic_dataclass
 from pydantic.fields import FieldInfo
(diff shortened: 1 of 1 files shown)


Viicos:
Thanks, I'll handle it in https://github.com/pydantic/pydantic-settings/pull/556. @hramezani, if you want to have this fixed before https://github.com/pydantic/pydantic-settings/pull/556 (which requires dropping 3.8 first), @kschwab could you please instead only incorporate the following: https://github.com/pydantic/pydantic-settings/pull/556/files/f1c9d3c76ec086c29e6d45f1b6aec649c1fed0ea#diff-e0fd650be4a8ab8347244e6e58611b92dbe05fe35c742284d1a58d2520f70453?

PR #556, opened 2025-02-26
Switch to `typing-inspection`

Viicos:
Also fix some logic to be a bit more robust.

I'll note that the code base rely heavily on inspection of typing objects, but this fragile for several reasons (one example is when determining if a union is complex: this doesn't take into account PEP 695 type aliases).

Ideally most of the logic should rely on the core schema instead, but this isn't trivial as well.

Closes https://github.com/pydantic/pydantic-settings/pull/553.


pydantic_settings/sources.py
```diff
@@ -1996,19 +1997,20 @@ def _metavar_format_recurse(self, obj: Any) -> str:
             return '...'
         elif isinstance(obj, Representation):
             return repr(obj)
-        elif isinstance(obj, typing_extensions.TypeAliasType):
+        elif typing_objects.is_typealiastype(obj):
             return str(obj)
 
-        if not isinstance(obj, (typing_base, WithArgsTypes, type)):
+        origin = get_origin(obj)
```
Viicos, line 2003: This function is still a big mess, in the future I'll add a string repr function in `typing-inspection`

pydantic_settings/sources.py
```diff
@@ -2304,25 +2306,22 @@ def read_env_file(
...
-        if annotation is not None and issubclass(annotation, RootModel):
-            # In some rare cases (see test_root_model_as_field),
-            # the root attribute is not available. For these cases, python 3.8 and 3.9
-            # return 'RootModelRootType'.
-            root_annotation = annotation.__annotations__.get('root', None)
-            if root_annotation is not None and root_annotation != 'RootModelRootType':
-                annotation = root_annotation
-    except TypeError:
-        pass
+    if annotation is not None and _lenient_issubclass(annotation, RootModel) and annotation is not RootModel:
```
Viicos, line 2309: `_lenient_issubclass` takes care of the `TypeError`.

Also, checking for `annotation is not RootModel` and relying on `model_fields['root'].annotation` is more robust

pydantic_settings/sources.py
```diff
@@ -2365,22 +2364,23 @@ def _annotation_contains_types(
 
 
 def _strip_annotated(annotation: Any) -> Any:
```
Viicos, line 2366: Python takes care of flattening Annotated forms already.

pydantic_settings/utils.py
```diff
@@ -22,3 +27,22 @@ def path_type_label(p: Path) -> str:
...
+            # Up until Python 3.10, isinstance(<generic_alias>, type) is True
+            # (e.g. list[int])
+            return False
+        raise
+
+
+if sys.version_info < (3, 10):
+    _WithArgsTypes = tuple()
+else:
+    _WithArgsTypes = (_GenericAlias, types.GenericAlias, types.UnionType)
```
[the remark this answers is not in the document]
Viicos: Actually this was reported [here](https://github.com/pydantic/pydantic-settings/issues/551).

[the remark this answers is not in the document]
Viicos: I'm keeping the inconsistency here because the metavar format function is really messy and I can't reason properly when looking into it. It needs to be refactored and as mentioned in another comment this will probably be done in `typing-inspection`.

hramezani (approved):
Thanks @Viicos, Looks Great 👍

Issue #549, opened 2025-02-21
2.8.0 breaks mypy


sparkiegeek:
Our CI pulled in pydantic-settings 2.8.0 and it triggered an internal error in mypy.

We invoke mypy using `--install-types --non-interactive` and are using mypy 1.15.0

```
Traceback (most recent call last):
  File "mypy/semanal.py", line 7240, in accept
  File "mypy/nodes.py", line 1177, in accept
  File "mypy/semanal.py", line 1728, in visit_class_def
  File "mypy/semanal.py", line 1944, in analyze_class
  File "mypy/semanal.py", line 1991, in analyze_class_body_common
  File "mypy/semanal.py", line 2076, in apply_class_plugin_hooks
  File "/usr/local/lib/python3.11/site-packages/pydantic/mypy.py", line 184, in _pydantic_model_class_maker_callback
    return transformer.transform()
           ^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/site-packages/pydantic/mypy.py", line 527, in transform
    self.add_initializer(fields, config, is_settings, is_root_model)
  File "/usr/local/lib/python3.11/site-packages/pydantic/mypy.py", line 927, in add_initializer
    if arg_name.startswith('__') or not arg_name.startswith('_'):
       ^^^^^^^^^^^^^^^^^^^
AttributeError: 'NoneType' object has no attribute 'startswith'
```

Pinning to < 2.8.0 doesn't hit the issue.

sr-verde:
Yes, I can confirm that issue. We have this problem in one project, too. It shows up when trying to update from 2.7.1 to 2.8.0. Mypy points to this class definition:

```
class MongoDBURIs(BaseSettings):
    db_one: str = "mongodb://localhost:27017/db1"
    db_two: str = "mongodb://localhost:27017/db2"
```

mv1005:
I had the same issue and it apparently was caused by an incompatible combination of `pydantic` (pulled in pinned to an older version by another package) and `padyntic-settings`. After upgrading `pydantic`, the issue no longer showed up.

hramezani:
Thanks @mv1005 for the update.

Probably we need to pin pydantic to > 2.10

What do you think @Viicos?

Viicos:
I investigated, indeed an unsafe static type checking error fixed in https://github.com/pydantic/pydantic/issues/10669 (included in 2.10) is the reason of this regression.

[...]

PR #557, opened 2025-02-26
Revert usage of positional only argument in `BaseSettings.__init__`

Viicos:
Fixes https://github.com/pydantic/pydantic-settings/issues/549.


hramezani (approved):

hramezani:
Thanks @Viicos

PR #560, opened 2025-02-28
Drop support for Python 3.8

Viicos:
Required for https://github.com/pydantic/pydantic-settings/pull/556.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -32,7 +32,7 @@ jobs:
       fail-fast: false
       matrix:
         os: [ubuntu-latest, macos-latest, windows-latest]
-        python: ['3.8', '3.9', '3.10', '3.11', '3.12', '3.13']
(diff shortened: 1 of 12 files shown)


hramezani:
Thanks @Viicos for this PR.

we need to regenerate the requirements file with Python 3.9
https://github.com/pydantic/pydantic-settings/blob/5f33b6205608ccd51828cf52656b9b37f6c79db8/requirements/linting.txt#L2

there is a PR that [introduces UV to the project](https://github.com/pydantic/pydantic-settings/pull/547). we can probably leave the requirements file as is because they will be removed there

Viicos:
> there is a PR that [introduces UV to the project](https://github.com/pydantic/pydantic-settings/pull/547). we can probably leave the requirements file as is because they will be removed there

Sounds good to me.

hramezani (approved):

PR #578, opened 2025-04-07
Azure Key Vault: Don't load disabled secret

AndreuCodina:


hramezani:
Thanks @AndreuCodina for the PR.

Could you please add a test for `enabled=False`?

AndreuCodina:
> Thanks @AndreuCodina for the PR.
> 
> Could you please add a test for `enabled=False`?

If I don't mock `.get_secret`, I never get a value in `settings`, and if I mock it, I always get the mocked value. I don't know how it can be tested.

hramezani:
> > Thanks @AndreuCodina for the PR.
> > Could you please add a test for `enabled=False`?
> 
> If I don't mock `.get_secret`, I never get a value in `settings`, and if I mock it, I always get the mocked value. I don't know how it can be tested.

You are mocking `list_properties_of_secrets` return values like:

```py
        expected_secrets = [
            type('', (), {'name': 'SqlServerUser', 'enabled': True}),
            type('', (), {'name': 'SqlServer--Password', 'enabled': True}),
        ]
```

You can add a new test and change the `enabled` value to `False`, then we expect a validation error because the value is not enabled. right?

AndreuCodina:
> > > Thanks @AndreuCodina for the PR.
> > > Could you please add a test for `enabled=False`?
> > 
> > 
> > If I don't mock `.get_secret`, I never get a value in `settings`, and if I mock it, I always get the mocked value. I don't know how it can be tested.
> 
> You are mocking `list_properties_of_secrets` return values like:
> 
> ```python
>         expected_secrets = [
>             type('', (), {'name': 'SqlServerUser', 'enabled': True}),
>             type('', (), {'name': 'SqlServer--Password', 'enabled': True}),
>         ]
> ```
> 
> You can add a new test and change the `enabled` value to `False`, then we expect a validation error because the value is not enabled. right?

I tried that, but I get this:

```json
{'SqlServerPassword': 'SecretValue', 'DisabledSqlServerPassword': 'SecretValue'}
````

This is the test:

```python
    def test_do_not_load_disabled_secrets(self, mocker: MockerFixture) -> None:
            """AzureKeyVault settings."""

            SqlServerPassword: str
            DisabledSqlServerPassword: str

        expected_secrets = [
            type('', (), {'name': 'SqlServerPassword', 'enabled': True}),
            type('', (), {'name': 'DisabledSqlServerPassword', 'enabled': False}),
        ]
        mocker.patch(
            f'{AzureKeyVaultSettingsSource.__module__}.{SecretClient.list_properties_of_secrets.__qualname__}',
            return_value=expected_secrets,
        )
        mocker.patch(
            f'{AzureKeyVaultSettingsSource.__module__}.{SecretClient.get_secret.__qualname__}',
            return_value=KeyVaultSecret(SecretProperties(), 'SecretValue'),
        )
        obj = AzureKeyVaultSettingsSource(
            AzureKeyVaultSettings, 'https://my-resource.vault.azure.net/', DefaultAzureCredential()
        )

        settings = obj()

        assert 'SqlServerPassword' in settings
        assert 'DisabledSqlServerPassword' not in settings
```

hramezani:
Thanks @AndreuCodina

PR #579, opened 2025-04-07
Cli subcommand with annotation

### LINKED ISSUE #573: Discriminated Unions and Nested CLI Subcommands
Utilising a use case that implements strategy patterns beyond what mutually exclusive groups can offer, we might write the following code 
```
$ cat cli.py
from typing import Literal, Union, Annotated

from pydantic_settings import BaseSettings, SettingsConfigDict, CliSubCommand, CliApp


class Cat(BaseModel):
    animal_type: Literal['cat'] = Field(..., alias='animal-type')

    def cli_cmd(self):
        print("meow")


class Dog(BaseModel):
    animal_type: Literal['dog'] = Field(..., alias='animal-type')

    def cli_cmd(self):
        print("woof")


class Settings(BaseSettings):
    model_config = SettingsConfigDict()

    animals: CliSubCommand[Annotated[Union[Cat, Dog], Field(discriminator="animal_type")]]

    def cli_cmd(self):
        CliApp.run_subcommand(self)


if __name__ == "__main__":
    CliApp.run(Settings, cli_args=["animals", "--animal-type", "cat"])
```

Which fails with the following error:
`pydantic_settings.sources.SettingsError: subcommand argument Settings.animals has type not derived from BaseModel`
because of the Annotated wrapper

Unless I am using this wrong, I don't think this is by design. It would make what I am currently building significantly easier if this were to work, as I have a nested interface but many of the object paths should only be set under certain strategies. 

However, the following non-cli settings management works:
```
from typing import Literal, Union, Annotated

class Cat(BaseModel):
    animal_type: Literal['cat']


class Dog(BaseModel):
    animal_type: Literal['dog']


class Settings(BaseSettings):
    model_config = SettingsConfigDict(env_nested_delimiter="__")

    animals: Annotated[Union[Cat, Dog], Field(discriminator="animal_type")]


if __name__ == "__main__":
    import os

    os.environ["animals__animal_type"] = "dog"

    settings = Settings()

    print(settings.model_dump())
```

```
$ uv tree
Resolved 8 packages in 1ms
pydantic-cli-test v0.1.0
├── pydantic v2.11.1
│   ├── annotated-types v0.7.0
│   ├── pydantic-core v2.33.0
│   │   └── typing-extensions v4.13.0
│   ├── typing-extensions v4.13.0
│   └── typing-inspection v0.4.0
│       └── typing-extensions v4.13.0
└── pydantic-settings v2.8.1
    ├── pydantic v2.11.1 (*)
    └── python-dotenv v1.1.0
```

kschwab:
Fixes #573

### DIFF
--- a/pydantic_settings/sources/utils.py
+++ b/pydantic_settings/sources/utils.py
@@ -102,6 +102,19 @@ def _annotation_contains_types(
     return annotation in types
 
 
+def _get_class_types(annotation: type[Any]) -> list[type[Any]]:
+    origin = get_origin(annotation)
+    if typing_objects.is_union(origin):
+        types = []
+        for arg in get_args(annotation):
+            types.extend(_get_class_types(arg))
+        return types
+    elif typing_objects.is_annotated(origin):
+        return _get_class_types(get_args(annotation)[0])
+    else:
+        return [annotation]
+
+
 def _strip_annotated(annotation: Any) -> Any:
     if typing_objects.is_annotated(get_origin(annotation)):
         return annotation.__origin__
@@ -188,6 +201,7 @@ __all__ = [
     '_annotation_is_complex',
     '_annotation_is_complex_inner',
     '_get_alias_names',
+    '_get_class_types',
     '_get_env_var_key',
     '_get_model_fields',
     '_is_function',
--- a/pydantic_settings/sources/utils.py
+++ b/pydantic_settings/sources/utils.py
@@ -54,6 +54,10 @@ def _annotation_is_complex(annotation: type[Any] | None, metadata: list[Any]) ->
 
     origin = get_origin(annotation)
 
+    # Check if annotation is of the form Union[type, ...].
+    if typing_objects.is_union(origin):
+        return _union_is_complex(annotation, metadata)
+
     # Check if annotation is of the form Annotated[type, metadata].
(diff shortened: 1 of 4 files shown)


pydantic_settings/sources/utils.py
```diff
@@ -54,6 +54,10 @@ def _annotation_is_complex(annotation: type[Any] | None, metadata: list[Any]) ->
 
     origin = get_origin(annotation)
 
+    # Check if annotation is of the form Union[type, ...].
+    if typing_objects.is_union(origin):
+        return _union_is_complex(annotation, metadata)
```
kschwab, line 59: @hramezani From what I observed, it appears to be related to incorrectly determining that nested annotated unions are not complex. This change fixed it, but causes other tests fail. I'm not sure if it is the correct fix. Can you take a look?

hramezani: Sorry @kschwab for the delay.

I am also not sure what the correct fix is here. As the change breaks some existing tests, I would prefer to remove it.

kschwab: No worries @hramezani. I'm ok with closing this as a won't fix as well.

kschwab:
Closing as won't fix due to maintenance and complexity concerns.

PR #581, opened 2025-04-09
CLI JSON Optional Default

### LINKED ISSUE #575: CLI: creating an optional nested model with all default values
I've been trying (and failing) to find a way to allow nested models with default values for every field to be optional by default or signal creation of a default instance.  For example, given:

```python
class Nested(BaseModel):
    foo: int = 1
    bar: int = 2

class Options(BaseSettings):
    nested: Optional[Nested] = None
```

I'd like to enable three invocations:
```bash
$ options   # nested is None, app ignores it

$ options --nested   # create a fully default nested

$ options --nested.foo 3   # create modified nested
```

It appears the best I can do is require someone pass at least one default value directly or allow json and pass an empty dict:

```bash
$ options --nested.foo 1   # same as default, works but redundant, error prone

$ options --nested {}   # works but not intuitive for users.  can the braces be eliminated?
```

Is there a way to do this I'm overlooking or could it be a possible enhancement? 
Thanks.

kschwab:
Fixes #575

### DIFF
--- a/pydantic_settings/sources/providers/cli.py
+++ b/pydantic_settings/sources/providers/cli.py
@@ -908,7 +908,9 @@ class CliSettingsSource(EnvSettingsSource, Generic[T]):
         preferred_alias = alias_names[0]
         if not self.cli_avoid_json:
             added_args.append(arg_names[0])
-            kwargs['help'] = f'set {arg_names[0]} from JSON string'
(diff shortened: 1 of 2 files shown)


hramezani (approved):

hramezani:
Thanks @kschwab

PR #587, opened 2025-04-13
CLI submodel suppress.

### LINKED ISSUE #586: CliSuppress does not work on sub-models
Hi,

The fields of a sub-model annotated with `CliSuppress` are still visible in the CLI help message. I would expect them to be hidden as well.

Example code:
```py
import pydantic
import pydantic_settings

class SubModel(pydantic.BaseModel):
    hidden_a: int
    hidden_b: int

class Settings(pydantic_settings.BaseSettings, cli_parse_args=True):
    visible_str: str
    hidden_str: pydantic_settings.CliSuppress[str]
    hidden_obj: pydantic_settings.CliSuppress[SubModel]

settings = Settings()
```

Result:
```
usage: test.py [-h] [--visible_str str] [--hidden_obj JSON] [--hidden_obj.hidden_a int] [--hidden_obj.hidden_b int]

options:
  --visible_str str     (required)

hidden_obj options:
  --hidden_obj JSON     set hidden_obj from JSON string
  --hidden_obj.hidden_a int
                        (required)
  --hidden_obj.hidden_b int
                        (required)
```

Reproduced with pydantic-settings 2.8.1

Thanks!

kschwab:
Fixes #586

### DIFF
--- a/pydantic_settings/sources/providers/cli.py
+++ b/pydantic_settings/sources/providers/cli.py
@@ -665,6 +665,7 @@ class CliSettingsSource(EnvSettingsSource, Generic[T]):
         group: Any,
         alias_prefixes: list[str],
         model_default: Any,
+        is_model_suppressed: bool = False,
     ) -> ArgumentParser:
         subparsers: Any = None
         alias_path_args: dict[str, str] = {}
@@ -738,7 +739,7 @@ class CliSettingsSource(EnvSettingsSource, Generic[T]):
                 is_parser_submodel = sub_models and not is_append_action
                 kwargs: dict[str, Any] = {}
                 kwargs['default'] = CLI_SUPPRESS
-                kwargs['help'] = self._help_format(field_name, field_info, model_default)
+                kwargs['help'] = self._help_format(field_name, field_info, model_default, is_model_suppressed)
                 kwargs['metavar'] = self._metavar_format(field_info.annotation)
                 kwargs['required'] = (
                     self.cli_enforce_required and field_info.is_required() and model_default is PydanticUndefined
@@ -782,6 +783,7 @@ class CliSettingsSource(EnvSettingsSource, Generic[T]):
                         field_info,
                         alias_names,
                         model_default=model_default,
+                        is_model_suppressed=is_model_suppressed,
                     )
                 elif not is_alias_path_only:
                     if group is not None:
@@ -869,6 +871,7 @@ class CliSettingsSource(EnvSettingsSource, Generic[T]):
         field_info: FieldInfo,
         alias_names: tuple[str, ...],
         model_default: Any,
+        is_model_suppressed: bool,
     ) -> None:
         if issubclass(model, CliMutuallyExclusiveGroup):
             # Argparse has deprecated "calling add_argument_group() or add_mutually_exclusive_group() on a
@@ -906,9 +909,10 @@ class CliSettingsSource(EnvSettingsSource, Generic[T]):
                 model_group_kwargs['description'] = desc_header
 
         preferred_alias = alias_names[0]
+        is_model_suppressed = self._is_field_suppressed(field_info) or is_model_suppressed
         if not self.cli_avoid_json:
             added_args.append(arg_names[0])
-            kwargs['help'] = f'set {arg_names[0]} from JSON string'
+            kwargs['help'] = CLI_SUPPRESS if is_model_suppressed else f'set {arg_names[0]} from JSON string'
             model_group = self._add_group(parser, **model_group_kwargs)
             self._add_argument(model_group, *(f'{flag_prefix}{name}' for name in arg_names), **kwargs)
         for model in sub_models:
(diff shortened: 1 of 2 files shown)


hramezani (approved):

pydantic_settings/sources/providers/cli.py
```diff
@@ -906,11 +909,12 @@ def _add_parser_submodels(
                 model_group_kwargs['description'] = desc_header
 
         preferred_alias = alias_names[0]
+        is_model_suppressed = self._is_field_suppressed(field_info) or is_model_suppressed
         if not self.cli_avoid_json:
             added_args.append(arg_names[0])
             kwargs['nargs'] = '?'
             kwargs['const'] = '{}'
-            kwargs['help'] = f'set {arg_names[0]} from JSON string (default: {{}})'
+            kwargs['help'] = kwargs['help'] = CLI_SUPPRESS if is_model_suppressed else f'set {arg_names[0]} from JSON string (default: {{}})'
```
hramezani, line 917:
```suggestion
            kwargs['help'] = kwargs['help'] = (
                CLI_SUPPRESS if is_model_suppressed else f'set {arg_names[0]} from JSON string (default: {{}})'
            )
```

tests/test_source_cli.py
```diff
@@ -2151,10 +2167,18 @@ class Settings(BaseSettings, cli_parse_args=True):
 
         assert (
             capsys.readouterr().out
-            == f"""usage: example.py [-h]
+            == f"""usage: example.py [-h] [--visible_obj JSON] [--visible_obj.visible_a int]
```
hramezani:
```suggestion
            == f"""usage: example.py [-h] [--visible_obj [JSON]] [--visible_obj.visible_a int]
```

tests/test_source_cli.py
```diff
@@ -2151,10 +2167,18 @@ class Settings(BaseSettings, cli_parse_args=True):
...
-            == f"""usage: example.py [-h]
+            == f"""usage: example.py [-h] [--visible_obj [JSON]] [--visible_obj.visible_a int]
+                  [--visible_obj.visible_b int]
 
 {ARGPARSE_OPTIONS_TEXT}:
-  -h, --help  show this help message and exit
+  -h, --help            show this help message and exit
+
+visible_obj options:
+  --visible_obj JSON    set visible_obj from JSON string
```
hramezani, line 2177:
```suggestion
  --visible_obj [JSON]  set visible_obj from JSON string (default: {{}})
```

Geo5:
Hey, thanks for working on this already!
I think the fix is not quite enough, because it fails to do what it is supposed to do if you add a description to the SubModel (either via `Field(description=...)` or as class docstring and setting `cli_use_class_docs_for_groups=True`).

I think the fix would be to also set `CLI_SUPPRESS` to the subparser `description` field here:
https://github.com/pydantic/pydantic-settings/blob/18747409238dbba5c4ba9779c46353648f79f7ca/pydantic_settings/sources/providers/cli.py#L899
e.g.:
```python
        is_model_suppressed = self._is_field_suppressed(field_info) or is_model_suppressed
        model_group_kwargs['description'] = CLI_SUPPRESS if is_model_suppressed else field_info.description
```
like doing it a bit later in this function when adding the json argument:
https://github.com/pydantic/pydantic-settings/blob/18747409238dbba5c4ba9779c46353648f79f7ca/pydantic_settings/sources/providers/cli.py#L925
I am not sure if this is documented behavior by argparse, but i looked at the source and this is what it does. ¹

-----------------------
¹ See here:
https://github.com/python/cpython/blob/main/Lib/argparse.py#L252
Which gets called here:
https://github.com/python/cpython/blob/main/Lib/argparse.py#L2622
With the `description` field, which leads to the section returning no help string here:
https://github.com/python/cpython/blob/main/Lib/argparse.py#L222
if all argument texts are also empty (which they are with the above pull request, as all SubModel fields behave as if they had `CLI_SUPPRESS` explicitly)

Geo5:
The test can be made to fail with this test changes:
```diff
diff --git a/tests/test_source_cli.py b/tests/test_source_cli.py
index 6c523ff..6eb7421 100644
--- a/tests/test_source_cli.py
+++ b/tests/test_source_cli.py
@@ -2153,6 +2153,8 @@ def test_cli_app_exceptions():
 
 def test_cli_suppress(capsys, monkeypatch):
     class DeepHiddenSubModel(BaseModel):
+        """DeepHiddenSubModel class docstring"""
+
         deep_hidden_a: int
         deep_hidden_b: int
 
@@ -2166,10 +2168,10 @@ def test_cli_suppress(capsys, monkeypatch):
         visible_b: int
         deep_hidden_obj: CliSuppress[DeepHiddenSubModel]
 
-    class Settings(BaseSettings, cli_parse_args=True):
+    class Settings(BaseSettings, cli_parse_args=True, cli_use_class_docs_for_groups=True):
         field_a: CliSuppress[int] = 0
         field_b: str = Field(default=1, description=CLI_SUPPRESS)
-        hidden_obj: CliSuppress[HiddenSubModel]
+        hidden_obj: CliSuppress[HiddenSubModel] = Field(description='hidden_obj description')
         visible_obj: SubModel
```

PR #603, opened 2025-04-19
Support loading a specific nested key from YAML in YamlConfigSettingsSource

Seunghan-Jung:
# Description

It would be great if YamlConfigSettingsSource could support loading settings from a specific nested key within a YAML file.

## Use Case

Often, YAML files contain configuration for multiple environments or sections, such as:

```yaml
A_service:
  key: "A_key"
 
B_service:
  key: "B_key"
````

In this case, it would be useful to load settings from a specific key, for example, development, rather than requiring separate YAML files for each environment.

Suggested Feature

Allow an optional argument (e.g., nested_key) to be passed to YamlConfigSettingsSource like this:

```python
YamlConfigSettingsSource(settings_cls, yaml_file="config.yaml", yaml_nested_key="A_service")
```

This would tell the source to load settings from config["development"] instead of the root.

Benefits
	•	Reduces the need for multiple YAML files
	•	Simplifies environment-specific configuration management
	•	Aligns with how many existing tools and frameworks organize config

### DIFF
--- a/pydantic_settings/main.py
+++ b/pydantic_settings/main.py
@@ -62,6 +62,7 @@ class SettingsConfigDict(ConfigDict, total=False):
     json_file_encoding: str | None
     yaml_file: PathType | None
     yaml_file_encoding: str | None
+    yaml_nested_key: str | None
     pyproject_toml_depth: int
     """
     Number of levels **up** from the current working directory to attempt to find a pyproject.toml
@@ -446,6 +447,7 @@ class BaseSettings(BaseModel):
         json_file_encoding=None,
         yaml_file=None,
         yaml_file_encoding=None,
+        yaml_nested_key=None,
         toml_file=None,
         secrets_dir=None,
         protected_namespaces=('model_validate', 'model_dump', 'settings_customise_sources'),
--- a/pydantic_settings/sources/providers/yaml.py
+++ b/pydantic_settings/sources/providers/yaml.py
@@ -39,6 +39,7 @@ class YamlConfigSettingsSource(InitSettingsSource, ConfigFileSourceMixin):
         settings_cls: type[BaseSettings],
         yaml_file: PathType | None = DEFAULT_PATH,
         yaml_file_encoding: str | None = None,
+        yaml_nested_key: str | None = None,
     ):
         self.yaml_file_path = yaml_file if yaml_file != DEFAULT_PATH else settings_cls.model_config.get('yaml_file')
         self.yaml_file_encoding = (
@@ -46,7 +47,13 @@ class YamlConfigSettingsSource(InitSettingsSource, ConfigFileSourceMixin):
             if yaml_file_encoding is not None
             else settings_cls.model_config.get('yaml_file_encoding')
         )
+        self.yaml_nested_key = (
+            yaml_nested_key if yaml_nested_key is not None else settings_cls.model_config.get('yaml_nested_key')
+        )
         self.yaml_data = self._read_files(self.yaml_file_path)
+
+        if self.yaml_nested_key:
+            self.yaml_data = self.yaml_data[self.yaml_nested_key]
         super().__init__(settings_cls, self.yaml_data)
 
     def _read_file(self, file_path: Path) -> dict[str, Any]:
--- a/tests/test_source_yaml.py
+++ b/tests/test_source_yaml.py
@@ -166,3 +166,34 @@ def test_multiple_file_yaml(tmp_path):
 
     s = Settings()
     assert s.model_dump() == {'yaml3': 3, 'yaml4': 4}
+
+
+@pytest.mark.skipif(yaml is None, reason='pyYAML is not installed')
+def test_yaml_nested_key(tmp_path):
+    p = tmp_path / '.env'
+    p.write_text(
+        """
+    foobar: "Hello"
+    nested:
+        nested_field: "world!"
+    """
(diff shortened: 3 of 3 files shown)


pydantic_settings/main.py
```diff
@@ -62,6 +62,7 @@ class SettingsConfigDict(ConfigDict, total=False):
     json_file_encoding: str | None
     yaml_file: PathType | None
     yaml_file_encoding: str | None
+    yaml_nested_key: str | None
```
hramezani, line 65: Let's call it `yaml_config_section`.

Please add a comment for description as well.

pydantic_settings/sources/providers/yaml.py
```diff
@@ -39,14 +39,21 @@ def __init__(
...
             if yaml_file_encoding is not None
             else settings_cls.model_config.get('yaml_file_encoding')
         )
+        self.yaml_nested_key = (
+            yaml_nested_key if yaml_nested_key is not None else settings_cls.model_config.get('yaml_nested_key')
+        )
         self.yaml_data = self._read_files(self.yaml_file_path)
+
+        if self.yaml_nested_key:
+            self.yaml_data = self.yaml_data[self.yaml_nested_key]
```
hramezani, line 56: Please raise a `SettingsError` when the key does not exist.
Also, pelase add test for it

hramezani:
Thanks @Seunghan-Jung for the PR.

I left some comments. please update

pydantic_settings/sources/providers/yaml.py
Seunghan-Jung: The intention of this if statement was to use the entire YAML (as it was previously) when 'yaml_nested_key' is not set, so I didn’t raise an exception.

pydantic_settings/main.py
Seunghan-Jung: [a3f5389](https://github.com/pydantic/pydantic-settings/pull/603/commits/a3f5389683dadd702eab964ad6bdbcff3c073550)
I’ve changed the name and added description comments as you suggested.

pydantic_settings/sources/providers/yaml.py
Seunghan-Jung: The intention is to raise a KeyError when an invalid key is provided, clearly reminding the user of the mistake. 

I added it because I thought it would be better to deliver a clearer message than the default KeyError message. [c8bc837](https://github.com/pydantic/pydantic-settings/pull/603/commits/c8bc837ec6fbbf5f26b14679beabb8c405196980)

tests/test_source_yaml.py
```diff
@@ -166,3 +166,65 @@ def settings_customise_sources(
...
+            cls,
+            settings_cls: type[BaseSettings],
+            init_settings: PydanticBaseSettingsSource,
+            env_settings: PydanticBaseSettingsSource,
+            dotenv_settings: PydanticBaseSettingsSource,
+            file_secret_settings: PydanticBaseSettingsSource,
+        ) -> tuple[PydanticBaseSettingsSource, ...]:
+            return (YamlConfigSettingsSource(settings_cls),)
+
+    with pytest.raises(KeyError):
```
hramezani, line 229: pelase test the error message as well:
```suggestion
    with pytest.raises(KeyError, match='...'):
```

Seunghan-Jung: Oh, thank you. I added it in [f93364b](https://github.com/pydantic/pydantic-settings/pull/603/commits/f93364ba26df82ef6c4ed182697b5161e24f3132).

tests/test_source_yaml.py
```diff
@@ -166,3 +166,65 @@ def settings_customise_sources(
...
+            cls,
+            settings_cls: type[BaseSettings],
+            init_settings: PydanticBaseSettingsSource,
+            env_settings: PydanticBaseSettingsSource,
+            dotenv_settings: PydanticBaseSettingsSource,
+            file_secret_settings: PydanticBaseSettingsSource,
+        ) -> tuple[PydanticBaseSettingsSource, ...]:
+            return (YamlConfigSettingsSource(settings_cls),)
+
+    with pytest.raises(KeyError, match=r'yaml_config_section key ".*" not found in .+'):
```
hramezani, line 229:
```suggestion
    with pytest.raises(KeyError, match=f'yaml_config_section key "invalid_key" not found in {p}'):
```

tests/test_source_yaml.py
```diff
@@ -166,3 +166,65 @@ def settings_customise_sources(
...
+            cls,
+            settings_cls: type[BaseSettings],
+            init_settings: PydanticBaseSettingsSource,
+            env_settings: PydanticBaseSettingsSource,
+            dotenv_settings: PydanticBaseSettingsSource,
+            file_secret_settings: PydanticBaseSettingsSource,
+        ) -> tuple[PydanticBaseSettingsSource, ...]:
+            return (YamlConfigSettingsSource(settings_cls),)
+
+    with pytest.raises(KeyError, match=f'yaml_config_section key "invalid_key" not found in {p}'):
```
hramezani, line 229:
```suggestion
    with pytest.raises(KeyError, match='yaml_config_section key "invalid_key" not found in .+'):
```

hramezani:
Thanks @Seunghan-Jung

Seunghan-Jung:
Thanks @hramezani

PR #607, opened 2025-04-23
Azure Key Vault case insensitive support and dash-underscore translation

d15ky:
# Add case-insensitive and dash-to-underscore support for Azure Key Vault settings source

This PR improves the `AzureKeyVaultSettingsSource` making it more flexible and reliable when working with Azure Key Vault.

## Summary of changes

1. **Support for case-insensitive mode**

   `AzureKeyVaultSettingsSource` now fully supports the `case_sensitive` option. While implementing this, an issue was found in the inherited `EnvironmentSettingsSource`, which did not correctly process nested models with validation aliases in case-insensitive mode. The related test has been updated to include such a case and now passes with the changes.

2. **New `dash_to_underscore` option (enabled by default)**

   A `dash_to_underscore` option has been added to translate between dashes (`-`) in Azure Key Vault secret names and underscores (`_`) in model field names. This mapping is enabled by default and is safe, as:
   - Python identifiers cannot contain dashes.
   - Azure Key Vault secret names cannot contain underscores.

   The option can be disabled for cases where this behavior is not desired though I cannot think of any. 

All tests pass, linting is clean, and the behavior is covered by updated test cases. Code style and structure were followed closely to match existing patterns in both the implementation and tests. 

## Examples

**Dash-to-underscore translation:**

```python
from pydantic_settings.sources import AzureKeyVaultSettingsSource

class Settings(BaseSettings):
    my_field: str

settings = Settings(_secrets_sources=[
    AzureKeyVaultSettingsSource(vault_url="https://my-vault.vault.azure.net/")
])
# Retrieves the 'my-field' secret and maps it to 'my_field'
```

**Nested model with alias, case-insensitive:**

```python

class Inner(BaseModel):
    some_value: str = Field(..., alias="MyKey")

class Outer(BaseSettings):
    inner: Inner

# With case_sensitive=False, environment variable 'MYKEY' will correctly map to 'some_value' field
```

### DIFF
--- a/pydantic_settings/sources/base.py
+++ b/pydantic_settings/sources/base.py
@@ -412,21 +412,24 @@ class PydanticBaseEnvSettingsSource(PydanticBaseSettingsSource):
             if not annotation or not hasattr(annotation, 'model_fields'):
                 values[name] = value
                 continue
+            else:
+                model_fields: dict[str, FieldInfo] = annotation.model_fields
 
             # Find field in sub model by looking in fields case insensitively
-            for sub_model_field_name, f in annotation.model_fields.items():
-                if not f.validation_alias and sub_model_field_name.lower() == name.lower():
-                    sub_model_field = f
+            for sub_model_field_name, sub_model_field in model_fields.items():
+                aliases, _ = _get_alias_names(sub_model_field_name, sub_model_field)
+                _search = (alias for alias in aliases if alias.lower() == name.lower())
+                if field_key := next(_search, None):
                     break
 
-            if not sub_model_field:
+            if not field_key:
                 values[name] = value
                 continue
 
             if _lenient_issubclass(sub_model_field.annotation, BaseModel) and isinstance(value, dict):
-                values[sub_model_field_name] = self._replace_field_names_case_insensitively(sub_model_field, value)
+                values[field_key] = self._replace_field_names_case_insensitively(sub_model_field, value)
             else:
-                values[sub_model_field_name] = value
+                values[field_key] = value
 
         return values
 
--- a/pydantic_settings/sources/providers/azure.py
+++ b/pydantic_settings/sources/providers/azure.py
@@ -5,6 +5,8 @@ from __future__ import annotations as _annotations
 from collections.abc import Iterator, Mapping
 from typing import TYPE_CHECKING, Optional
 
+from pydantic.fields import FieldInfo
+
 from .env import EnvSettingsSource
 
 if TYPE_CHECKING:
@@ -42,27 +44,33 @@ class AzureKeyVaultMapping(Mapping[str, Optional[str]]):
     def __init__(
         self,
         secret_client: SecretClient,
+        case_sensitive: bool,
     ) -> None:
         self._loaded_secrets = {}
         self._secret_client = secret_client
-        self._secret_names: list[str] = [
+        self._case_sensitive = case_sensitive
+        self._secret_map: dict[str, str] = self._load_remote()
+
+    def _load_remote(self) -> dict[str, str]:
+        secret_names: Iterator[str] = (
             secret.name for secret in self._secret_client.list_properties_of_secrets() if secret.name and secret.enabled
-        ]
+        )
+        if self._case_sensitive:
+            return {name: name for name in secret_names}
+        return {name.lower(): name for name in secret_names}
 
     def __getitem__(self, key: str) -> str | None:
-        if key not in self._loaded_secrets and key in self._secret_names:
-            try:
-                self._loaded_secrets[key] = self._secret_client.get_secret(key).value
-            except Exception:
-                raise KeyError(key)
-
+        if not self._case_sensitive:
+            key = key.lower()
+        if key not in self._loaded_secrets and key in self._secret_map:
+            self._loaded_secrets[key] = self._secret_client.get_secret(self._secret_map[key]).value
         return self._loaded_secrets[key]
 
     def __len__(self) -> int:
-        return len(self._secret_names)
+        return len(self._secret_map)
 
     def __iter__(self) -> Iterator[str]:
-        return iter(self._secret_names)
+        return iter(self._secret_map.keys())
 
 
 class AzureKeyVaultSettingsSource(EnvSettingsSource):
@@ -74,6 +82,8 @@ class AzureKeyVaultSettingsSource(EnvSettingsSource):
         settings_cls: type[BaseSettings],
         url: str,
         credential: TokenCredential,
+        dash_to_underscore: bool = True,
+        case_sensitive: bool | None = None,
         env_prefix: str | None = None,
         env_parse_none_str: str | None = None,
         env_parse_enums: bool | None = None,
@@ -81,9 +91,10 @@ class AzureKeyVaultSettingsSource(EnvSettingsSource):
         import_azure_key_vault()
         self._url = url
         self._credential = credential
+        self._dash_to_underscore = dash_to_underscore
         super().__init__(
             settings_cls,
-            case_sensitive=True,
+            case_sensitive=case_sensitive,
             env_prefix=env_prefix,
             env_nested_delimiter='--',
             env_ignore_empty=False,
@@ -93,7 +104,12 @@ class AzureKeyVaultSettingsSource(EnvSettingsSource):
 
     def _load_env_vars(self) -> Mapping[str, Optional[str]]:
         secret_client = SecretClient(vault_url=self._url, credential=self._credential)
-        return AzureKeyVaultMapping(secret_client)
+        return AzureKeyVaultMapping(secret_client, self.case_sensitive)
+
+    def _extract_field_info(self, field: FieldInfo, field_name: str) -> list[tuple[str, str, bool]]:
+        if self._dash_to_underscore:
+            return list((x[0], x[1].replace('_', '-'), x[2]) for x in super()._extract_field_info(field, field_name))
+        return super()._extract_field_info(field, field_name)
 
     def __repr__(self) -> str:
         return f'{self.__class__.__name__}(url={self._url!r}, env_nested_delimiter={self.env_nested_delimiter!r})'
--- a/tests/test_settings.py
+++ b/tests/test_settings.py
@@ -2738,18 +2738,18 @@ def test_nested_models_leaf_vs_deeper_env_dict_assumed(env):
 
 def test_case_insensitive_nested_optional(env):
     class NestedSettings(BaseModel):
-        FOO: str
+        FOO: str = Field(..., alias='Foo')
         BaR: int
 
     class Settings(BaseSettings):
         model_config = SettingsConfigDict(env_nested_delimiter='__', case_sensitive=False)
 
-        nested: Optional[NestedSettings]
+        nEstEd: Optional[NestedSettings] = Field(..., alias='NesTed')
 
     env.set('nested__FoO', 'string')
     env.set('nested__bar', '123')
     s = Settings()
-    assert s.model_dump() == {'nested': {'BaR': 123, 'FOO': 'string'}}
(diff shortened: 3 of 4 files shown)


hramezani:
Thanks @d15ky for the PR.

> While implementing this, an issue was found in the inherited EnvironmentSettingsSource, which did not correctly process nested models with validation aliases in case-insensitive mode. The related test has been updated to include such a case and now passes with the changes.

Is it possible to take these changes out and make another PR? Also, please don't update the existing test and add a new tests.

d15ky:
@hramezani thanks for the quick review! Sure, I'll move the bugfix to a separate PR.

Regarding the test: I updated it because its name, `test_case_insensitive_nested_optional`, seemed consistent with the changes. That said, I'll revert the modification and add a new test in the separate PR instead.

d15ky:
@hramezani I've separated the bugfix into a dedicated PR.

I've removed the test changes from this one, but kept the bugfix logic itself, as it's required for the AKV tests to pass with the new case-insensitive functionality.

Let me know if anything else needs adjusting!

hramezani:
Thanks @d15ky for updating the PR.

@AndreuCodina could you review the PR?

AndreuCodina:
It's a complex topic.

Dash to underscore was in my initial Key Vault source, but I removed it. In the official provider of Key Vault, 2 dashes are used for nesting (https://learn.microsoft.com/en-us/aspnet/core/security/key-vault-configuration?view=aspnetcore-9.0#secret-storage-in-the-production-environment-with-azure-key-vault) or arrays (https://learn.microsoft.com/en-us/aspnet/core/security/key-vault-configuration?view=aspnetcore-9.0#bind-an-array-to-a-class), and 1 dash can be used for prefixes (https://learn.microsoft.com/en-us/aspnet/core/security/key-vault-configuration?view=aspnetcore-9.0#use-a-key-name-prefix), and they are stripped.

As we use .NET and Python, we use PascalCase secrets in Key Vault, and we are able to read `MySecret` from .NET and Python. With your secret convention, you can only use 1 provider for a specific language (Python) instead of using more programming languages, and that's not a reasonable enterprise solution.

These are our settings:

```python
class ApplicationSettings(BaseSettings):
    model_config = SettingsConfigDict(alias_generator=to_pascal)
    
    sql_connection_string: str
```

I guess you don't use PascalCase outside the Key Vault source, so you can't apply `alias_generator` to all fields, and instead you'd have to use `alias` for the secrets from Azure Key Vault.

I also don't like boilerplate, but you have to understand your solution is valid, but "-" in Key Vault is usually a special character. For example, Spring Boot works with ".", and with the Key Vault provider, "-" is replaced with "." (https://learn.microsoft.com/en-us/azure/developer/java/spring-framework/secret-management#use---instead-of--in-secret-names). So, in that ecosystem, "-" doesn't mean "different word".

TBH I'd prefer a translator from PascalCase to snake_case instead of replacing a special character.

d15ky:
@AndreuCodina Thank you for the detailed response. I'm not sure if I’ve fully understood all the concerns, so I’ll echo my understanding and provide counterpoints to clarify my intent.

1) **Dashes for nesting, arrays, and prefixes**

The proposed solution doesn’t interfere with that behavior. For example, a double dash in my implementation would still correctly map to a nested model. The transformation is **exclusively** applied when Pydantic searches for a field to map to — it doesn’t affect aliases, nesting logic, or any other part of the resolution process. 

2) **PascalCase and language interoperability**

I’m not proposing to replace PascalCase conventions — this is an **opt-in** translation that only affects Python variable names (not aliases) and only during the final map search when other approaches failed. In your example setup, using PascalCase secrets with `alias_generator=to_pascal` would continue to work as intended.

I agree that there’s a risk of the Python side of the team assuming a specific dash convention, but since:
- dashes appear to be open to interpretation (e.g., Spring Boot translating them to dots),
- and since `pydantic-settings` doesn't support writing to Key Vault (only reading),
  
it seems like the real risk here lies more in **a lack of team-wide naming conventions** than in pydantic’s behavior. This translation doesn’t enforce or imply a naming standard on the Key Vault itself — a badly named KeyVault variable that would break the flow of .NET side still needs to be created manually. 

3) **Spring Boot example**

To me, it actually supports the idea that **dashes are flexible and interpreted differently** depending on the ecosystem — which makes it reasonable for Python tooling to offer a translation strategy tailored to Python conventions. I don't see any other way dashes can be interpreted by Python in this context except as delimiters. Once again, nesting logic is unaffected by this translation and would take precedence in resolution. 

---

So if I understand correctly, the core question is not about technical correctness or risk of breakage — it's about whether this kind of client-side, read-only convention should be supported by pydantic-settings to accommodate Pythonic naming patterns. The transformation itself is safe, isolated, and opt-in, and its inclusion would simply acknowledge a common use case without enforcing it. 

Since pydantic-settings is read-only and requires manual creation of secrets anyway, I think it’s reasonable to offer this behavior as a tool — leaving it up to teams to apply it (or not) based on their context. 

With that said, I’m open to remove the feature entirely if needed, especially since most of this PR is focused on case-insensitivity. Alternatively, I could split it into a separate PR to continue the discussion separately.

AndreuCodina:
> @AndreuCodina Thank you for the detailed response. I'm not sure if I’ve fully understood all the concerns, so I’ll echo my understanding and provide counterpoints to clarify my intent.
> 
> 1. **Dashes for nesting, arrays, and prefixes**
> 
> The proposed solution doesn’t interfere with that behavior. For example, a double dash in my implementation would still correctly map to a nested model. The transformation is **exclusively** applied when Pydantic searches for a field to map to — it doesn’t affect aliases, nesting logic, or any other part of the resolution process.
> 
> 2. **PascalCase and language interoperability**
> 
> I’m not proposing to replace PascalCase conventions — this is an **opt-in** translation that only affects Python variable names (not aliases) and only during the final map search when other approaches failed. In your example setup, using PascalCase secrets with `alias_generator=to_pascal` would continue to work as intended.
> 
> I agree that there’s a risk of the Python side of the team assuming a specific dash convention, but since:
> 
> * dashes appear to be open to interpretation (e.g., Spring Boot translating them to dots),
> * and since `pydantic-settings` doesn't support writing to Key Vault (only reading),
> 
> it seems like the real risk here lies more in **a lack of team-wide naming conventions** than in pydantic’s behavior. This translation doesn’t enforce or imply a naming standard on the Key Vault itself — a badly named KeyVault variable that would break the flow of .NET side still needs to be created manually.
> 
> 3. **Spring Boot example**
> 
> To me, it actually supports the idea that **dashes are flexible and interpreted differently** depending on the ecosystem — which makes it reasonable for Python tooling to offer a translation strategy tailored to Python conventions. I don't see any other way dashes can be interpreted by Python in this context except as delimiters. Once again, nesting logic is unaffected by this translation and would take precedence in resolution.
> 
> So if I understand correctly, the core question is not about technical correctness or risk of breakage — it's about whether this kind of client-side, read-only convention should be supported by pydantic-settings to accommodate Pythonic naming patterns. The transformation itself is safe, isolated, and opt-in, and its inclusion would simply acknowledge a common use case without enforcing it.
> 
> Since pydantic-settings is read-only and requires manual creation of secrets anyway, I think it’s reasonable to offer this behavior as a tool — leaving it up to teams to apply it (or not) based on their context.
> 
> With that said, I’m open to remove the feature entirely if needed, especially since most of this PR is focused on case-insensitivity. Alternatively, I could split it into a separate PR to continue the discussion separately.

Your solution works, but:

- AFAIK, a single dash is not used in any current provider to represent the separation between two words (e.g. "MySecret" is "My Secret"). Instead, a dash represents a special case in any provider, even in Spring Boot, so this provider would be the first to say "you can split words with dashes".
- With case insensitivity, a lot of languages can read from PascalCase, camelCase or lowercase, so I don't think naming the secrets with one dash is a good idea, and the PR tries to make a user-friendly solution the rare case of using a dash to represent the separation between words.

Meanwhile your solution prevents the use of `alias_generator` of `alias` and offers a cleaner solution, it forces a naming convention in Key Vault that isn't compatible or shared by other programming languages, so my personal opinion would be to stop using that convention and introduce a feature in pydantic-settings for, when a capital letter is detected, change it to a lowercase letter + underscore if it's not the first letter. For example: `SqlConnectionString|sqlConnectionString` -> `sql_connection_string`.

FYI in our case, we use PascalCase for Key Vault, .env files and App Configuration.  And, as we use `to_pascal`, the environment variables map seamlessly. It's easy and reduces development time.

But still, if you want to have this feature, I accept it.

hramezani:
Thanks both for your contributions here.

I am not an expert in KeyVault. But let's disable the new behavior by default, and the user enables it.
Also, we need to document it to let the user know the behavior exists and how it can be enabled.

@AndreuCodina are you happy with my proposed approach?

AndreuCodina:
> Thanks both for your contributions here.
> 
> I am not an expert in KeyVault. But let's disable the new behavior by default, and the user enables it. Also, we need to document it to let the user know the behavior exists and how it can be enabled.
> 
> @AndreuCodina are you happy with my proposed approach?

Yes, but I don't know exactly what are the real implications of removing the try...except with the `KeyError`.

AndreuCodina:
One thing to add @d15ky, Key Vault should always be case-insensitive:

> Objects are uniquely identified within Key Vault using a case-insensitive identifier called the object identifier. [...] Key Vault object identifiers are also valid URLs, but should always be compared as case-insensitive strings.

https://learn.microsoft.com/en-us/azure/key-vault/general/about-keys-secrets-certificates#object-identifiers

d15ky:
@AndreuCodina I removed it because it was masking the real exception thrown by the Azure library, making it look like a KeyError. I believe this could mislead developers, so it's better to let the original exception bubble up.

d15ky:
> One thing to add @d15ky, Key Vault should always be case-insensitive:

Well, I’ve only enabled case-insensitivity, not the other way around. It was always _case-sensitive_ after the last update (2.9.0+) — that’s what motivated me to make this PR in the first place. I’m fine if you suggest making it always case-insensitive.

AndreuCodina:
> > One thing to add @d15ky, Key Vault should always be case-insensitive:
> 
> Well, I’ve only enabled case-insensitivity, not the other way around. It was always _case-sensitive_ after the last update (2.9.0+) — that’s what motivated me to make this PR in the first place. I’m fine if you suggest making it always case-insensitive.

Normally you only find information regarding the .NET provider, but this documentation is common, so I think it's the approach.

hramezani:
Let's have it not as default behavior.

@d15ky please add the documentation for it

d15ky:
@hramezani I've added the documentation and updated the test.

While browsing the Azure Key Vault documentation for references, I noted that kebab-case is actually used in secret names ([Azure SDK for Python documentation](https://learn.microsoft.com/en-us/python/api/overview/azure/keyvault-secrets-readme?view=azure-python#set-a-secret)), and I haven't found any direct guidance discouraging its use or recommending PascalCase. So I decided not to add a warning against implementing it.

docs/index.md
```diff
@@ -1889,6 +1889,51 @@ class AzureKeyVaultSettings(BaseSettings):
...
+    @classmethod
+    def settings_customise_sources(
+        cls,
+        settings_cls: type[BaseSettings],
+        init_settings: PydanticBaseSettingsSource,
+        env_settings: PydanticBaseSettingsSource,
+        dotenv_settings: PydanticBaseSettingsSource,
+        file_secret_settings: PydanticBaseSettingsSource,
+    ) -> tuple[PydanticBaseSettingsSource, ...]:
+        az_key_vault_settings = AzureKeyVaultSettingsSource(
```
hramezani, line 1924: I think you forgot to pass `dash_to_underscore=True` here

d15ky: thanks for catching it! idk how I managed to forget the main point of the example... it's fixed

hramezani (approved):

hramezani:
Thanks @d15ky

PR #608, opened 2025-04-23
Fix case-insensitive handling of nested aliases in EnvironmentSettingsSource

d15ky:
# Fix case-insensitive handling of nested aliases in EnvironmentSettingsSource

This PR fixes a bug in `EnvironmentSettingsSource` where nested models with validation aliases were not correctly populated from environment variables when `case_sensitive=False`.

Previously, case-insensitive matching did not take aliases into account in nested contexts, resulting in missing values. This fix ensures correct resolution in such cases.

All tests pass, linting is clean, and a dedicated test has been added to validate the fix without modifying existing test coverage. Code style and structure were followed closely to match the existing patterns in both implementation and testing.

## Example

```python

class Inner(BaseModel):
    some_value: str = Field(..., alias="MyKey")

class Outer(BaseSettings):
    inner: Inner

# With case_sensitive=False, 'MYKEY' will correctly map to 'some_value'
```

### DIFF
--- a/pydantic_settings/sources/base.py
+++ b/pydantic_settings/sources/base.py
@@ -412,21 +412,24 @@ class PydanticBaseEnvSettingsSource(PydanticBaseSettingsSource):
             if not annotation or not hasattr(annotation, 'model_fields'):
                 values[name] = value
                 continue
+            else:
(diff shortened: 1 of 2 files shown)


hramezani:
Thanks @d15ky

PR #609, opened 2025-04-24
Fix for case-insensitive alias matching in `InitSettingsSource`

Shindora:
# Issue
When using case-insensitive settings with aliases, the InitSettingsSource wasn't correctly matching field aliases in a case-insensitive manner due to not considering case sensitivity when obtaining alias names.

# Changes
Modified the `__init__` method of `InitSettingsSource` to pass the **case_sensitive** parameter to get_alias_names() when generating match aliases
Now properly creates two sets of aliases:
- Canonical aliases (always case-sensitive) for the output key
- Match aliases that respect the configured case sensitivity for matching input values

Add unittest `test_init_settings_source_extra_fields_case_sensitive`

### DIFF
--- a/pydantic_settings/sources/base.py
+++ b/pydantic_settings/sources/base.py
@@ -306,11 +306,7 @@ class InitSettingsSource(PydanticBaseSettingsSource):
         )
 
     def __repr__(self) -> str:
-        return (
-            f'{self.__class__.__name__}(init_kwargs={self.init_kwargs!r}, '
-            f'case_sensitive={self.case_sensitive!r}, '
-            f'nested_model_default_partial_update={self.nested_model_default_partial_update!r})'
-        )
+        return f'{self.__class__.__name__}(init_kwargs={self.init_kwargs!r})'
 
 
 class PydanticBaseEnvSettingsSource(PydanticBaseSettingsSource):
--- a/pydantic_settings/sources/base.py
+++ b/pydantic_settings/sources/base.py
@@ -258,19 +258,37 @@ class InitSettingsSource(PydanticBaseSettingsSource):
         settings_cls: type[BaseSettings],
         init_kwargs: dict[str, Any],
         nested_model_default_partial_update: bool | None = None,
+        case_sensitive: bool | None = None,
     ):
-        self.init_kwargs = {}
-        init_kwarg_names = set(init_kwargs.keys())
-        for field_name, field_info in settings_cls.model_fields.items():
-            alias_names, *_ = _get_alias_names(field_name, field_info)
-            init_kwarg_name = init_kwarg_names & set(alias_names)
-            if init_kwarg_name:
-                preferred_alias = alias_names[0]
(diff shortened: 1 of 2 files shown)


hramezani:
Thanks @Shindora for the PR.

Could you please add some examples that don't work with the existing code? Also, you need to add tests for these changes.

@kschwab Do you have time to take a look?

hramezani:
@Shindora, you've added two tests to the PR, but both pass without your change in `base.py`

Shindora:
> @Shindora, you've added two tests to the PR, but both pass without your change in `base.py`

I just add the unittest for both cases: sensitive and insensitive

hramezani:
@Shindora, the 'test_case_sensitive ' test passes without your change in `base.py`.

hramezani:
This is the only failing case
```py
    class CaseInsensitiveSettings(BaseSettings):
        foo: str = Field(..., alias='FOO')
        model_config = SettingsConfigDict(case_sensitive=False, extra='allow')

    # Test case-insensitive with extra field
    monkeypatch.setattr(os, 'environ', value={})
    init_kwargs = {'Foo': 'foo_value', 'extra_field': 'extra_value'}
    settings = CaseInsensitiveSettings(**init_kwargs)
    assert settings.foo == 'foo_value'
    assert settings.__pydantic_extra__ == {'extra_field': 'extra_value'}
```
Please revert your change in `test_case_sensitive`

cc @kschwab

Shindora:
> This is the only failing case
> 
> ```python
>     class CaseInsensitiveSettings(BaseSettings):
>         foo: str = Field(..., alias='FOO')
>         model_config = SettingsConfigDict(case_sensitive=False, extra='allow')
> 
>     # Test case-insensitive with extra field
>     monkeypatch.setattr(os, 'environ', value={})
>     init_kwargs = {'Foo': 'foo_value', 'extra_field': 'extra_value'}
>     settings = CaseInsensitiveSettings(**init_kwargs)
>     assert settings.foo == 'foo_value'
>     assert settings.__pydantic_extra__ == {'extra_field': 'extra_value'}
> ```
> 
> Please revert your change in `test_case_sensitive`
> 
> cc @kschwab

reverted ! I add the new test case to cover my adjustment `test_init_settings_source_extra_fields_case_sensitive`. Please take a look.

hramezani:
Thanks @Shindora for the update.

Actually, `InitSettingsSource` is similar to passing data directly to pydantic. So, I am not sure we want to support `case_sensitive` in `InitSettingsSource`. I'm afraid it breaks people's code. Especially, the amount of code change to support this case seems too much to me.

What do you think @kschwab?

kschwab:
@hramezani I would agree with you. I have similar concerns. i.e., case sensitive seems more applicable to environment or CLI setting like sources.

hramezani:
Thanks @kschwab for your feedback here.

I closed the issue as we decided not to support it.

Thanks @Shindora for your contribution.

Issue #623, opened 2025-05-26
Top-level CLI alias for nested model


maxnoe:
We are looking to replace `traitlets.config` with pydantic_settings and one feature we are missing that is important to us is the ability to define short, top-level command line options for things that are defined deeper in the hierarchy.

E.g. consider:

```python
from pydantic_settings import BaseSettings, CliApp


class SubSettings(BaseSettings):
    option: str = "foo"


class Settings(BaseSettings):
    sub: SubSettings

    def cli_cmd(self) -> None:
        print(self.model_dump())


CliApp.run(Settings)
```

In this example, I'd like to expose `--sub.option` just as `--option`, e.g. by defining an alias mapping.

maxnoe:
Looking at the code, maybe adding a field `cli_aliases` to the `SettingsConfigDict` would be a way here?

So that the code would look like this:

```python
from pydantic_settings import BaseSettings, CliApp, SettingsConfigDict


class SubSettings(BaseSettings):
    option: str = "foo"

class Settings(BaseSettings):
    sub: SubSettings = SubSettings()

    model_config = SettingsConfigDict(
        cli_aliases={
            "alias": "sub.option",
        },
    )

    def cli_cmd(self) -> None:
        print(self.model_dump())


CliApp.run(Settings)
```

And then `python example.py --alias="bar"` would set `sub.option="bar"`

karta9821:
It should rather be:

```python
model_config = SettingsConfigDict(
    cli_aliases={
        "sub.option": "alias",
    },
)
```

What is the expected behavior in the following case?

```python
from pydantic_settings import BaseSettings, CliApp, SettingsConfigDict
from pydantic import Field

class SubSettings(BaseSettings):
    option: str = "foo"

class Settings(BaseSettings):
    sub: SubSettings = SubSettings()
    option: str = "bar"
    option2: str = Field("foobar", alias="option")

    model_config = SettingsConfigDict(
        cli_aliases={
            "sub.option": "option",
        },
    )
```

hramezani:
Fixed in https://github.com/pydantic/pydantic-settings/commit/180e74e324d8418e22b5dcb196fd6390fa964537

PR #624, opened 2025-05-28
feat: Add `cli_aliases` to CLI settings

karta9821:
It addresses to https://github.com/pydantic/pydantic-settings/issues/623


karta9821:
@maxnoe Please verify if it is okay

hramezani:
@kschwab do you have time to take a look here?

maxnoe:
Thanks for the quick implementation, I played around with the branch here and it works as I would expect!

Can't comment too much about the code itself.

tests/test_source_cli.py
```diff
@@ -2525,3 +2525,38 @@ def settings_customise_sources(
...
+        option: str = Field(default='foo')
+
+    class Settings(BaseSettings):
+        nested: Nested = Nested()
+        option2: str = Field(default='foo2')
+
+        model_config = SettingsConfigDict(cli_aliases={'option2': 'nested.option'})
+
+    with pytest.raises(SettingsError, match='Settings.option2 has multiple aliases: option2 and nested.option'):
+        CliApp.run(Settings, cli_args=['--option2', 'bar'])
```
maxnoe, line 2562: I tested this locally and it works, but maybe worth adding a test for a twice-nested option?

```
class FooSettings(BaseSettings):
    foo_val: str = "foo"

class BarSettings(BaseSettings):
    bar_val: str = "bar"
    foo: FooSettings = FooSettings()


class Settings(BaseSettings):
        cli_aliases={
            "foo-val": "bar.foo.foo_val",
            "bar-val": "bar.bar_val",
        },
    )
    bar: BarSettings = BarSettings()
```

karta9821:
please review

hramezani:
I am waiting for @kschwab for a review, as he implemented most of the CLI settings source code.

kschwab:
Thanks @karta9821 and @maxnoe for the PR. I am back from vacation. The only two items I see are:

- Documentation needs to be added.
- I would simplify and support multiple aliases, similar to how pydantic AliasChoices works. ie, `Map[str, str | list[str]]`, where key should be the target field, and values are the aliases.

The only real concern I have is confusion between `cli_alias` and pydantic `alias`. Is it possible to give the config option a different name? @hramezani do you have thoughts here?

Other than those items, everything else looks good.

hramezani:
Thanks @kschwab for checking the PR

> The only real concern I have is confusion between cli_alias and pydantic alias. Is it possible to give the config option a different name? @hramezani do you have thoughts here?

Agree, Do you have a suggestion for the config name? 

@karta9821 Please address the points mentioned by @kschwab. Then we can merge the PR.

karta9821:
> Thanks @kschwab for checking the PR
> 
> > The only real concern I have is confusion between cli_alias and pydantic alias. Is it possible to give the config option a different name? @hramezani do you have thoughts here?
> 
> Agree, Do you have a suggestion for the config name?
> 
> @karta9821 Please address the points mentioned by @kschwab. Then we can merge the PR.

Sure, I can adjust the code :)

How about `cli_shortcuts`

hramezani:
> > Thanks @kschwab for checking the PR
> > > The only real concern I have is confusion between cli_alias and pydantic alias. Is it possible to give the config option a different name? @hramezani do you have thoughts here?
> > 
> > 
> > Agree, Do you have a suggestion for the config name?
> > @karta9821 Please address the points mentioned by @kschwab. Then we can merge the PR.
> 
> Sure, I can adjust the code :)
> 
> How about `cli_shortcuts`

Looks good to me. what do you think @kschwab

kschwab:
@hramezani @karta9821 yes, `cli_shortcuts` sounds good to me as well.

karta9821:
Fixed

hramezani:
@karta9821 Thanks for the update. please add the documentation for it

karta9821:
Added

hramezani:
Thanks @karta9821 for the quick update.

@kschwab please take a final look when you have time.

kschwab:
Hey @karta9821 thanks for the changes! It's a great addition, looks good to me.

hramezani:
Thanks all

PR #627, opened 2025-06-11
Add support for systemd's LoadCredential mechanism

Tom-Hubrecht:
C.f. https://systemd.io/CREDENTIALS/

This allows using systemd's native secrets management mechanism by default


hramezani:
Thanks @Tom-Hubrecht for creating this PR. Unfortunately, we can't accept these new settings source because of the following reasons:

- People are not interested in this settings source, and this is the first request.
- pydantic-settings provide the ability to have this settings source as a custom source in your code
- It is really hard for us to maintain these settings sources, and we are already maintaining a lot of them.

Thanks for your understanding.

PR #637, opened 2025-06-22
Adapt test_protected_namespace_defaults for dev. Pydantic

musicinmybrain:
Based on testing with a current development snapshot of Pydantic and Python 3.14.

(In Fedora Rawhide, Python 3.14 is now the system Python, so we have a bleeding-edge snapshot of Pydantic with incomplete Python 3.14 support, because this is better than not having an installable Pydantic at all yet. I have tested this PR downstream in that environment.)


hramezani:
Thanks @musicinmybrain for this PR.

`pydantic-settings` doesn't officially support Python 3.14, but we have to add the support soon.
We need to add Python 3.14 support in `pydantic-core` and `pydantic` first.

Let's keep the PR open and merge it when we want to support Python 3.14.

hramezani (approved):

hramezani:
Thanks @musicinmybrain

PR #643, opened 2025-06-26
CLI Serialize Support

### LINKED ISSUE #641: Can I serialize args for calling a subprocess?
I.e. I have a program that I use BaseSettings for, and I'd like to invoke it in a "type safe" way; is that something that is possible right now?

For example:
```python
class MyProgram(BaseSettings):
  hi: str

args: MyProgram
subprocess.run("my_program.py", args.serialize()) # I'd like to do something like this
```

kschwab:
Fixes #641


hramezani (approved):

hramezani:
Great work! 

Thanks @kschwab

PR #644, opened 2025-06-26
Inspect type aliases to determine if an annotation is complex

### LINKED ISSUE #536: Pydantic Settings cannot deserialize JSON value of type declared with PEP-695 style type alias
With the following Pydantic settings model

```python
from typing import Annotated, Literal

class Bar(BaseModel):
    type: Literal["bar"] = "bar"
    bar: int
    
class Baz(BaseModel):
    type: Literal["baz"] = "baz"
    baz: str
    
type BarOrBaz = Annotated[Bar | Baz, Field(discriminator="type")]

class Foo(BaseSettings):
    foo: BarOrBaz
```

and the env

```
export FOO='{"type": "bar", "bar": 42}'
```

Pydantic Settings fails to deserialize FOO:

```python
Foo()
```

```
ValidationError: 1 validation error for Foo
foo
  Input should be a valid dictionary or object to extract fields from [type=model_attributes_type, input_value='{"type": "bar", "bar": 42}', input_type=str]
    For further information visit https://errors.pydantic.dev/2.10/v/model_attributes_type
```

Declaring BarOrBaz without the PEP-695 style type alias (i.e. leaving out the `type` keyword) works as expected - validation succeeds.

**Version Info**

```
             pydantic version: 2.10.6
        pydantic-core version: 2.27.2
          pydantic-core build: profile=release pgo=false
                 install path: /path/to/my/venv/lib/python3.13/site-packages/pydantic
               python version: 3.13.1 (main, Dec  3 2024, 17:59:52) [Clang 19.1.5 ]
                     platform: macOS-15.3-arm64-arm-64bit-Mach-O
             related packages: mypy-1.15.0 fastapi-0.115.8 pydantic-settings-2.7.1 typing_extensions-4.12.2
                       commit: unknown
```

tselepakis:
This PR ensures that type aliases declared using type (PEP 695) are correctly resolved and handled during environment value parsing and model field introspection.

I aimed to keep the changes minimal, but addressing compatibility issues required introducing a couple of new patterns, specifically to work around syntax errors triggered by pytest when using Python versions <3.12, as well as ruff linter errors. If there are better or more idiomatic approaches, I'm happy to discuss them. I'm not certain the current solutions are the best long-term fit for the project.

Fixes #536

### DIFF
--- a/pydantic_settings/sources/base.py
+++ b/pydantic_settings/sources/base.py
@@ -356,10 +356,8 @@ class PydanticBaseEnvSettingsSource(PydanticBaseSettingsSource):
 
         if not v_alias or self.config.get('populate_by_name', False):
             annotation = field.annotation
-            if annotation and (
-                typing_objects.is_typealiastype(annotation) or typing_objects.is_typealiastype(get_origin(annotation))
-            ):
-                annotation = _strip_annotated(annotation.__value__)
+            if typing_objects.is_typealiastype(annotation) or typing_objects.is_typealiastype(get_origin(annotation)):
+                annotation = _strip_annotated(annotation.__value__)  # type: ignore[union-attr]
             if is_union_origin(get_origin(annotation)) and _union_is_complex(annotation, field.metadata):
                 field_info.append((field_name, self._apply_case_sensitive(self.env_prefix + field_name), True))
             else:
--- a/pydantic_settings/sources/base.py
+++ b/pydantic_settings/sources/base.py
@@ -16,6 +16,7 @@ from pydantic._internal._typing_extra import (  # type: ignore[attr-defined]
 from pydantic._internal._utils import is_model_class
 from pydantic.fields import FieldInfo
 from typing_extensions import get_args
+from typing_inspection import typing_objects
 from typing_inspection.introspection import is_union_origin
 
 from ..exceptions import SettingsError
@@ -25,6 +26,7 @@ from .utils import (
     _annotation_is_complex,
     _get_alias_names,
     _get_model_fields,
+    _strip_annotated,
     _union_is_complex,
 )
 
@@ -353,7 +355,10 @@ class PydanticBaseEnvSettingsSource(PydanticBaseSettingsSource):
                 field_info.append((v_alias, self._apply_case_sensitive(v_alias), False))
 
         if not v_alias or self.config.get('populate_by_name', False):
-            if is_union_origin(get_origin(field.annotation)) and _union_is_complex(field.annotation, field.metadata):
+            annotation = field.annotation
+            if typing_objects.is_typealiastype(annotation):
+                annotation = _strip_annotated(annotation.__value__)
+            if is_union_origin(get_origin(annotation)) and _union_is_complex(annotation, field.metadata):
                 field_info.append((field_name, self._apply_case_sensitive(self.env_prefix + field_name), True))
             else:
                 field_info.append((field_name, self._apply_case_sensitive(self.env_prefix + field_name), False))
--- a/pydantic_settings/sources/base.py
+++ b/pydantic_settings/sources/base.py
@@ -356,7 +356,9 @@ class PydanticBaseEnvSettingsSource(PydanticBaseSettingsSource):
 
         if not v_alias or self.config.get('populate_by_name', False):
             annotation = field.annotation
-            if typing_objects.is_typealiastype(annotation):
+            if annotation and (
+                typing_objects.is_typealiastype(annotation) or typing_objects.is_typealiastype(get_origin(annotation))
+            ):
                 annotation = _strip_annotated(annotation.__value__)
             if is_union_origin(get_origin(annotation)) and _union_is_complex(annotation, field.metadata):
                 field_info.append((field_name, self._apply_case_sensitive(self.env_prefix + field_name), True))
--- a/pydantic_settings/sources/utils.py
+++ b/pydantic_settings/sources/utils.py
@@ -40,12 +40,10 @@ def parse_env_vars(
     }
 
 
-def _annotation_is_complex(annotation: type[Any] | None, metadata: list[Any]) -> bool:
+def _annotation_is_complex(annotation: Any, metadata: list[Any]) -> bool:
     # If the model is a root model, the root annotation should be used to
     # evaluate the complexity.
-    if annotation and (
-        typing_objects.is_typealiastype(annotation) or typing_objects.is_typealiastype(get_origin(annotation))
-    ):
+    if typing_objects.is_typealiastype(annotation) or typing_objects.is_typealiastype(get_origin(annotation)):
         annotation = annotation.__value__
     if annotation is not None and _lenient_issubclass(annotation, RootModel) and annotation is not RootModel:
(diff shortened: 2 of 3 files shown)


Viicos:
Thanks @tselepakis, I would suggest using the following patch instead:

```diff
diff --git a/pydantic_settings/sources/base.py b/pydantic_settings/sources/base.py
index b2c4d16..881546b 100644
--- a/pydantic_settings/sources/base.py
+++ b/pydantic_settings/sources/base.py
@@ -16,6 +16,7 @@ from pydantic._internal._typing_extra import (  # type: ignore[attr-defined]
 from pydantic._internal._utils import is_model_class
 from pydantic.fields import FieldInfo
 from typing_extensions import get_args
+from typing_inspection import typing_objects
 from typing_inspection.introspection import is_union_origin
 
 from ..exceptions import SettingsError
@@ -353,6 +354,12 @@ class PydanticBaseEnvSettingsSource(PydanticBaseSettingsSource):
                 field_info.append((v_alias, self._apply_case_sensitive(v_alias), False))
 
         if not v_alias or self.config.get('populate_by_name', False):
+            if typing_objects.is_typealiastype(field.annotation) or typing_objects.is_typealiastype(get_origin(field.annotation)):
+                annotation = field.annotation.__value__
+            else:
+                annotation = field.annotation
+            if typing_objects.is_annotated(get_origin(annotation)):
+                annotation = annotation.__origin__
             if is_union_origin(get_origin(field.annotation)) and _union_is_complex(field.annotation, field.metadata):
                 field_info.append((field_name, self._apply_case_sensitive(self.env_prefix + field_name), True))
             else:
diff --git a/pydantic_settings/sources/utils.py b/pydantic_settings/sources/utils.py
index 270a8c1..498a8dc 100644
--- a/pydantic_settings/sources/utils.py
+++ b/pydantic_settings/sources/utils.py
@@ -43,6 +43,9 @@ def parse_env_vars(
 def _annotation_is_complex(annotation: type[Any] | None, metadata: list[Any]) -> bool:
     # If the model is a root model, the root annotation should be used to
     # evaluate the complexity.
+    if typing_objects.is_typealiastype(annotation) or typing_objects.is_typealiastype(get_origin(annotation)):
+        annotation = annotation.__value__
+
     if annotation is not None and _lenient_issubclass(annotation, RootModel) and annotation is not RootModel:
         annotation = cast('type[RootModel[Any]]', annotation)
         root_annotation = annotation.model_fields['root'].annotation
```

That's really hacky and looking at the code related to this I'm pretty sure it contains many other bugs, but we can't do better considering the current state of `pydantic-settings`.

tselepakis:
Thanks for the review and the detailed patch, @Viicos. I appreciate you taking the time to provide a solution.

I have a question regarding the first change in base.py. You've introduced a new annotation variable that strips away TypeAlias and Annotated wrappers.

However, the is_union_origin check immediately after that, still uses get_origin(field.annotation) instead of get_origin(annotation). Was this intentional, or should the `is_union_origin` check also use the unwrapped `annotation` variable?

Additionally, could you please explain the specific bug or use case that this logic is intended to cover (if I remove this part my test still passes, perhaps we need another test to cover this case)? I'm trying to understand which scenario requires us to check for is_typealiastype and is_annotated here. As you mentioned that the solution is "hacky", understanding the exact case would be very helpful for me.

Viicos:
> However, the is_union_origin check immediately after that, still uses get_origin(field.annotation) instead of get_origin(annotation). Was this intentional, or should the `is_union_origin` check also use the unwrapped `annotation` variable?

That's an oversight sorry, it should check `annotation`, not `field.annotation`.

> Additionally, could you please explain the specific bug or use case that this logic is intended to cover (if I remove this part my test still passes, perhaps we need another test to cover this case)? I'm trying to understand which scenario requires us to check for is_typealiastype and is_annotated here. As you mentioned that the solution is "hacky", understanding the exact case would be very helpful for me.

I believe both options would have worked but yours was more likely to cause unexpected issues as you implemented the type alias "unpacking" in `get_origin()`/`get_args()`, which in theory can be used anywhere in the code base. My proposed change is only scoped to the places where it is necessary to unpack them.

pydantic_settings/sources/base.py
```diff
@@ -353,7 +355,10 @@ def _extract_field_info(self, field: FieldInfo, field_name: str) -> list[tuple[s
                 field_info.append((v_alias, self._apply_case_sensitive(v_alias), False))
 
         if not v_alias or self.config.get('populate_by_name', False):
-            if is_union_origin(get_origin(field.annotation)) and _union_is_complex(field.annotation, field.metadata):
+            annotation = field.annotation
+            if annotation and typing_objects.is_typealiastype(annotation):
```
Viicos, line 359: Is there a specific need to check for the truthiness of `annotation` first? (Which should always evaluate to `True` unless something unexpected is happening).

tselepakis: @Viicos you are absolutely right, I hadn't took into consideration the generic type alias case. As you will see, the `annotation` truthiness check is back as mypy complains that annotation might be `None`.

Viicos (approved):
Thanks @tselepakis!

Fabbi:
this breaks for type aliases with `BeforeValidator` like
```python
from sqlalchemy.engine.url import URL, make_url

type URLStr = Annotated[URL, BeforeValidator(make_url)]

class Settings(BaseSettings):
    DATABASE_URL: URLStr
```

adding `NoDecode` to the typeAlias doesn't work either, because `fields.py:from_annotation` skips unpacking the type alias, thus not picking up the `NoDecode`.

What works is 

```python
from sqlalchemy.engine.url import URL, make_url

type URLStr = Annotated[URL, BeforeValidator(make_url)]

class Settings(BaseSettings):
    DATABASE_URL: Annotated[URLStr, NoDecode]
```

which kind of defeats the purpose of the alias

tselepakis:
Nice catch @Fabbi . I think the problem has to do with the missing metadata here:
https://github.com/pydantic/pydantic-settings/blob/main/pydantic_settings/sources/base.py#L183

@Viicos shall we introduce a new util function to also support annotated fields there? Something like this maybe?
```python
def _get_field_metadata(field: Any) -> list[Any]:
    annotation = field.annotation
    metadata = field.metadata
    if typing_objects.is_typealiastype(annotation) or typing_objects.is_typealiastype(get_origin(annotation)):
        annotation = annotation.__value__
    if typing_objects.is_annotated(origin):
        _, *meta = get_args(annotation)
        metadata += meta
    return metadata
```

If you agree with the approach I can prepare an MR.
https://github.com/pydantic/pydantic-settings/compare/main...tselepakis:pydantic-settings:issue-536-fix

Viicos:
Sorry for the slow response. This approach can work, but more generally dealing with annotations that way is really hard to get right in practice. For instance, it looks like your approach doesn't take into account nested PEP 695 type aliases (e.g the type in `Annotated` is itself a PEP 695 type alias). I think we'd accept this as a quick fix but I won't be surprised if it will not be sufficient for other edge cases.

tselepakis:
I opened a PR for this quick fix and I will have a better look in a couple of days. Thanks for the feedback.

PR #648, opened 2025-07-02
Avoid env_prefix falling back to env vars without prefix

### LINKED ISSUE #441: `env_prefix` falls back to env vars without prefix
Using my example from #437: https://github.com/hofrob/pydantic-settings-env-prefix

If you `uv run hello` the result is:

```
hello='ABC'
```

This value is coming from the first line in `.env`: `HELLO=ABC`

I don't think the intention of `env_prefix` is to just fall back to whatever env var it can find that matches the variable without the prefix. If you have nested or complex settings with multiple API keys managed via different `env_prefix` values, you'll not want to use an API key with the wrong service. I'd consider this API key compromised.

tselepakis:
This is a rather simple fix to avoid falling back to env vars without a prefix. 

Fixes #441

### DIFF
--- a/pydantic_settings/main.py
+++ b/pydantic_settings/main.py
@@ -8,12 +8,11 @@ from collections.abc import Mapping
 from types import SimpleNamespace
 from typing import Any, ClassVar, TypeVar
 
-from pydantic import ConfigDict, create_model
+from pydantic import ConfigDict
 from pydantic._internal._config import config_keys
 from pydantic._internal._signature import _field_name_for_signature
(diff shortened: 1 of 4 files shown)


hramezani:
Thanks @tselepakis for this PR. 

I am busy right now, I will review it later. Meanwhile, could you please post a message on the issue and ask the people there for feedback?

hramezani:
Thanks @tselepakis for the PR. Let's merge the PR and publish it in the next release to see how it works.

PR #649, opened 2025-07-02
CLI Serialization Fixes

kschwab:
* Adds serialization of sub commands.
* Fixes serialization of kebab case of flags.

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -1173,6 +1173,29 @@ CliApp.run(Git, cli_args=['clone', 'repo', 'dir']).model_dump() == {
 
 When executing a subcommand with an asynchronous cli_cmd, Pydantic settings automatically detects whether the current thread already has an active event loop. If so, the async command is run in a fresh thread to avoid conflicts. Otherwise, it uses asyncio.run() in the current thread. This handling ensures your asynchronous subcommands "just work" without additional manual setup.
 
+### Serializing CLI Arguments
+
+An instantiated Pydantic model can be serialized into its CLI arguments using the `CliApp.serialize` method.
+
+```py
+from pydantic import BaseModel
+
+from pydantic_settings import CliApp
+
+
+class Nested(BaseModel):
+    that: int
+
+
+class Settings(BaseModel):
+    this: str
+    nested: Nested
+
+
+print(CliApp.serialize(Settings(this='hello', nested=Nested(that=123))))
+#> ['--this', 'hello', '--nested.that', '123']
+```
+
 ### Mutually Exclusive Groups
 
 CLI mutually exclusive groups can be created by inheriting from the `CliMutuallyExclusiveGroup` class.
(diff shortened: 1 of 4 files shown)


hramezani:
Thanks @kschwab 

Is it a fix for an issue?

kschwab:
@hramezani yes, it is an issue I found locally. I can open a formal ticket if you would like.

`CliApp.serialize` is incorrect when `cli_kebab_case` is enabled. Currently on main it will do the below:

```python
from pydantic_settings import BaseSettings, CliApp

class Cfg(BaseSettings, cli_kebab_case=True):
    foo_kebab: int = 123

args = CliApp.serialize(Cfg())

print(args)
#> ['--foo_kebab', '123']

CliApp.run(Cfg, cli_args=args)
"""
usage: example.py [-h] [--foo-kebab int]
example.py: error: unrecognized arguments: --foo_kebab 123
"""
```

With this fix, it will correctly apply the kebab case to serialization:
```python
from pydantic_settings import BaseSettings, CliApp

class Cfg(BaseSettings, cli_kebab_case=True):
    foo_kebab: int = 123

print(CliApp.serialize(Cfg()))
#> ['--foo-kebab', '123'] 

print(CliApp.run(Cfg, cli_args=args))
#> foo_kebab=123
```

I also missed serialization of subcommands, which addressed as well in this PR. See added [test](https://github.com/pydantic/pydantic-settings/pull/649/files#diff-3dc8e47d20379b4d4316beb241d79a6f8a10d40bcd2e62fa2d58c80a1c671854).

kschwab:
I also just noticed that `cli_implicit_flags` case will also likely need an adjustment.

kschwab:
Ok @hramezani, this is ready for review again. When I looked closer, there were a handful of bugs I found with the original serialization. The change set looks larger then it is. I [reverted](https://github.com/pydantic/pydantic-settings/pull/649/commits/49ab894b1a1c4e48416dab1727a823d1af6134ff) the original serialization implementation, and consolidated it into two standalone functions.

The bugs I found and fixed were:
* Serialization [ordering](https://github.com/pydantic/pydantic-settings/pull/649/files#diff-3dc8e47d20379b4d4316beb241d79a6f8a10d40bcd2e62fa2d58c80a1c671854R2680) could be wrong. i.e., ordering of positionals and subcommands has to be strictly maintained.
* Skip serialization of [defaults](https://github.com/pydantic/pydantic-settings/pull/649/files#diff-3dc8e47d20379b4d4316beb241d79a6f8a10d40bcd2e62fa2d58c80a1c671854R2666). Only modified values should be serialized.
* Handle serialization of [kebab](https://github.com/pydantic/pydantic-settings/pull/649/files#diff-3dc8e47d20379b4d4316beb241d79a6f8a10d40bcd2e62fa2d58c80a1c671854R2411) case.
* Handle [implicit](https://github.com/pydantic/pydantic-settings/pull/649/files#diff-3dc8e47d20379b4d4316beb241d79a6f8a10d40bcd2e62fa2d58c80a1c671854R1522) bool flags.

All the newly added tests currently fail on main.

hramezani:
Thanks @kschwab, Great job!

Issue #654, opened 2025-07-10
CLI integration with existing parser is broken.


braindevices:
version >=2.10 the additional argument cannot be add to the parser
 
test script:
```
from pydantic_settings import CliApp, CliSettingsSource, SettingsConfigDict, BaseSettings
import pydantic_settings
import argparse
class CLISettings(BaseSettings):
    model_config = SettingsConfigDict(
        case_sensitive=True,
        cli_kebab_case=True,
        cli_parse_args=True,
        cli_implicit_flags=True
    )

    test: str
    

def parse_args():
    print(f"{pydantic_settings.VERSION=}")
    parser = argparse.ArgumentParser(
        formatter_class=argparse.RawDescriptionHelpFormatter,
        description=__doc__
    )
    cli_settings = CliSettingsSource(CLISettings, root_parser=parser)
    parser.add_argument(
        "-v","--verbose",
        dest='verbose',
        default=0,
        action="count",
    )
    parsed_args = parser.parse_args()
    print(f"{parsed_args=}")
    settings = CliApp.run(
        CLISettings,
        cli_args=parsed_args,
        cli_settings_source=cli_settings
    )
    return settings

if __name__ == "__main__":
    parse_args()
```

here is result:
```
+ exec python test-pydantic.py -h
pydantic_settings.VERSION='2.10.0'
usage: test-pydantic.py [-h] [--test str]

options:
  -h, --help  show this help message and exit
  --test str  (required)

+ exec python test-pydantic.py --test hello -vvv
pydantic_settings.VERSION='2.10.0'
usage: test-pydantic.py [-h] [--test str]
test-pydantic.py: error: unrecognized arguments: -vvv

```

expecting result as <2.10
```
+ exec python test-pydantic.py -h
pydantic_settings.VERSION='2.9.1'
usage: test-pydantic.py [-h] [--test str] [-v]

options:
  -h, --help     show this help message and exit
  --test str     (required)
  -v, --verbose

+ exec python test-pydantic.py --test hello -vvv
pydantic_settings.VERSION='2.9.1'
parsed_args=Namespace(verbose=3, test='hello')

```

hramezani:
Thanks @braindevices for reporting this.

@kschwab could you please check this?

kschwab:
[...]

PR #611, opened 2025-04-30
fix: Respect 'cli_parse_args' from model_config with settings_customise_sources

karta9821:
Fixes: https://github.com/pydantic/pydantic-settings/issues/610

Selected Reviewer: @samuelcolvin

Issue #610, opened 2025-04-29
Unexpected behaviour in `CliSettingsSource` when `cli_parse_args=True` in `model_config`


trygve-baerland:
Suppose I want to create a simple CLI tool, where the only allowed source is from the command line.
Something along the lines of:

```python
import sys

from pydantic_settings import (
    BaseSettings,
    CliSettingsSource,
    PydanticBaseSettingsSource,
    SettingsConfigDict,
)


class Settings(BaseSettings):
    model_config = SettingsConfigDict(cli_parse_args=True)

    value: int

    @classmethod
    def settings_customise_sources(
        cls,
        settings_cls: type[BaseSettings],
        init_settings: PydanticBaseSettingsSource,
        env_settings: PydanticBaseSettingsSource,
        dotenv_settings: PydanticBaseSettingsSource,
        file_secret_settings: PydanticBaseSettingsSource,
    ) -> tuple[PydanticBaseSettingsSource, ...]:
        """Set up config priority."""
        return (CliSettingsSource(settings_cls),)


sys.argv = ["example.py", "--value", "3"]

print(Settings().model_dump())
```

which, for me, will output the following:

```bash
pydantic_core._pydantic_core.ValidationError: 1 validation error for Settings
value
  Field required [type=missing, input_value={}, input_type=dict]
    For further information visit https://errors.pydantic.dev/2.11/v/missing
```

whereas I would expect `{'value': 3}`.

I understand that in this particular case, I can easily fix this by using `CliSettingsSource(settings_cls, cli_parse_args=True)` instead of using a `model_config`.
However, this requirement is not consistent with other CLI-related config variables (e.g. `cli_prog_name` or `cli_kebab_case`), which will be read from `settings_cls.model_config` if not provided in the `CliSettingsSource` constructor.

My suspicion is that this is due to the check [here](https://github.com/pydantic/pydantic-settings/blob/main/pydantic_settings/sources/providers/cli.py#L251), where both `None` and `False` are used to represent "do not parse command line arguments".
This can very well be the intended behaviour, and there's probably some use case I'm not being properly aware of. In that case, feel free to say so, and I'll happily close this issue. Otherwise, I'm also more than happy to contribute and submit a PR if you want.

hramezani:
@kschwab Do you have time to take a look?

kschwab:
[...]

PR #656, opened 2025-07-15
Remove parsing of command line arguments from `CliSettingsSource.__init__`.

trygve-baerland:
As part of the discussions in #610 and #654 , PR #611 needs to be reverted due to breaking backwards compatability. In particular, it stems from `CliSettingsSource.__init__` eagerly parsing command line arguments when passed with `cli_parse_args`.

This PR is meant as an alternative solution to the problem:

1. The call to `_load_env_vars` in `CliSettingsSource.__init__` is removed.
2. Checks for `CliSettingsSource.env_vars` put in places where the caller needs the arguments to have been parsed. As far as I found, this turned out to be during serialization (in `CliSettingsSource._serialized_args`), and during `CliSettingsSource.__call__`.
3. I've also added a test that's almost identical to the example found in #654 .

> NOTE: This PR is primarily meant as part of a discussion (and as a learning experience for me), so feel free to reject it on account of code quality or any other reason. Any feedback, though, is greatly appreciated.

### DIFF
--- a/pydantic_settings/main.py
+++ b/pydantic_settings/main.py
@@ -391,7 +391,8 @@ class BaseSettings(BaseModel):
             dotenv_settings=dotenv_settings,
             file_secret_settings=file_secret_settings,
         ) + (default_settings,)
-        if not any([source for source in sources if isinstance(source, CliSettingsSource)]):
+        custom_cli_sources = [source for source in sources if isinstance(source, CliSettingsSource)]
+        if not any(custom_cli_sources):
             if isinstance(cli_settings_source, CliSettingsSource):
                 sources = (cli_settings_source,) + sources
             elif cli_parse_args is not None:
@@ -414,6 +415,10 @@ class BaseSettings(BaseModel):
                     case_sensitive=case_sensitive,
                 )
                 sources = (cli_settings,) + sources
+        # We ensure that if command line arguments haven't been parsed yet, we do so.
+        elif cli_parse_args and not custom_cli_sources[0].env_vars:
+            custom_cli_sources[0](args=cli_parse_args)
+
         if sources:
             state: dict[str, Any] = {}
             states: dict[str, dict[str, Any]] = {}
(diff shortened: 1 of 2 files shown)


hramezani:
Thanks @trygve-baerland for the PR.

@kschwab can you take a look at this PR?

kschwab:
Hey @trygve-baerland thanks for the PR. I agree with your intent. I think the case where this matters is during `settings_customise_sources`, where it should pull configurations from the `settings_cls`.

I would revert the above and original changes, as they are still breaking.

However, I think the best place to address this is [here](https://github.com/pydantic/pydantic-settings/blob/main/pydantic_settings/main.py#L394), where if we get a CLI source from `settings_customise_sources`, and it hasn't parsed the args, we parse them. I think the below should be all that is necessary:

```python
custom_cli_source = [source for source in sources if isinstance(source, CliSettingsSource)]
if not custom_cli_source:
    # etc 
elif cli_parse_args and not custom_cli_source[0].env_vars:
    custom_cli_source[0](args=cli_parse_args)
# etc
```

trygve-baerland:
Thank you for your feedback, @kschwab. 

I think your approach makes a lot of sense. It also has the added benefit of being less prone to unintended side effects, which is what got us here in the first place :)

EDIT: I am totally fine with still doing the revert, but I don't see how these changes breaks backwards compatability. Not saying I'm at all certain it doesn't, just that I don't see it.

trygve-baerland:
@kschwab : I've made the changes as you suggested (you were bang on with it).

Further, I've rebased onto #655, and can wait with this PR until that's been merged.
I've also added back the test that will be removed by #655 .

hramezani:
Thanks @trygve-baerland for the update. I merged my PR.

trygve-baerland:
Super, @hramezani ! I've rebased onto `main`, so I think it's good for a second look, @kschwab

kschwab:
Looks good! Thanks @trygve-baerland for the changes.

hramezani:
Thanks all 🙏

PR #658, opened 2025-07-17
turn off allow_abbrev on subparsers

### LINKED ISSUE #657: Subcommands should not be able to use abbreviated flags
#477 disabled abbreviations on the root parser, but not on subparsers.

So if you have a command `foo` with subcommand `bar`, you cannot do `foo --hel`, but can do `foo bar --hel` and it will run `--help`.

mroch:
Fixes #657


hramezani:
@kschwab, please take a look when you have time.

kschwab:
Great catch, thanks @mroch. Looks good 👍🏾

hramezani:
Thanks @kschwab for the review

PR #663, opened 2025-07-31
Warn if model_config sets unused keys for missing settings sources

### LINKED ISSUE #660: 🐛 YAML-related values in config dict are never used
The `SettingsConfigDict` provides three keys related to the `YamlConfigSettingsSource`:

```python
class SettingsConfigDict(ConfigDict, total=False):
    # ...
    yaml_file: PathType | None
    yaml_file_encoding: str | None
    yaml_config_section: str | None
    # ...
```

These are lines 65-67 from `pydantic_settings.main` in `v2.10.1`. Based on the presence of these values, you would expect them to be used in the configuration of the sources (lines 349-432 in `main.py` for me), but they aren't. In fact, they are never used in the entire file.

The initialised settings sources are:

- `DefaultSettingsSource`
- `InitSettingsSource`
- `EnvSettingsSource`
- `DotEnvSettingsSource`
- `SecretsSettingsSource`

## 🔁 Reproduction

You can replicate this behaviour using the following two files in the same directory. Run `main.py` with the command `uv run --script main.py`.

```python
# /// script
# requires-python = ">=3.13"
# dependencies = [
#     "pydantic-settings[yaml]",
# ]
# ///


# main.py
    BaseSettings,
    SettingsConfigDict,
)


class Config(BaseSettings):
    model_config = SettingsConfigDict(yaml_file="config.yaml")

    some_value: int
    another_one: str


settings = Config()
```

```yaml
# config.yaml
some_value: 2
another_one: "A random string"
```

When running this, the default behaviour in this case (where no setting sources are specified except the uninitialized `YamlConfigSettingsSource`) is that an empty dictionary of values is returned from `BaseSettings._settings_build_values`, as seen in lines 430 to 432.

```python
            # no one should mean to do this, but I think returning an empty dict is marginally preferable
            # to an informative error and much better than a confusing error
            return {}
```

This leads to the following error in our case:

```txt
  File "/home/jakob/code/pydantic_settings_bug/main.py", line 23, in <module>
    settings = Config()
  File "/home/jakob/.cache/uv/environments-v2/main-e030d240e66b6dd2/lib/python3.13/site-packages/pydantic_settings/main.py", line 188, in __init__
    super().__init__(
    ~~~~~~~~~~~~~~~~^
        **__pydantic_self__._settings_build_values(
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    ...<27 lines>...
        )
        ^
    )
    ^
  File "/home/jakob/.cache/uv/environments-v2/main-e030d240e66b6dd2/lib/python3.13/site-packages/pydantic/main.py", line 253, in __init__
    validated_self = self.__pydantic_validator__.validate_python(data, self_instance=self)
pydantic_core._pydantic_core.ValidationError: 2 validation errors for Config
some_value
    For further information visit https://errors.pydantic.dev/2.11/v/missing
another_one
    For further information visit https://errors.pydantic.dev/2.11/v/missing
```

As you can see, input_value={} shows the default behavior mentioned above.

## ✅ Workaround

My current solution is to use the `settings_customise_sources` hook. Following is a working example:

```python
# /// script
# requires-python = ">=3.13"
# dependencies = [
#     "pydantic-settings[yaml]",
# ]
# ///


# main.py
    BaseSettings,
    PydanticBaseSettingsSource,
    YamlConfigSettingsSource,
)


class Config(BaseSettings):
    some_value: int
    another_one: str

    @classmethod
        cls,
        return (
            YamlConfigSettingsSource(
                settings_cls=settings_cls,
                yaml_file="config.yaml",
            ),
        )


settings = Config()
print(settings)
```

## 💡 Proposed Fixes

A couple of options for resolving this:

- Remove the YAML-related keys (yaml_file, etc.) from SettingsConfigDict. These are currently misleading and unused.
- Support them properly by integrating YamlConfigSettingsSource into the default source chain, respecting these values.
  - This would introduce a breaking change to the settings_customise_sources hook, since the number of parameters would need to grow.
  - An opt-in mechanism (e.g., via use_yaml=True) could soften the impact.

Removing the unused config keys would also technically be a breaking change, but given they currently do nothing, it seems the cleaner path.

HomerusJa:
When a user now changes the value for one of the values for the [other settings sources](https://docs.pydantic.dev/latest/concepts/pydantic_settings/#other-settings-source) in the model_config but doesn't add these sources via settings_customise_sources, he will receive a descriptive UserWarning.

I added unit tests, but I would love it if someone could review them, as this is one of the first times I've written proper unit tests.

Closes #660

### DIFF
--- a/pydantic_settings/main.py
+++ b/pydantic_settings/main.py
@@ -3,6 +3,7 @@ from __future__ import annotations as _annotations
 import asyncio
 import inspect
 import threading
+import warnings
 from argparse import Namespace
 from collections.abc import Mapping
 from types import SimpleNamespace
@@ -24,10 +25,14 @@ from .sources import (
     DotenvType,
     EnvSettingsSource,
     InitSettingsSource,
+    JsonConfigSettingsSource,
     PathType,
     PydanticBaseSettingsSource,
     PydanticModel,
+    PyprojectTomlConfigSettingsSource,
     SecretsSettingsSource,
+    TomlConfigSettingsSource,
+    YamlConfigSettingsSource,
     get_subcommand,
 )
 
@@ -417,6 +422,8 @@ class BaseSettings(BaseModel):
         elif cli_parse_args not in (None, False) and not custom_cli_sources[0].env_vars:
             custom_cli_sources[0](args=cli_parse_args)  # type: ignore
 
+        self._settings_warn_unused_config_keys(sources, self.model_config)
+
         if sources:
             state: dict[str, Any] = {}
             states: dict[str, dict[str, Any]] = {}
@@ -435,6 +442,36 @@ class BaseSettings(BaseModel):
             # no one should mean to do this, but I think returning an empty dict is marginally preferable
             # to an informative error and much better than a confusing error
             return {}
+
+    @staticmethod
+    def _settings_warn_unused_config_keys(sources: tuple[object, ...], model_config: SettingsConfigDict) -> None:
+        """
+        Warns if any values in model_config were set but the corresponding settings source has not been initialised.
+
+        The list alternative sources and their config keys can be found here:
+        https://docs.pydantic.dev/latest/concepts/pydantic_settings/#other-settings-source
+
+        Args:
+            sources: The tuple of configured sources
+        """
+
+        def warn_if_not_used(source_type: type[PydanticBaseSettingsSource], keys: tuple[str, ...]) -> None:
+            if not any(isinstance(source, source_type) for source in sources):
+                for key in keys:
+                    if model_config.get(key) is not None:
+                        warnings.warn(
+                            f'Config key `{key}` is set in model_config but will be ignored because no '
+                            f'{source_type.__name__} source is configured. To use this config key, add a '
+                            f'{source_type.__name__} source to the settings sources via the '
+                            'settings_customise_sources hook.',
+                            UserWarning,
+                            stacklevel=3,
+                        )
+
+        warn_if_not_used(JsonConfigSettingsSource, ('json_file', 'json_file_encoding'))
+        warn_if_not_used(PyprojectTomlConfigSettingsSource, ('pyproject_toml_depth', 'pyproject_toml_table_header'))
+        warn_if_not_used(TomlConfigSettingsSource, ('toml_file',))
+        warn_if_not_used(YamlConfigSettingsSource, ('yaml_file', 'yaml_file_encoding', 'yaml_config_section'))
 
     model_config: ClassVar[SettingsConfigDict] = SettingsConfigDict(
         extra='forbid',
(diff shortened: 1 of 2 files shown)


tests/test_settings.py
```diff
@@ -3132,3 +3132,33 @@ class Settings(BaseSettings):
 
     s = Settings()
     assert s.model_dump() == {'a': ['one', 'two']}
+
+
+def test_warns_if_config_keys_are_set_but_source_is_missing():
```
hramezani, line 3137: I would change the test like that:
```py
def test_warns_if_config_keys_are_set_but_source_is_missing():
    class Settings(BaseSettings):
            json_file='config.json',
            pyproject_toml_depth=2,
            toml_file='config.toml',
            yaml_file='config.yaml',
            yaml_config_section='myapp',
        )

    with pytest.warns() as record:
        Settings()

    assert len(record) == 5

    key_class_pairs = [
        ('json_file', 'JsonConfigSettingsSource'),
        ('pyproject_toml_depth', 'PyprojectTomlConfigSettingsSource'),
        ('toml_file', 'TomlConfigSettingsSource'),
        ('yaml_file', 'YamlConfigSettingsSource'),
        ('yaml_config_section', 'YamlConfigSettingsSource'),
    ]

    for warning, key_class_pair in zip(record, key_class_pairs):
        assert warning.category is UserWarning
        expected_message = (
            f'Config key `{key_class_pair[0]}` is set in model_config but will be ignored because no '
            f'{key_class_pair[1]} source is configured. To use this config key, add a {key_class_pair[1]} '
            f'source to the settings sources via the settings_customise_sources hook.'
        )
        assert warning.message.args[0] == expected_message
```

pydantic_settings/main.py
```diff
@@ -436,6 +443,36 @@ def _settings_build_values(
...
 
+    @staticmethod
+    def _settings_warn_unused_config_keys(sources: tuple[object, ...], model_config: SettingsConfigDict) -> None:
+        """
+        Warns if any values in model_config were set but the corresponding settings source has not been initialised.
+
+        The list alternative sources and their config keys can be found here:
+        https://docs.pydantic.dev/latest/concepts/pydantic_settings/#other-settings-source
+
+        Args:
```
hramezani, line 454: You need to add `model_config` as well

hramezani:
Thanks @HomerusJa for the PR. I left some comments

tests/test_settings.py
hramezani: > However, the current way we are checking this depends on the order of the key-class pairs, so it depends on an implementation detail of the _settings_warn_unused_config_keys method. Is this a problem?

We don't want to change the order(probably it is not important for the users, but with this test, we will guarantee the warning messages order)

> I purposefully only checked whether the key and class are in the message, and not for the exact formulation of the message, because now this test would fail even if you just fixed a typo. Is matching the message exactly preferable to doing it my way? If yes, why?

Yes, it is important because when introducing a new feature, consistency is essential. Some user code might be dependent on the error message (which I do not prefer)

hramezani (approved):

hramezani:
Thanks @HomerusJa

HomerusJa:
Thanks for your help, @hramezani!

PR #669, opened 2025-08-16
CLI Windows Path Fix

### LINKED ISSUE #647: support for CliPositionalArg[list[Path]] under windows
Hi,

thank you for providing this wonderful library!

I'm using a ... files: CliPositionalArg[list[Path]] in an arguments class, parsed via CliApp.run, and on windows paths .\something\something.exe will hit the json decoder which stumbles over the backslash of the windows path.

currently I'm working around this by escaping any backslashes in sys.argv before passing it to CliApp.run, which does affect using actual json in arguments, which I don't need, but I wonder if there is a better way to solve this.

To me it seems that multiple command line arguments are converted to json before in turn decoded as json, so maybe the encoding to json just needs escaping?

Thank you very much

Moritz

kschwab:
Fixes #647

### DIFF
--- a/pydantic_settings/main.py
+++ b/pydantic_settings/main.py
@@ -645,4 +645,4 @@ class CliApp:
         """
 
         base_settings_cls = CliApp._get_base_settings_cls(type(model))
-        return CliSettingsSource._serialized_args(model, base_settings_cls.model_config)
+        return CliSettingsSource[Any](base_settings_cls)._serialized_args(model)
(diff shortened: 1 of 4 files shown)


hramezani:
Thanks @kschwab for your effort on this PR. really appreciated ❤️

mo22:
@kschwab thanks! can confirm this works.

hramezani (approved):

Issue #679, opened 2025-09-15
Non-Deterministic Field Resolution


enrico-stauss:
Hi,

While I genuinely like pydantic and pydantic settings for many use-cases, I have been batteling it for a few hour now, just to discover that there seems to be non-deterministic behaviour when resolving fields.
It is a bit hard to put in words, so let's provide the basic schema first:
```python
from pydantic import Field, AliasChoices
from pydantic_settings import BaseSettings, SettingsConfigDict

class NestedSettings(BaseSettings):
    model_config = SettingsConfigDict(
        validate_by_alias=True,
        validate_by_name=True,
        extra="allow",
    )
    user: str = Field(validation_alias=AliasChoices("user", "db_user"))

class AppSettings(BaseSettings):
    model_config = SettingsConfigDict(
        env_file_encoding="utf-8",
        env_nested_delimiter="__",
        env_nested_max_split=2,
    )
    connection: NestedSettings
```

and here is a corresponding unit test which essentially contains 4 times the same test case:
```
import os
from schemata import AppSettings

class TestMWE:
    def test_1(self):
        os.environ["DB_USER"] = "inner"
        os.environ["CONNECTION__DB_USER"] = "outer"
        settings = AppSettings(**{"connection": {"user": "explicit"}})
        assert settings.connection.user == "explicit"

    def test_2(self):
        os.environ["DB_USER"] = "inner"
        os.environ["CONNECTION__DB_USER"] = "outer"
        settings = AppSettings(**{"connection": {"user": "explicit"}})
        assert settings.connection.user == "outer"

    def test_3(self):
        os.environ["DB_USER"] = "inner3"
        os.environ["CONNECTION__DB_USER"] = "outer3"
        settings = AppSettings(**{"connection": {"user": "explicit3"}})
        assert settings.connection.user == "explicit3"

    def test_4(self):
        os.environ["DB_USER"] = "inner4"
        os.environ["CONNECTION__DB_USER"] = "outer4"
        settings = AppSettings(**{"connection": {"user": "explicit4"}})
        assert settings.connection.user == "outer4"
```

Now I execute this test in a docker container with `--build --rm` to ensure a fresh and clean environment, but what I get is, that the test cases that fail flip from time to time. Here are 2 subsequent executions:

```bash
$ docker compose run --build --rm test test_mwe.py
[+] Building 1.8s (13/13) FINISHED                                                                                                           docker:default
=================================================================== test session starts ====================================================================
[...]

Issue #426, opened 2024-09-27
env_prefix Ignored in Some Cases in Pydantic Settings


bolshakov:
#### Environment:

- **Pydantic version**: 2.9.2
- **Pydantic settings version**: 2.5.2
- **Python version**: 3.12.5

#### Description:
I encountered an issue where the `env_prefix` seems to be ignored under specific circumstances when using `SettingsConfigDict` in a nested settings structure.

#### Code to Reproduce:

```python
from pydantic_settings import BaseSettings, SettingsConfigDict

class InnerSettings(BaseSettings):
    port: int = 5432

class Settings(BaseSettings):
    model_config = SettingsConfigDict(env_prefix="SETTINGS__", env_nested_delimiter="__")
    
    inner_settings: InnerSettings = InnerSettings()
    port: int = 5433

settings = Settings()
print(settings.model_dump())
```

#### Steps to Reproduce:

1. Run the following test cases:

```bash
# Execution 1: Without env vars
poetry run python test.py
# Output: {'inner_settings': {'port': 5432}, 'port': 5433}

# Execution 2: With SETTINGS__PORT
SETTINGS__PORT=42 poetry run python test.py
# Output: {'inner_settings': {'port': 5432}, 'port': 42}

# Execution 3: With nested SETTINGS__INNER_SETTINGS__PORT
SETTINGS__INNER_SETTINGS__PORT=42 poetry run python test.py
# Output: {'inner_settings': {'port': 42}, 'port': 5433}

# Execution 4: With PORT and nested SETTINGS__INNER_SETTINGS__PORT
PORT=666 SETTINGS__INNER_SETTINGS__PORT=42 poetry run python test.py
# Output: {'inner_settings': {'port': 42}, 'port': 5433}

# Execution 5: With only PORT
PORT=666 poetry run python test.py
# Output: {'inner_settings': {'port': 666}, 'port': 5433}

# Execution 6: With both PORT and SETTINGS__PORT
PORT=666 SETTINGS__PORT=42 poetry run python test.py
# Output: {'inner_settings': {'port': 666}, 'port': 42}
```

#### Expected Behavior:

The `env_prefix` should correctly apply to nested settings and non-prefixed environment variables should be ignored.

#### Actual Behavior:

In some cases, the non-prefixed `PORT` variable overrides the prefixed settings.

hramezani:
[...]

PR #681, opened 2025-09-19
Make `InitSettingsSource` resolution deterministic

enrico-stauss:
This PR fixes the issue described in https://github.com/pydantic/pydantic-settings/issues/679, by making the alias resolution in `InitSettingsSource` deterministically follow the configured precedence.

### DIFF
--- a/pydantic_settings/sources/base.py
+++ b/pydantic_settings/sources/base.py
@@ -268,8 +268,9 @@ class InitSettingsSource(PydanticBaseSettingsSource):
             init_kwarg_name = init_kwarg_names & set(alias_names)
             if init_kwarg_name:
                 preferred_alias = alias_names[0]
+                preferred_set_alias = next(alias for alias in alias_names if alias in init_kwarg_name)
                 init_kwarg_names -= init_kwarg_name
-                self.init_kwargs[preferred_alias] = init_kwargs[init_kwarg_name.pop()]
+                self.init_kwargs[preferred_alias] = init_kwargs[preferred_set_alias]
         self.init_kwargs.update({key: val for key, val in init_kwargs.items() if key in init_kwarg_names})
 
         super().__init__(settings_cls)
--- a/tests/test_settings.py
+++ b/tests/test_settings.py
@@ -699,11 +699,6 @@ def test_alias_resolution_init_source(env):
     assert Example(name='john', PREFIX_SURNAME='doe').model_dump() == {'name': 'john', 'last_name': 'doe'}
 
 
-@pytest.mark.xfail(
-    reason='This test is expected to fail in some sessions due to `InitSettingsSource`s non-deterministic behaviour.'
-    "Simply repeating through parameterization won't help as the underlying reason is the non-deterministic "
-    'nature of python sets, which will remain constant within the same session.'
-)
 def test_init_kwargs_alias_resolution_deterministic():
     class Example(BaseSettings):
         name: str
--- a/tests/test_settings.py
+++ b/tests/test_settings.py
@@ -699,6 +699,21 @@ def test_alias_resolution_init_source(env):
     assert Example(name='john', PREFIX_SURNAME='doe').model_dump() == {'name': 'john', 'last_name': 'doe'}
 
 
+@pytest.mark.xfail(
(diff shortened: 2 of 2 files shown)


hramezani:
Thanks @enrico-stauss for the PR.

@kschwab Could you review the PR?

kschwab:
This is a great catch, thanks @enrico-stauss. Looks good to me. Could simplify to a single line if desired:

```python
init_kwarg_name = sorted(init_kwarg_names & set(alias_names))
```

enrico-stauss:
> This is a great catch, thanks @enrico-stauss. Looks good to me. Could simplify to a single line if desired:
> 
> ```python
> init_kwarg_name = sorted(init_kwarg_names & set(alias_names))
> ```

Not sure that I understand correctly @kschwab , you want to sort the set and then always take the last element by `init_kwarg_name.pop()`? This would make it deterministic but ignore the user-set prioritization, right?

Thanks for the approval, I'm glad I could contribute :)

May I ask for the reason why nesting BaseSettings is disencouraged? Could it be that this PR voids that?

kschwab:
> Not sure that I understand correctly @kschwab , you want to sort the set and then always take the last element by init_kwarg_name.pop()? This would make it deterministic but ignore the user-set prioritization, right?

You're right. I think currently precedence is implicit, but I agree. Stick with what you have, it's more correct.

> May I ask for the reason why nesting BaseSettings is disencouraged? Could it be that this PR voids that?

Your fix is independent of nested BaseSettings. It's a good fix. Nesting BaseSettings can have unintended consequences since all the source modules will be re-ran at different levels. i.e., YMMV. @hramezani can likely provide more insight here.

Otherwise, everything else looks good. Thanks again for the fix!

hramezani:
Thanks @kschwab for the review.

> Nesting BaseSettings can have unintended consequences since all the source modules will be re-ran at different levels.

This is exactly the reason and we mentioned it in the docs as well.

hramezani:
Thanks both ❤️

enrico-stauss:
> Thanks @kschwab for the review.
> 
> 
> 
> > Nesting BaseSettings can have unintended consequences since all the source modules will be re-ran at different levels.
> 
> 
> 
> This is exactly the reason and we mentioned it in the docs as well.
> 
> 
> 
> 

My thoughts were going in the direction of "it maybe not having unintended consequences anymore" 😄

I had obviously read the dochmentation, but wanted to understand a bit better of what nature these unintended consequences are. Is there an assessment or something available?

hramezani:
if the nested model inherits from `BaseSettings` it will be initialized during the parent model init.

Example from https://github.com/pydantic/pydantic-settings/issues/426

```py

class InnerSettings(BaseSettings):
    port: int = 5432

class Settings(BaseSettings):
    model_config = SettingsConfigDict(env_prefix="SETTINGS__", env_nested_delimiter="__")
    
    inner_settings: InnerSettings = InnerSettings()
    port: int = 5433

settings = Settings()
print(settings.model_dump())
```

if we run the above code with `PORT=666 python test.py`, you will get `{'inner_settings': {'port': 666}, 'port': 5433}` which is wrong.
but if the  `InnerSettings` inherits from `BaseModel`, you will get `{'inner_settings': {'port': 5432}, 'port': 5433}` that is correct.

enrico-stauss:
Thanks for the concrete example! I personally would probably consider this a feature, not a bug, since the resolution order makes sense to me. In the example I would encourage users to do:
```python

class InnerSettings(BaseSettings):
    port: int = 5432

class Settings(BaseSettings):
    model_config = SettingsConfigDict(env_prefix="SETTINGS__", env_nested_delimiter="__")
    
    inner_settings: InnerSettings = InnerSettings(port=5432)  # Explicitly pass the default to the initializer
    port: int = 5433
```
if the default to the inner BaseSettings object should take precedence over environment variables.

But I agree that this could be a bit unintuitive, given that there is no way to completely deactivate parsing of non-prefixed environment variables for inner BaseSettings objects. Maybe for Pydantic V3 you could consider having a mechanism that (_by default_) surpresses parsing of non-prefixed environment variables for inner BaseSettings and allows users to opt-in via the SettingsConfigDict of the outer class?

This was my last thoughts on the topic, I promise I'll leave you in peace now :D
Thanks and kind regards
Enrico

PR #686, opened 2025-09-28
Support for enum kebab case.

### LINKED ISSUE #685: How do you use an enum value in --help instead of the member name?
So I use an enum for some configuration and some enums have a `_` in the member field like `example_a` which shows up in the `--help` message. I want to display the enum value `example-a` instead. Is there a way to do this?
I tried `use_enum_values=True` and to define a `__str__` in the enum class, but that doesnt work.

Minimum code to reproduce:
```python
from enum import StrEnum

class Example(StrEnum):
    example_a = "example-a"
    example_b = "example-b"


class Settings(BaseSettings):
    example: Example

if __name__ == "__main__":
    settings = Settings()
```

Current output
```
python main.py --help
usage: main.py [-h] [--example {example_a,example_b}]

options:
  --example {example_a,example_b}
                        (required)
```

Desired output
```
python main.py --help
usage: main.py [-h] [--example {example_a,example_b}]

options:
  --example {example-a,example-b}
                        (required)
```

kschwab:
Fixes #685

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -1354,7 +1354,9 @@ print(Settings().model_dump())
 
 #### CLI Kebab Case for Arguments
 
-Change whether CLI arguments should use kebab case by enabling `cli_kebab_case`.
+Change whether CLI arguments should use kebab case by enabling `cli_kebab_case`. By default, `cli_kebab_case=True` will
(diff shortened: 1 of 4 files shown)


hramezani:
Thanks @kschwab

PR #690, opened 2025-10-05
Add NestedSecretsSettings source

makukha:
Merged code and docs from [pydantic-file-secrets](https://github.com/makukha/pydantic-file-secrets) v0.4.4

### DIFF
--- a/pydantic_settings/sources/base.py
+++ b/pydantic_settings/sources/base.py
@@ -265,12 +265,27 @@ class InitSettingsSource(PydanticBaseSettingsSource):
         init_kwarg_names = set(init_kwargs.keys())
         for field_name, field_info in settings_cls.model_fields.items():
             alias_names, *_ = _get_alias_names(field_name, field_info)
-            init_kwarg_name = init_kwarg_names & set(alias_names)
+            # When populate_by_name is True, allow using the field name as an input key,
+            # but normalize to the preferred alias to keep keys consistent across sources.
+            matchable_names = set(alias_names)
+            include_name = settings_cls.model_config.get('populate_by_name', False)
+            if include_name:
+                matchable_names.add(field_name)
+            init_kwarg_name = init_kwarg_names & matchable_names
             if init_kwarg_name:
-                preferred_alias = alias_names[0]
-                preferred_set_alias = next(alias for alias in alias_names if alias in init_kwarg_name)
+                preferred_alias = alias_names[0] if alias_names else field_name
+                # Choose provided key deterministically: prefer the first alias in alias_names order;
+                # fall back to field_name if allowed and provided.
+                provided_key = next((alias for alias in alias_names if alias in init_kwarg_names), None)
+                if provided_key is None and include_name and field_name in init_kwarg_names:
+                    provided_key = field_name
+                # provided_key should not be None here because init_kwarg_name is non-empty
+                assert provided_key is not None
                 init_kwarg_names -= init_kwarg_name
-                self.init_kwargs[preferred_alias] = init_kwargs[preferred_set_alias]
+                self.init_kwargs[preferred_alias] = init_kwargs[provided_key]
+        # Include any remaining init kwargs (e.g., extras) unchanged
+        # Note: If populate_by_name is True and the provided key is the field name, but
+        # no alias exists, we keep it as-is so it can be processed as extra if allowed.
         self.init_kwargs.update({key: val for key, val in init_kwargs.items() if key in init_kwarg_names})
 
         super().__init__(settings_cls)
--- a/tests/test_precedence_and_merging.py
+++ b/tests/test_precedence_and_merging.py
@@ -1,0 +1,118 @@
+from __future__ import annotations as _annotations
+
+from pathlib import Path
+
+from pydantic import AnyHttpUrl, Field
+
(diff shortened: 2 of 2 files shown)


makukha:
@hramezani the PR is ready for review.

pydantic_settings/sources/providers/nested_secrets.py
```diff
@@ -0,0 +1,168 @@
...
+                raise SettingsError(f'secrets_dir must reference a directory, not a {path_type_label(path)}')
+            secrets_dir_size = sum(f.stat().st_size for f in path.glob('**/*') if f.is_file())
+            if secrets_dir_size > self.secrets_dir_max_size:
+                raise SettingsError(f'secrets_dir size is above {self.secrets_dir_max_size} bytes')
+
+    @staticmethod
+    def load_secrets(path: Path) -> dict[str, str]:
+        return {
+            str(p.relative_to(path)): p.read_text().strip()
+            for p in map(Path, iglob(f'{path}/**/*', recursive=True))
```
Spenhouet, line 154: Instead of `map(Path, iglob(f'{path}/**/*', recursive=True))` you can just do `path.rglob("*")`

pydantic_settings/sources/providers/nested_secrets.py
```diff
@@ -0,0 +1,168 @@
...
+            self.validate_secrets_path(path)
+
+        # construct parent
+        super().__init__(
+            settings_cls,
+            case_sensitive=self.case_sensitive,
+            env_prefix=self.secrets_prefix,
+            env_nested_delimiter=self.secrets_nested_delimiter,
+            env_ignore_empty=False,  # match SecretsSettingsSource behaviour
+            env_parse_none_str=None,  # match SecretsSettingsSource behaviour
```
Spenhouet, line 114: Based on your original implementation: Why not include `env_parse_enums=True`?

pydantic_settings/sources/providers/nested_secrets.py
```diff
@@ -0,0 +1,168 @@
...
+        self.env_parse_none_str = None  # update manually because of None
+
+        # update parent members
+        if not len(self.secrets_paths):
+            self.env_vars = {}
+        else:
+            secrets = reduce(
+                lambda d1, d2: dict((*d1.items(), *d2.items())),
+                (self.load_secrets(p) for p in self.secrets_paths),
+            )
```
Spenhouet, line 125: We can also support nested directories using the nested delimiter e.g. a file path like `db/password` could be mapped to `db__password`, which would write that into a nested model `db` with field `password`.

```python
            # Transform keys for nested delimiter support
            if self.secrets_nested_delimiter and self.secrets_nested_delimiter != os.sep:
                transformed_secrets = {}
                for key, value in secrets.items():
                    new_key = key.replace(os.sep, self.secrets_nested_delimiter)
                    transformed_secrets[new_key] = value
                secrets = transformed_secrets
```

I see that the docs state that this is already supported, so I'm not sure.

docs/index.md
```diff
@@ -1859,6 +1859,262 @@ Last, run your application inside a Docker container and supply your newly creat
...
+    db: DbSettings
+
+    model_config = SettingsConfigDict(
+        env_prefix='MY_',
+        env_nested_delimiter='__',
+        secrets_dir='secrets',
+        secrets_nested_delimiter='_',
+    )
+
+    model_config = SettingsConfigDict(
```
Spenhouet, line 1974: model_config is defined twice

docs/index.md
```diff
@@ -1859,6 +1859,262 @@ Last, run your application inside a Docker container and supply your newly creat
...
```
Spenhouet, line 2043: model_config is defined twice

Spenhouet (approved):

pydantic_settings/sources/providers/nested_secrets.py
makukha: Yea! Nice catch.

pydantic_settings/sources/providers/nested_secrets.py
makukha: Agreed. This won't change anything, but would be better from documentation point of view.

pydantic_settings/sources/providers/nested_secrets.py
makukha: Not sure what you mean here)

Spenhouet: In my implementation I added this, but don't really remember why 😅

Maybe it was to also support the nested delimiter within the secret name i.e. a secret file named `db__password`.

makukha: Oh, I see, you mean the case of both nested directory and file name prefix. This is case already supported: https://github.com/makukha/pydantic-settings/blob/c39546cbe742855b7c96f2f9b5bcc23e64f7e8f1/tests/test_source_nested_secrets.py#L130

hramezani:
Thanks @makukha 
Really great addition. I hope you will help us later in fixing potential bugs and later improvement

makukha:
Thank you, @hramezani, I'm glad that it was helpful. Looking forward to maintain it in future, and potentially merge with `SecretsSettingsSource` in v3.

PR #693, opened 2025-10-13
Coerce env vars if strict is True.

### LINKED ISSUE #675: When using CLI parsing, how to apply strict parsing for other sources?
I can't just set strict on everything because then CLI parsing will fail so this has to  be selectively applied per settings source.
I specifically want this to work with yaml source, but having it succeed with json would be a start as well.

My failing attempt:
```py
from pydantic import ValidationError
    BaseSettings,
    CliApp,
    JsonConfigSettingsSource,
    PydanticBaseSettingsSource,
    SettingsConfigDict,
    YamlConfigSettingsSource,
)


class Settings(BaseSettings):
        cli_parse_args=True,
        cli_kebab_case=False,
        validate_default=False,
    )

    mybool: bool

    @classmethod
        cls,
        strict = True
        json = True

        if strict:
            # Doesn't quite help
            StrictBase = type(
                "StrictBase", (settings_cls,), {"model_config": SettingsConfigDict(strict=True)}
            )
            base = StrictBase
        else:
            base = settings_cls

        if json:
            """
            {
                "mybool": 0
            }
            """
            return (JsonConfigSettingsSource(base, json_file="myconfig.json"),)
        else:
            """
            mybool: 0
            """
            return (YamlConfigSettingsSource(base, yaml_file="myconfig.yaml"),)


def cli():
    try:
        settings = CliApp.run(Settings)
    except ValidationError:
        pass
    else:
        assert False, "Nope, I want ValidationError to occur (and swallow instead)"
```

kschwab:
Fixes #675

### DIFF
--- a/pydantic_settings/main.py
+++ b/pydantic_settings/main.py
@@ -426,6 +426,7 @@ class BaseSettings(BaseModel):
 
         if sources:
             state: dict[str, Any] = {}
+            defaults: dict[str, Any] = {}
             states: dict[str, dict[str, Any]] = {}
             for source in sources:
                 if isinstance(source, PydanticBaseSettingsSource):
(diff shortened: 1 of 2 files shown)


hramezani:
Thanks @kschwab

PR #695, opened 2025-10-20
Unpack type alisases when looking for `NoDecode`

tselepakis:
Provide field metadata utility function to properly handle Annotated types.

The issue was mentioned [here](https://github.com/pydantic/pydantic-settings/pull/644#issuecomment-3345774638).

Fixes https://github.com/pydantic/pydantic-settings/issues/715.

### DIFF
--- a/pydantic_settings/sources/utils.py
+++ b/pydantic_settings/sources/utils.py
@@ -11,6 +11,7 @@ from typing import Any, cast, get_args, get_origin
 from pydantic import BaseModel, Json, RootModel, Secret
 from pydantic._internal._utils import is_model_class
 from pydantic.dataclasses import is_pydantic_dataclass
+from pydantic.fields import FieldInfo
 from typing_inspection import typing_objects
 
 from ..exceptions import SettingsError
@@ -70,6 +71,18 @@ def _annotation_is_complex(annotation: Any, metadata: list[Any]) -> bool:
         or hasattr(origin, '__pydantic_core_schema__')
         or hasattr(origin, '__get_pydantic_core_schema__')
     )
+
+
+def _get_field_metadata(field: FieldInfo) -> list[Any]:
+    annotation = field.annotation
+    metadata = field.metadata
+    if typing_objects.is_typealiastype(annotation) or typing_objects.is_typealiastype(get_origin(annotation)):
+        annotation = annotation.__value__  # type: ignore[union-attr]
+    origin = get_origin(annotation)
+    if typing_objects.is_annotated(origin):
+        _, *meta = get_args(annotation)
+        metadata += meta
+    return metadata
 
 
 def _annotation_is_complex_inner(annotation: type[Any] | None) -> bool:
--- a/pydantic_settings/sources/base.py
+++ b/pydantic_settings/sources/base.py
@@ -24,6 +24,7 @@ from .types import EnvNoneType, ForceDecode, NoDecode, PathType, PydanticModel, 
 from .utils import (
     _annotation_is_complex,
     _get_alias_names,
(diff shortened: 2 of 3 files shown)


tselepakis:
@Viicos  let's continue the discussion about [this](https://github.com/pydantic/pydantic-settings/pull/644#issuecomment-3422352409) here if you like.  I had a quick look on the tests and I wasn't able to find any relevant test that we have nested annotations with NoDecode. Although I understand your concern regarding nesting, I am not sure I can replicate the issue. Do we actually care about the underling structure? I think we only need to identify that this field should not be decoded.

I tried something like this and it seems to work.
```python
IntOrStr = TypeAliasType('IntOrStr', int | str)
A = TypeAliasType('A', Annotated[list[IntOrStr], NoDecode])
```

this also works
```python
ListOrTuple = TypeAliasType('ListOrTuple', list[str] | tuple[str])
A = TypeAliasType('A', Annotated[ListOrTuple, NoDecode])
```

pydantic_settings/sources/utils.py
```diff
@@ -73,6 +73,18 @@ def _annotation_is_complex(annotation: Any, metadata: list[Any]) -> bool:
     )
 
 
+def _get_field_metadata(field: Any) -> list[Any]:
```
Viicos:
```suggestion
def _get_field_metadata(field: FieldInfo) -> list[Any]:
```

Viicos:
> I tried something like this and it seems to work.
> 
> ```python
> IntOrStr = TypeAliasType('IntOrStr', int | str)
> A = TypeAliasType('A', Annotated[list[IntOrStr], NoDecode])
> ```

What I was referring to by "nested" type aliases, is with something like:

```py
B = TypeAliasType('B', Annotated[list[str], NoDecode])
A = TypeAliasType('A', Annotated[B, ...])

class Settings(BaseSettings):
    a: A
```

To be tested but I think right now this isn't taken into account? If not, let's go with what we have currently now, as it looks like this was [a regression](https://github.com/pydantic/pydantic-settings/pull/644#issuecomment-3345774638) initially.

tselepakis:
@Viicos you are correct, we don't take into account this case. I worked a solution but I am not sure if this is the correct approach. Does it makes sense to gather all metadata together or we care only about NoDecode?

```python
def _get_annotation_metadata(annotation: Any) -> list[Any]:
    if typing_objects.is_typealiastype(annotation) or typing_objects.is_typealiastype(get_origin(annotation)):
        annotation = annotation.__value__

    if not typing_objects.is_annotated(origin):
        return []

    inner_annotation, *metadata = get_args(annotation)
    if typing_objects.is_typealiastype(inner_annotation) or typing_objects.is_typealiastype(
        get_origin(inner_annotation)
    ):
        metadata += _get_annotation_metadata(inner_annotation)

    return metadata


def _get_field_metadata(field: FieldInfo) -> list[Any]:
    annotation = field.annotation
    metadata = field.metadata

    metadata += _get_annotation_metadata(annotation)

    return metadata

```

Viicos:
@tselepakis more generally (as I mentioned in https://github.com/pydantic/pydantic-settings/pull/644#issuecomment-3012050230) the approach pydantic-settings is taking isn't really ideal and I fear we are going to face more limitations as time goes by. Let's merge as is and we can iterate later as ideally I'd like to make use of https://github.com/pydantic/typing-inspection/issues/35 for this.

Happy to merge once the remaining feedback is applied.

tselepakis:
Sounds good to me. PR updated. I think we are good to go.

Viicos (approved):
Thanks @tselepakis!

PR #698, opened 2025-10-30
add deep merge functionality to config file sources

### LINKED ISSUE #590: Settings Merging from files doesn't allow for partial overrides
I wrote a class to simulate the .dotenv loading of toml files where it generated the list of toml files to load in.
The problem I discovered is that the file merging is shallow ie if you have two files that both specify the same header key if you stry to override just one setting in the header from the previous file's version it wipes everything else out.

example 

main.toml
```
[db]
host="some_host"
user="some_user"
pss="pass"
port=3306
db="soem_db
```
override.toml
```
[db]
db="some_other_db"
```

The data that gets loaded and then validated is
```
[db]
db="some_other_db"
```
instead of
```
[db]
host="some_host"
user="some_user"
pss="pass"
port=3306
db="soem_other_db
```

This is because the _read_files in ConfigFileSourceMixin uses the update methods of the dict which just replaces the value if the key already exists.

It would be nice to add a function that could recessively merge the dicts together that may one could do partial overrides. Thi affects all the other loaders like yaml and json as well.

pmeier:
Closes #590. @hramezani I took the liberty and put your suggestion from https://github.com/pydantic/pydantic-settings/issues/590#issuecomment-2804282721 into a PR, because I'd like to see this supported out of the box.

I'm aware tests are missing. I'll send them a little later.

Do you want the `deep_merge` flag also to be part of the `SettingsConfigDict`?

### DIFF
--- a/pydantic_settings/sources/base.py
+++ b/pydantic_settings/sources/base.py
@@ -198,17 +198,16 @@ class ConfigFileSourceMixin(ABC):
         if isinstance(files, (str, os.PathLike)):
             files = [files]
         vars: dict[str, Any] = {}
-        update = deep_update if deep_merge else self._shallow_update
         for file in files:
             file_path = Path(file).expanduser()
-            if file_path.is_file():
-                vars = update(vars, self._read_file(file_path))
-        return vars
+            if not file_path.is_file():
+                continue
 
-    def _shallow_update(self, vars: dict[str, Any], updating_vars: dict[str, Any]) -> dict[str, Any]:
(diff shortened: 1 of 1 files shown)


hramezani:
Thanks @pmeier for the PR and sorry for the late response.

Let's keep the flag on the settings source level and not add it to `SettingsConfigDict` for now.

Please update the documentation as well

pmeier:
@hramezani Ready for review now.

hramezani (approved):

hramezani:
Thanks @pmeier

PR #700, opened 2025-11-02
Restore init kwarg names before returning final state dictionary.

### LINKED ISSUE #696: Model Validator (before) does not receive alias in input
I recently upgraded pydantic-settings (2.6.1 -> 2.11.0) and now have the following issue. I have a settings class with fields that have a `validation_alias` using `AliasChoices`. I want to use the aliases to deprecate an old field name without breaking changes. Therefore, I added `model_validator` that checks if old field names are supplied and issues a warning.

```python
# file: alias.py
import warnings
from typing import Any

from pydantic import AliasChoices, Field, model_validator


class Settings(BaseSettings):
    NAME: str = Field(
        default="",
        validation_alias=AliasChoices("NAME", "OLD_NAME"),
    )
    VALUE: str = Field(
        default="",
        validation_alias=AliasChoices("VALUE", "OLD_VALUE"),
    )

    @model_validator(mode="before")
    def check_for_deprecated_attributes(cls, data: Any) -> Any:
        if isinstance(data, dict):
            print("DEBUG: data", data)
            old_keys = {k for k in data.keys() if k.startswith("OLD_")}
            if old_keys:
                warnings.warn(
                    "The config attributes starting with 'OLD_' are deprecated and will be removed in the"
                    " future. Please remove the prefix 'OLD_'."
                    f" The following deprecated attributes were set: {old_keys!r}",
                    DeprecationWarning,
                    stacklevel=4,
                )
        return data

if __name__ == "__main__":
    settings = Settings(OLD_NAME="foo")
```

If I run `python -Wd alias.py`, I get:

```
DEBUG: data {'NAME': 'foo'}
```

I.e., the before-validator already receives the "resolved" field name.

But I would expect the following (which works with v2.6.1):

```
DEBUG: data {'OLD_NAME': 'foo'}
/home/rafael/alias.py:35: DeprecationWarning: The config attributes starting with 'OLD_' are deprecated and will be removed in the future. Please remove the prefix 'OLD_'. The following deprecated attributes were set: {'OLD_NAME'}
  settings = Settings(OLD_NAME="foo")
```

I don't know whether this is a regression or the new behavior is expected.

Thanks for your feedback! 🙂

kschwab:
Fixes #696

### DIFF
--- a/pydantic_settings/main.py
+++ b/pydantic_settings/main.py
@@ -35,6 +35,7 @@ from .sources import (
     YamlConfigSettingsSource,
     get_subcommand,
 )
+from .sources.utils import _get_alias_names
 
 T = TypeVar('T')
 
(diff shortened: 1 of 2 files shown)


hramezani:
Thanks @kschwab

Issue #697, opened 2025-10-30
Support of Python 3.14


tschm:
Thank you

hramezani:
Thanks @tschm for this issue.

we will add it soon

hramezani:
[pydantic-settings 2.12.0](https://pypi.org/project/pydantic-settings/) with `Python 3.14` support has been released!

PR #704, opened 2025-11-07
Add Python 3.14

hramezani:
Fixes https://github.com/pydantic/pydantic-settings/issues/697


hramezani:
@kschwab I created a PR to add `Python 3.14` support.

There are some CLI failing tests that are related to the output difference.
Should we change the test and assert a different output based on the Python version?

hramezani:
@kschwab I merged the PR. please take a look at the change that I made for cli tests and improve them later if you think there is a better way to handle them. Thanks!

Issue #706, opened 2025-11-10
Deterministic alias selection for init kwargs doesn't support `validate_by_name`


Mantisus:
Recent fix #688 works perfectly with `populate_by_name`, but doesn't support `validate_by_name`.

Current Pydantic [documentation](https://docs.pydantic.dev/latest/api/config/#pydantic.config.ConfigDict.populate_by_name) recommends using `ConfigDict(validate_by_name=True, validate_by_alias=True)` instead of `ConfigDict(populate_by_name=True)`

hramezani:
Thanks @Mantisus for reporting this.

@chbndrhnns Could you please take a look?

chbndrhnns:
I am not certain what is expected here but I created a MR with how I understand the problem. Let me know if that's what you want.

hramezani:
Thanks @chbndrhnns for the PR.

@Mantisus Could you please confirm the fix?

Mantisus:
@hramezani Yes, everything works correctly with this fix. 🙂 

Thanks @chbndrhnns

PR #707, opened 2025-11-10
fix: Deterministic alias selection when using validate_by_name

chbndrhnns:
Fixes https://github.com/pydantic/pydantic-settings/issues/706


hramezani:
Thanks @chbndrhnns

PR #708, opened 2025-11-17
Add support for AWS Secrets Manager VersionId parameter

jcyamacho:
### Summary
Adds support for specifying a `VersionId` when retrieving secrets from AWS Secrets Manager, enabling retrieval of specific secret versions.

### Changes
- Added `version_id` parameter to `AWSSecretsManagerSettingsSource` constructor
- Fixed handling of optional `VersionId` to only include it in the request when provided (prevents errors when `version_id` is `None`)

### Implementation Details
- The `version_id` parameter is optional and defaults to `None`
- When `version_id` is provided, it's passed to the AWS Secrets Manager `get_secret_value` API call
- The fix ensures `VersionId` is only included in the request dictionary when a value is provided, avoiding potential API errors

### Files Changed
- `pydantic_settings/sources/providers/aws.py`

### DIFF
--- a/pydantic_settings/sources/providers/aws.py
+++ b/pydantic_settings/sources/providers/aws.py
@@ -60,9 +60,7 @@ class AWSSecretsManagerSettingsSource(EnvSettingsSource):
         )
 
     def _load_env_vars(self) -> Mapping[str, str | None]:
-        request = {
-            'SecretId': self._secret_id,
-        }
+        request = {'SecretId': self._secret_id}
 
         if self._version_id:
             request['VersionId'] = self._version_id


pydantic_settings/sources/providers/aws.py
```diff
@@ -58,7 +60,14 @@ def __init__(
         )
 
     def _load_env_vars(self) -> Mapping[str, str | None]:
-        response = self._secretsmanager_client.get_secret_value(SecretId=self._secret_id)  # type: ignore
+        request = {
+            'SecretId': self._secret_id,
+        }
```
hramezani:
```suggestion
        request = {'SecretId': self._secret_id}
```

hramezani:
Thanks @jcyamacho

PR #717, opened 2025-11-25
CliToggleFlag and CliDualFlag

### LINKED ISSUE #703: Option to disable generation of `--no-<flag>` for implicit boolean flags
With `cli_implicit_flags=True`, pydantic-settings always generates both `--flag` and `--no-flag`. For some CLIs this is noisy and undesirable; we’d like a way to expose only the positive form.

**Current behavior**
Boolean fields become implicit flags and the help shows `--flag, --no-flag` (mirrors argparse’s `BooleanOptionalAction`).

**Desired behavior**
Add an opt-in switch to disable the negated form, at least at model level, with a per-field override. For example:

* `SettingsConfigDict(cli_boolean_negation="both" | "positive_only", ...)` (default: `"both"`), **or**
* a field-level annotation like `CliPositiveOnlyFlag[bool]`, **or**
* a field kwarg e.g. `Field(cli_negation=False)`.

**Why**

* Cleaner help (`-v | --version` without `--no-version`).
* Matches many established CLIs where only the positive form exists; reduces confusion and boilerplate in docs.
* Backward compatible if default stays as today.

**Alternatives considered**
Integrating a custom root parser to replace `BooleanOptionalAction` with `store_true` only for certain flags. Works, but is a workaround and leaks argparse details into user code.

**References**

* Docs: implicit boolean flags always generate `--flag/--no-flag`. ([Pydantic](https://docs.pydantic.dev/latest/api/pydantic_settings))
* Related discussion about boolean flag UX. ([GitHub](https://github.com/pydantic/pydantic-settings/issues/361))

**Environment**
pydantic-settings 2.11.0
Python 3.13.8

kschwab:
Fixes #703

### DIFF
--- a/pydantic_settings/sources/base.py
+++ b/pydantic_settings/sources/base.py
@@ -24,6 +24,7 @@ from .types import EnvNoneType, ForceDecode, NoDecode, PathType, PydanticModel, 
 from .utils import (
     _annotation_is_complex,
     _get_alias_names,
+    _get_field_metadata,
     _get_model_fields,
     _strip_annotated,
     _union_is_complex,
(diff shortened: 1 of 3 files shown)


hramezani (approved):

hramezani:
Thanks @kschwab

PR #720, opened 2025-11-27
Hide default None value in cli_hide_none_type is set

chovanecadam:
When a field defaults to None, displaying `(default: null)` provides no additional information to users. The absence of a default value already implies None/null for optional fields, and this matches the behavior of most CLI tools (argparse, click) where optional arguments don't display `default: None`.  Users expect optional CLI arguments to default to None implicitly.

This change aligns pydantic-settings with standard CLI tool conventions (argparse, click) where None defaults are hidden. Also, this reduces visual noise in help messages:

```python
class Settings(BaseSettings):
        cli_hide_none_type=True,
    )

    foo: Path | None
    bar: str | None
    baz: str | None = None
    timeout: int = 50
    config: Path | None = None
```

**before**

```
usage: main.py [-h] [--foo Path] [--bar str] [--baz str] [--timeout int] [--config Path]

options:
  -h, --help     show this help message and exit
  --foo Path     (required)
  --bar str      (required)
  --baz str      (default: null)
  --timeout int  (default: 50)
  --config Path  (default: null)
```

**after**

```
usage: main.py [-h] [--foo Path] [--bar str] [--baz str] [--timeout int] [--config Path]

options:
  -h, --help     show this help message and exit
  --foo Path     (required)
  --bar str      (required)
  --baz str
  --timeout int  (default: 50)
  --config Path
```

As this MR alters the behavior of `cli_hide_none_type`, you might prefer adding another flag to support this, e.g. `cli_hide_none_type_defaults`?

### DIFF
--- a/pydantic_settings/sources/providers/cli.py
+++ b/pydantic_settings/sources/providers/cli.py
@@ -1258,7 +1258,9 @@ class CliSettingsSource(EnvSettingsSource, Generic[T]):
                 ifdef = 'ifdef: ' if model_default is None else ''
                 _help += f' ({ifdef}required)' if _help else f'({ifdef}required)'
         else:
-            default = f'(default: {self.cli_parse_none_str})'
+            default = ''
+            if not self.cli_hide_none_type:
+                default = f'(default: {self.cli_parse_none_str})'
             if is_model_class(type(model_default)) or is_pydantic_dataclass(type(model_default)):
                 default = f'(default: {getattr(model_default, field_name)})'
             elif model_default not in (PydanticUndefined, None) and _is_function(model_default):


hramezani:
Thanks @chovanecadam for this patch. We need a test for this.

@kschwab Could you please take a look when you have time?

kschwab:
Thanks @chovanecadam for the changes. Instead of creating a new flag, can you instead amend `cli_hide_none_type` to accept a string. For reference you can see https://github.com/pydantic/pydantic-settings/pull/717/files#diff-d6b3b9e867bfaecf05a505710da6459825c33cf8d191bd12cf82265e5f272763R191

Other than that, looks good. Just needs docs and tests.

hramezani:
@chovanecadam Could you address @kschwab comments?

hramezani:
Closing because of inactivity!

PR #727, opened 2025-12-06
CLI fix fox external list args.

### LINKED ISSUE #726: CliSettingsSource breaks external list arguments
In the following example i am adding two external cli arguments with action `append` in addition to a CliSettings source. Notice that after pydantic-settings parses the arguments the type and value of those external arguments change from `list` to `str`.
Depending on the contents of the list the change is different even; a `list[str]` changes to `'["abc"]'` but a `list[int]` changes to the empty string.

```python
    CliApp,
    CliSettingsSource,
    SettingsConfigDict,
    BaseSettings,
)
import pydantic_settings
import argparse


class CLISettings(BaseSettings):


def parse_args():
    print(f"{pydantic_settings.VERSION=}")
    parser = argparse.ArgumentParser(
        formatter_class=argparse.RawDescriptionHelpFormatter, description=__doc__
    )
    cli_settings = CliSettingsSource(CLISettings, root_parser=parser)
    parser.add_argument(
        "--test-cli",
        action="append",
        default=[1],
    )
    parser.add_argument(
        "--test-cli2",
        action="append",
        default=["abc"],
    )
    parsed_args = parser.parse_args()
    print(f"{parsed_args=}")
    settings = CliApp.run(
        CLISettings, cli_args=parsed_args, cli_settings_source=cli_settings
    )
    print(f"{parsed_args=}")
    return settings


def main() -> None:
    parse_args()


if __name__ == "__main__":
    main()
```
```console
$ uv run test-settings-nargs                                                                                                                                                     ❮❮
pydantic_settings.VERSION='2.12.0'
parsed_args=Namespace(test_cli=[1], test_cli2=['abc'])
parsed_args=Namespace(test_cli='', test_cli2='["abc"]')
```

The culprit seems to be [`CliSettingsSource._merge_parsed_list(...)`](https://github.com/pydantic/pydantic-settings/blob/70e7a28041704add73b02722efd3e4a7cd8b0ca6/pydantic_settings/sources/providers/cli.py#L630) which always converts the result to `str` after merging. Note that this error also existed on version 2.10, so #669 did not introduce it.

kschwab:
Fixes #726

### DIFF
--- a/pydantic_settings/sources/providers/cli.py
+++ b/pydantic_settings/sources/providers/cli.py
@@ -2,6 +2,7 @@
 
 from __future__ import annotations as _annotations
 
+import copy
 import json
 import re
 import shlex
@@ -476,7 +477,7 @@ class CliSettingsSource(EnvSettingsSource, Generic[T]):
                 args = sys.argv[1:]
             return self._load_env_vars(parsed_args=self._parse_args(self.root_parser, args))
         elif parsed_args is not None:
-            return self._load_env_vars(parsed_args=parsed_args)
+            return self._load_env_vars(parsed_args=copy.copy(parsed_args))
         else:
             return super().__call__()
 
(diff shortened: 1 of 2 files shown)


hramezani:
Thanks @kschwab

PR #730, opened 2025-12-12
fix: handle case-insensitive retrieval in GoogleSecretManagerSettingsSource

### LINKED ISSUE #728: `GoogleSecretManagerSettingsSource` fails to retrieve uppercase secrets when `case_sensitive=False`
**Description:**
When `GoogleSecretManagerSettingsSource` is initialized with `case_sensitive=False`, it lowercases all requested keys. It then attempts to fetch secrets from Google Secret Manager using these lowercase names. Since Google Secret Manager resource names are case-sensitive, this fails for any secret that actually contains uppercase letters.

The current implementation catches the resulting `google.api_core.exceptions.NotFound` (or `PermissionDenied`) exception within a broad `except Exception:` block and silently returns `None`, masking the issue.

**Steps to Reproduce:**
1. Create a secret in Google Secret Manager named `MY-SECRET`.
2. Define a Pydantic Settings model with `case_sensitive=False` (default for `BaseSettings` if not overridden, or explicitly set).
3. Attempt to load settings using `GoogleSecretManagerSettingsSource`.
4. The source will request `projects/.../secrets/my-secret/versions/latest`.
5. GCP returns 404 Not Found.
6. `pydantic-settings` returns `None` for the field.

**Expected Behavior:**
The source should handle case-insensitivity by resolving the case-insensitive key to the actual case-sensitive secret name present in the GCP project (if it exists).

**Technical Details:**
The `GoogleSecretManagerMapping` needs to be updated to:
1.  Store a mapping of `normalized_name -> actual_name` during the `list_secrets` phase or lazily.
2.  Use this mapping to look up the correct secret ID before calling `access_secret_version`.

**Reproduction Test Code:**
```python
import pytest
from pydantic_settings.sources.providers.gcp import GoogleSecretManagerMapping
from google.cloud.secretmanager import SecretManagerServiceClient

def test_reproduce_issue(mocker):
    # Mock GCP Client
    client = mocker.Mock(spec=SecretManagerServiceClient)
    client.common_project_path.return_value = 'projects/test-project'
    client.secret_version_path = lambda project, secret, version: f'projects/{project}/secrets/{secret}/versions/{version}'
    client.parse_secret_path = SecretManagerServiceClient.parse_secret_path
    
    # The secret exists in GCP as UPPERCASE
    secret = mocker.Mock()
    secret.name = 'projects/test-project/secrets/MY-SECRET'
    client.list_secrets.return_value = [secret]

    # Setup success for exact match
    secret_response = mocker.Mock()
    secret_response.payload.data.decode.return_value = 'secret-value'
    
    def mock_access(name):
        # GCP is case-sensitive!
        if 'MY-SECRET' in name:
            return secret_response
        raise Exception("Not Found") # GCP behavior for mismatch case
    
    client.access_secret_version = mock_access

    # Initialize Source with case_sensitive=False
    source = GoogleSecretManagerMapping(client, 'test-project', case_sensitive=False)
    
    # This fails (returns None) because it tries to fetch 'my-secret'
    # Expected: should return 'secret-value'
    assert source['my-secret'] == 'secret-value'
```

ezwiefel:
Fixes #728.

This PR updates `GoogleSecretManagerMapping` to support robust case-insensitive secret retrieval.

**Changes:**
- Maintains a map of both exact and lowercased secret names to their actual GCP identifiers.
- Ensures that secrets with uppercase or mixed-case names (e.g. `TEST_SECRET`, `mySecret`) are correctly retrieved when `case_sensitive=False`.
- Implements deterministic collision resolution: if multiple secrets normalize to the same lowercase key (e.g. `MY_SECRET` and `my_secret`), the lexicographically last one (preferring lowercase) is chosen.
- Issues a `UserWarning` when such collisions occur.
- Adds regression tests covering various casing scenarios and collision handling.
- Updates documentation to explain the new behavior.

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -2452,6 +2452,9 @@ For nested models, Secret Manager supports the `env_nested_delimiter` setting as
 ### Important Notes
 
 1. **Case Sensitivity**: By default, secret names are case-sensitive.
+    *   If you set `case_sensitive=False`, `pydantic-settings` will attempt to resolve secrets in a case-insensitive manner. It prioritizes exact matches over case-insensitive matches. For some examples of this, imagine `case_sensitive=False` and the model attribute is named `my_secret`:
+        * If Google Secret Manager has both `MY_SECRET` and `my_secret` defined - the value of `my_secret` will be returned.
+        * If Google Secret Manager has `MY_SECRET`, `My_Secret`, and `my_Secret` defined - a warning will be raised and the value of `my_Secret` will be returned - as the secret names are first sorted in ASCII sort order (where lowercased letters are greater than upper case letters) and the last one is chosen (which would be `my_Secret` in this case).
 2. **Secret Naming**: Create secrets in Google Secret Manager with names that match your field names (including any prefix). According the [Secret Manager documentation](https://cloud.google.com/secret-manager/docs/creating-and-accessing-secrets#create-a-secret), a secret name can contain uppercase and lowercase letters, numerals, hyphens, and underscores. The maximum allowed length for a name is 255 characters.
 3. **Secret Versions**: The GoogleSecretManagerSettingsSource uses the "latest" version of secrets.
 


hramezani:
Thanks @ezwiefel

PR #740, opened 2026-01-01
Remove `(default: …)` in the help message for `CliToggleFlag`

### LINKED ISSUE #737: Suppress the `(default: …)` display for toggle options.
Toggle options were introduced in #717, but I think it would be better to disable the `(default: …)` description for toggle options.

In the example below, it’s unclear to users what `(default: False)` is referring to.

```sh
$ python app.py -h
usage: app.py [-h] [--verbose]

options:
  --verbose   verbose log output (default: False)
```

```python
class RootCommand(
    BaseSettings,
    cli_parse_args=True,
    cli_implicit_flags='toggle',
):
    verbose: CliToggleFlag[bool] = Field(default=False, description='verbose log output')

    def cli_cmd(self):
        print(self)


    CliApp.run(RootCommand, cli_args=args)


if __name__ == '__main__':
    main()
```

It’s even more confusing when `default=True`.

```sh
$ python app.py -h
usage: app.py [-h] [--no-verbose]

options:
  -h, --help    show this help message and exit
  --no-verbose  Suppress log output (default: True)
```

```python
class RootCommand(
    BaseSettings,
    cli_parse_args=True,
    cli_implicit_flags='toggle',
):
    verbose: CliToggleFlag[bool] = Field(default=True, description='Suppress log output')
```

kzrnm:
Fix #737

Stopped showing `(default: True)` for `CliToggleFlag`.
If no description is provided, it displays `(toggle)` instead.

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -168,19 +168,70 @@ There are two ways to do this:
 
 Check the [`Field` aliases documentation](fields.md#field-aliases) for more information about aliases.
 
-`env_prefix` does not apply to fields with alias. It means the environment variable name is the same
-as field alias:
+
+To apply `env_prefix` not only to variable names but also to aliases, set `env_prefix_target='all'`.
+To apply `env_prefix` only to aliases and not to variable names, set `env_prefix_target='alias'`.
+To apply `env_prefix` only to variable names (the default behavior), set `env_prefix_target='variable'`.
+
 
 ```py
+import os
+
 from pydantic import Field
 
 from pydantic_settings import BaseSettings, SettingsConfigDict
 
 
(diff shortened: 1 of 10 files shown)


hramezani:
Thanks @kzrnm for the PR

@kschwab Could you please review the PR when you have time?

kschwab:
@kzrnm thanks for the PR. I agree with #737. Can you update the PR to suppress it entirely as opposed to adding `(toggle)` if no description is provided. It's cleaner and more consistent that way. Other than that, everything else looks good.

kschwab:
Thanks @kzrnm 👍🏾 

@hramezani looks good to me.

Issue #744, opened 2026-01-02
Race condition on initialization in v2.12.0


tomas-bouda-myriad:
After upgrade to v2.12.0 we are encountering a weird behavior during initialization of app settings.

**Multiple runs** of interpreter lead to **non-deterministic** successes and fails, i.e. run this script manually 10 times to witness 5 fails and 5 successes.

```python
import os
from typing import Annotated

from pydantic import Field
from pydantic_settings import BaseSettings as PydanticBaseSettings
from pydantic_settings import SettingsConfigDict


class MyAppSettings(PydanticBaseSettings):
    model_config = SettingsConfigDict(populate_by_name=True)
    env_kind: Annotated[str, Field(alias="ENV_KIND2")]


def main() -> None:
    os.environ["ENV_KIND"] = "dev"
    settings = MyAppSettings.model_validate({"env_kind": "hosted"})
    print(settings)


if __name__ == "__main__":
    main()
```

1. Success
```
env_kind='hosted'
```

2. Error
```
pydantic_core._pydantic_core.ValidationError: 1 validation error for MyAppSettings
env_kind
  Extra inputs are not permitted [type=extra_forbidden, input_value='dev', input_type=str]
    For further information visit https://errors.pydantic.dev/2.12/v/extra_forbidden
```

When you downgrade to v2.11.0 the problem disappears.
```
uv pip list | grep pydantic
pydantic                                 2.12.5
pydantic-ai-slim                         1.25.1
pydantic-core                            2.41.5
pydantic-graph                           1.25.1
pydantic-settings                        2.12.0

uv run python --version
Python 3.11.12
```

By trial and error this problem seems to result from a combination of 
- value set in **ENV** var
- field using **alias** (and **populate_by_name**)
- calling `model_validate` with override

Simply take my script and try to manually run it multiple times. It should alternate between ok and error. 

According to changelog there were modifications of source order in 2.12.0 and we guess there's a non-deterministic merging of dictionaries when the variable happens to be set in multiple sources (just a guess).

hramezani:
Thanks @tomas-bouda-myriad for reporting this issue.
I think it was introduced in https://github.com/pydantic/pydantic-settings/commit/7a6e96ebfc89ddb5b20dfb068c8ed46b6ff18a7f
@chbndrhnns Could you please take a look?

chbndrhnns:
The bug occurs when:

1. A field has an alias (e.g., 'ENV_KIND2') 
[...]

PR #746, opened 2026-01-02
fix: When restoring init kwargs, use deterministic order

chbndrhnns:
Fixes https://github.com/pydantic/pydantic-settings/issues/744

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -863,10 +863,10 @@ from pydantic_settings import BaseSettings
 
 class User(BaseSettings, cli_parse_args=True):
     first_name: str = Field(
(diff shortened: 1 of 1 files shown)


hramezani:
Thanks @chbndrhnns for the PR.

Please add test for your changes.

Thanks!

chbndrhnns:
I added the test. However, as [pointed out](https://github.com/pydantic/pydantic-settings/issues/744#issuecomment-3705463331), we would need to set a certain hash seed BEFORE python starts. I am not sure how to integrate this in the test. With the fix now, we should never see it failing, though.

hramezani (approved):

hramezani:
Thanks @chbndrhnns

PR #747, opened 2026-01-02
Avoid conflicts with the `NAME` environment variable in WSL

### LINKED ISSUE #745: The alias documentation is incorrect
```py
import sys

from pydantic import AliasChoices, AliasPath, Field

class User(BaseSettings, cli_parse_args=True):
    first_name: str = Field(
        validation_alias=AliasChoices('f', 'fname', AliasPath('name', 0))
    )
    last_name: str = Field(
        validation_alias=AliasChoices('l', 'lname', AliasPath('name', 1))
    )


sys.argv = ['example.py', '--fname', 'John', '--lname', 'Doe']
print(User().model_dump())
# pydantic_settings.exceptions.SettingsError: error parsing value for field "first_name" from source "EnvSettingsSource"
```

https://docs.pydantic.dev/latest/concepts/pydantic_settings/#aliases

kzrnm:
Fix #745

On WSL, the `NAME` environment variable is set automatically, which causes an error when the example is run as-is via copy and paste.
To avoid this, it was renamed to `fullname`.

I also considered using `USERNAME`, but avoided it as well since it appears to be used by zsh.

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -863,10 +863,10 @@ from pydantic_settings import BaseSettings
 
 class User(BaseSettings, cli_parse_args=True):
     first_name: str = Field(
-        validation_alias=AliasChoices('f', 'fname', AliasPath('name', 0))
+        validation_alias=AliasChoices('f', 'fname', AliasPath('fullname', 0))
     )
     last_name: str = Field(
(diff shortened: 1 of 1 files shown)


hramezani:
Thanks @kzrnm

PR #749, opened 2026-01-02
Add env_prefix_target

### LINKED ISSUE #748: There is no way to apply `env_prefix` to aliases
Even though `env_prefix='USER_'` is set as shown below, environment variables without the prefix are still being used.

It’s hard to believe that this behavior is intentional—having `env_prefix` applied but ignored for aliases is not a realistic use case. This seems to be the same kind of issue described in #441.

Is this behavior meant to mirror how `Field(env=...)` worked in v1 by mapping it to `Field(validation_alias=...)` in v2?
Ignoring the prefix makes sense for the `env` argument, but doing so for `validation_alias` seems like a bug to me.
If ignoring the prefix is necessary, shouldn’t a separate mechanism be added to explicitly ignore the prefix?

Introduced in #25
related: #220, #245, #441, #648

```py
import os
import sys

from pydantic import AliasChoices, Field

class User(BaseSettings, cli_parse_args=True):
    model_config = SettingsConfigDict(env_prefix='USER_')
    first_name: str = Field(validation_alias=AliasChoices('fname'))
    last_name: str = Field(validation_alias=AliasChoices('l', 'lname'))

    @classmethod
    def dump(cls):
        try:
            print(cls().model_dump())
        except Exception as e:
            print(e)
        print('----')


sys.argv = ['example.py']

os.environ.clear()
os.environ['FNAME'] = 'John'
os.environ['LNAME'] = 'Doe'
User.dump()
# {'first_name': 'John', 'last_name': 'Doe'}

os.environ.clear()
os.environ['USER_FNAME'] = 'John'
os.environ['USER_LNAME'] = 'Doe'

User.dump()
# pydantic_core._pydantic_core.ValidationError: 2 validation errors for User
# fname
#   Field required [type=missing, input_value={}, input_type=dict]
#     For further information visit https://errors.pydantic.dev/2.12/v/missing
# l
#   Field required [type=missing, input_value={}, input_type=dict]
#     For further information visit https://errors.pydantic.dev/2.12/v/missing
```

## A concrete example of when this becomes a problem

As a common case, consider a `--path` parameter with a short alias `-p`, and assume `env_prefix='USER_'`.

Without an alias, I get the expected error when no arguments are provided. With an alias, however, `$PATH` is passed instead. This is almost certainly not what anyone would want.

```py
class LongOnly(
    BaseSettings,
    cli_parse_args=True,
    cli_implicit_flags='toggle',
):
    model_config = SettingsConfigDict(env_prefix='USER_')
    path: pathlib.Path


class LongShort(
    BaseSettings,
    cli_parse_args=True,
    cli_implicit_flags='toggle',
):
    model_config = SettingsConfigDict(env_prefix='USER_')
    path: pathlib.Path = Field(validation_alias=AliasChoices('p', 'path'))


    print(CliApp.run(LongShort, cli_args=args))
    print(CliApp.run(LongOnly, cli_args=args))


if __name__ == '__main__':
    main()
```

kzrnm:
Fix #748

This PR allows `env_prefix` to be applied to aliases as well, without introducing any breaking changes.

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -168,22 +168,6 @@ There are two ways to do this:
 
 Check the [`Field` aliases documentation](fields.md#field-aliases) for more information about aliases.
 
-`env_prefix` does not apply to fields with alias. It means the environment variable name is the same
-as field alias:
-
-```py
-from pydantic import Field
-
-from pydantic_settings import BaseSettings, SettingsConfigDict
-
-
-class Settings(BaseSettings):
-    model_config = SettingsConfigDict(env_prefix='my_prefix_')
-
-    foo: str = Field('xxx', alias='FooAlias')  # (1)!
-```
-
-1. `env_prefix` will be ignored and the value will be read from `FooAlias` environment variable.
 
 To apply `env_prefix` not only to variable names but also to aliases, set `env_prefix_target='all'`.
 To apply `env_prefix` only to aliases and not to variable names, set `env_prefix_target='alias'`.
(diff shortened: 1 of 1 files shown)


hramezani:
We have to remove the old doc because by this change `env_prefix` can work with `alias`
https://github.com/pydantic/pydantic-settings/pull/749#discussion_r2659618988

hramezani:
Thanks @kzrnm

PR #755, opened 2026-01-13
Add support for CLI serialize styles.

### LINKED ISSUE #753: CLI argument serializing support for Argparse and Lazy styles
Currently all `(dict, list, set)` types get serialized using JSON style when calling `CliApp.serialize` ([code](https://github.com/pydantic/pydantic-settings/blob/v2.12.0/pydantic_settings/sources/providers/cli.py#L1329))

Meanwhile CLI argument parsing of lists supports intermixing of any of the below three styles ([link](https://docs.pydantic.dev/latest/concepts/pydantic_settings/#lists)):

- JSON style: `--field='[1,2]'`
- Argparse style: `--field 1 --field 2`
- Lazy style: `--field=1,2`

It would be helpful to allow specifying which style the user wants when calling `CliApp.serialize`.  For example, something like `CliApp.serialize(model, style="argparse")`

kschwab:
Fixes #753


hramezani:
Thanks @kschwab

PR #761, opened 2026-01-21
Use Pydantic model_dump in serializer

### LINKED ISSUE #754: Bug: Field PlainSerializer/WrapSerializer ignored during serialization
Pydantic offers `PlainSerializer` ([link](https://docs.pydantic.dev/latest/concepts/serialization/#field-plain-serializer)) and `WrapSerializer ` ([link](https://docs.pydantic.dev/latest/concepts/serialization/#field-wrap-serializer)) to customize field serialization behavior.  Unfortunately pydantic-settings doesn't seem to respect these serializers when calling `CliApp.serialize`.  An example:

```python
>>> from typing import Annotated, Any
>>> from pydantic import BaseModel, BeforeValidator, PlainSerializer, WrapSerializer, SerializerFunctionWrapHandler
>>> from pydantic_settings import BaseSettings, CliApp, SettingsConfigDict
>>> 
>>> def before_validator(x: Any) -> Any:
...     print("before_validator")
...     return x * 2
... 
>>> def plain_serializer(x: Any) -> Any:
...     print("plain_serializer")
...     return x * 2
... 
>>> def wrap_serializer(x: Any, handler: SerializerFunctionWrapHandler) -> int:
...     print("wrap_serializer")
...     return x * 2
... 
>>> 
>>> class Model(BaseSettings):
...     x1: Annotated[int, BeforeValidator(before_validator), PlainSerializer(plain_serializer)]
...     x2: Annotated[int, BeforeValidator(before_validator), WrapSerializer(wrap_serializer)]
... 
>>> m = Model(x1=2, x2=10)
before_validator
before_validator
>>> CliApp.serialize(m)
['--x1', '4', '--x2', '20']
```

baumgold:
Attempts to fix #754 by using the Pydantic serializer when serializing a model via the CLI

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -1315,6 +1315,31 @@ print(CliApp.serialize(settings, dict_style='env'))
 """
 ```
 
+To use [Pydantic field serializers](https://docs.pydantic.dev/latest/concepts/serialization/#field-serializers) during CLI serialization, pass `use_serializers=True`. This is opt-in (defaults to `False`) to preserve existing behavior.
+
+```py
+from pydantic import BaseModel, field_serializer
+
+from pydantic_settings import CliApp
+
+
+class Settings(BaseModel):
+    count: int
+
+    @field_serializer('count')
+    def double_count(self, v: int) -> int:
+        return v * 2
+
+
+settings = Settings(count=3)
+
+print(CliApp.serialize(settings))
+#> ['--count', '3']
+
+print(CliApp.serialize(settings, use_serializers=True))
+#> ['--count', '6']
+```
+
 ### Mutually Exclusive Groups
 
 CLI mutually exclusive groups can be created by inheriting from the `CliMutuallyExclusiveGroup` class.
--- a/pydantic_settings/main.py
+++ b/pydantic_settings/main.py
@@ -807,6 +807,7 @@ class CliApp:
         list_style: Literal['json', 'argparse', 'lazy'] = 'json',
         dict_style: Literal['json', 'env'] = 'json',
         positionals_first: bool = False,
+        use_serializers: bool = False,
     ) -> list[str]:
         """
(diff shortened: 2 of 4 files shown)


kschwab:
+@hramezani 

@baumgold this is a good idea, thanks for the PR. We just need to make sure to maintain backwards compatibility. Can you add a bool kwarg (`use_serializers`?) to the serialize method with the default set to `False`. Then you can just make `dumped_model` an empty dictionary if flag is not set and update `dumped_field` to:

```python
dumped_field = dumped_model.get(field_name, model_default)
```

You will also need to add some tests.

baumgold:
@kschwab - I've enhanced the PR as you suggested to maintain backwards compatibility.

hramezani:
Thanks @baumgold for updating PR.

I think we still need
* @kschwab review
* Test
* Documentatin

baumgold:
Apologies for the slow follow-up on this — thanks for your patience!

Here is a summary of what has been added since the initial review:

- **Backwards compatibility**: The feature is opt-in via a `use_serializers: bool = False` parameter on `CliApp.serialize()`, so existing behaviour is completely unchanged by default.
- **Tests**: Added four new test cases covering `field_serializer` transforms with `use_serializers=True`, the default-`False` behaviour, pydantic dataclasses, and propagation through nested subfields.
- **Docs**: Added a paragraph and code example under the *Serializing Arguments* section of `docs/index.md` showing the opt-in `use_serializers=True` behaviour.

@kschwab — would you be able to take another look when you get a chance? Happy to make any further changes you suggest.

hramezani:
thanks @baumgold for updating your PR.

Could you please add a test with `BaseSettings`? or change one of the existing tests and use `BaseSettings`
Also, a test with PlainSerializer/WrapSerializer from Annotated (the original issue's pattern) would be great

PR #762, opened 2026-01-22
Allow `snake_case_conversion` with `env_prefix` for Azure Key Vault source

cstarkers:
This PR adds support for using the `snake_case_conversion` in combination with `env_prefix`  in the Azure keyvault source.

Before this change, `snake_case_conversion=True` causes `env_prefix` to be ignored.

Selected Reviewer: @hramezani

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -1158,7 +1158,7 @@ For `BaseModel` and `pydantic.dataclasses.dataclass` types, `CliApp.run` will in
 * `cli_implicit_flags=True`
 * `cli_kebab_case=True`
 
-### Asynchronous CLI Commands
+#### Asynchronous Commands
 
 Pydantic settings supports running asynchronous CLI commands via `CliApp.run` and `CliApp.run_subcommand`. With this feature, you can define async def methods within your Pydantic models (including subcommands) and have them executed just like their synchronous counterparts. Specifically:
 
@@ -1224,7 +1224,43 @@ CliApp.run(Git, cli_args=['clone', 'repo', 'dir']).model_dump() == {
 
 When executing a subcommand with an asynchronous cli_cmd, Pydantic settings automatically detects whether the current thread already has an active event loop. If so, the async command is run in a fresh thread to avoid conflicts. Otherwise, it uses asyncio.run() in the current thread. This handling ensures your asynchronous subcommands "just work" without additional manual setup.
 
-### Serializing CLI Arguments
+#### Printing Help
+
+The `print_help` and `format_help` methods are available for printing or formatting help.
+
+```python
+from pydantic_settings import BaseSettings, CliApp
+
+
+class Settings(BaseSettings, cli_prog_name='example'):
+
+    def cli_cmd(self) -> None:
+        # Will print help for the current command or subcommand instance.
+        CliApp.print_help(self)
+
+        # Will return formatted help for the current command or subcommand instance.
+        CliApp.format_help(self)
+
+
+CliApp.run(Settings, cli_args=[])
+"""
+usage: example [-h]
+
+options:
+  -h, --help  show this help message and exit
+"""
+
+# You can also print or format help on the class itself.
+print(CliApp.format_help(Settings))
+"""
+usage: example [-h]
+
+options:
+  -h, --help  show this help message and exit
+"""
+```
+
+#### Serializing Arguments
 
 An instantiated Pydantic model can be serialized into its CLI arguments using the `CliApp.serialize` method. Serialization styles can be controlled using the `list_style`, `dict_style`, and `positionals_first` flags.
 
@@ -1897,6 +1933,7 @@ parser methods that can be customised, along with their argparse counterparts (t
 * `add_argument_group_method` - (`argparse.ArgumentParser.add_argument_group`)
 * `add_parser_method` - (`argparse._SubParsersAction.add_parser`)
 * `add_subparsers_method` - (`argparse.ArgumentParser.add_subparsers`)
+* `format_help_method` - (`argparse.ArgumentParser.format_help`)
 * `formatter_class` - (`argparse.RawDescriptionHelpFormatter`)
 
 For a non-argparse parser the parser methods can be set to `None` if not supported. The CLI settings will only raise an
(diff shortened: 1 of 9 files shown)


hramezani:
Thanks @cstarkers for this PR.

@AndreuCodina Could you please review the PR?

AndreuCodina:
Well, in the official provider, this behavior is implemented with a specialization of `KeyVaultSecretManager`, but we don't have it in `pydantic-settings`. The documented example is about only loading the secrets if they have the prefix.

But... regarding `env_prefix` in `pydantic-settings`, what's the standard behavior, @hramezani?

hramezani:
> Well, in the official provider, this behavior is implemented with a specialization of `KeyVaultSecretManager`, but we don't have it in `pydantic-settings`. The documented example is about only loading the secrets if they have the prefix.
> 
> But... regarding `env_prefix` in `pydantic-settings`, what's the standard behavior, @hramezani? a) If the behavior is only loading keys in the source if they start with the prefix, then this PR shouldn't be approved. b) If the behavior is just stripping the prefix, then this PR looks good.

Thanks @AndreuCodina for the review. If we set the `env_prefix`, then `pydantic-settings` collects value from `env_prefix + field_name`. 

Here is an example with env settings source:
```python


class Settings(BaseSettings):
    model_config = SettingsConfigDict(env_prefix='FOO')

    bar: str


```
If we run the above script with `BAR=test python test.py` it will raise a validation error because it expects an env with `FOO_BAR`.


Could you explain what happens if we merge the PR with an example?

AndreuCodina:
> > Well, in the official provider, this behavior is implemented with a specialization of `KeyVaultSecretManager`, but we don't have it in `pydantic-settings`. The documented example is about only loading the secrets if they have the prefix.
> > But... regarding `env_prefix` in `pydantic-settings`, what's the standard behavior, @hramezani? a) If the behavior is only loading keys in the source if they start with the prefix, then this PR shouldn't be approved. b) If the behavior is just stripping the prefix, then this PR looks good.
> 
> Thanks @AndreuCodina for the review. If we set the `env_prefix`, then `pydantic-settings` collects value from `env_prefix + field_name`.
> 
> Here is an example with env settings source:
> 
> ```python
> from pydantic_settings import BaseSettings, SettingsConfigDict
> 
> 
> class Settings(BaseSettings):
>     model_config = SettingsConfigDict(env_prefix='FOO')
> 
>     bar: str
> 
> 
> print(Settings().model_dump())
> ```
> 
> If we run the above script with `BAR=test python test.py` it will raise a validation error because it expects an env with `FOO_BAR`.
> 
> Could you explain what happens if we merge the PR with an example?

Thanks. So, it only loads settings with that prefix.

Then, `AzureKeyVaultSettingsSource`, when `env_prefix` is provided, should **only** load secrets with that prefix.

The code of the PR **always** loads the secrets, but keeps the original `env_prefix` and only converts the rest of the secret name to snake case.

cstarkers:
Hi @AndreuCodina and @hramezani, thanks both of you for review.

> Thanks. So, it only loads settings with that prefix.
> 
> Then, `AzureKeyVaultSettingsSource`, when `env_prefix` is provided, should **only** load secrets with that prefix.
> 
> The code of the PR **always** loads the secrets, but keeps the original `env_prefix` and only converts the rest of the secret name to snake case.

That's not quite right.  If  `env_prefix` is set, but the keyvault's copy of the secret does not start with the prefix, it will not be loaded.  Here's a little test demonstrating that:
```python
   def test_snake_case_conversion_negative(self, mocker: MockerFixture) -> None:
            my_field_from_kebab_case: str

            @classmethod
                cls,
                return (
                    AzureKeyVaultSettingsSource(
                        settings_cls,
                        'https://my-resource.vault.azure.net/',
                        DefaultAzureCredential(),
                        snake_case_conversion=True,
                        env_prefix="some-prefix"
                    ),
                )

        expected_secrets = [
            type(
                '', (), {'name': f'my-field-from-kebab-case', 'enabled': True}
            ),
        ]
        expected_secret_value = 'SecretValue'

        mocker.patch(
            f'{AzureKeyVaultSettingsSource.__module__}.{SecretClient.list_properties_of_secrets.__qualname__}',
            return_value=expected_secrets,
        )
        mocker.patch(
            f'{AzureKeyVaultSettingsSource.__module__}.{SecretClient.get_secret.__qualname__}',
            return_value=KeyVaultSecret(SecretProperties(), expected_secret_value),
        )
        with pytest.raises(ValidationError):
            settings = AzureKeyVaultSettings()
```

I did decide to only apply snake_case_conversion to the field name, and not the prefix itself.  My thinking was that the prefix is going to be provided as a string to the `AzureKeyVaultSettingsSource`, so unlike the field names the prefix can just be provided in the correct case for the keyvault.  If you think that's the wrong decision and it should also convert the prefix let me know, I can change it

hramezani:
@cstarkers, please include the test in [your comment](https://github.com/pydantic/pydantic-settings/pull/762#issuecomment-3802244655)

hramezani:
> Then, AzureKeyVaultSettingsSource, when env_prefix is provided, should only load secrets with that prefix.
The code of the PR always loads the secrets, but keeps the original env_prefix and only converts the rest of the secret name to snake case.

Not sure I get your comment correctly. but the prefix should remain the same.
Would be great if you can explain what you think is the right behaviour by an example code.

cstarkers:
> Not sure I get your comment correctly. but the prefix should remain the same.
> Would be great if you can explain what you think is the right behaviour by an example code.

The current PR code makes it the responsibility of the user to ensure that their prefix is a valid azure keyvault name.

Imagine if your prefix was `some_prefix_` and you had a field named `my_field`.   In the current PR implementation, that just wouldn't work - you cannot have an Azure keyvault key named `some_prefix_my-field` (keyvault doesn't allow underscores, and the current PR implementation only converts field names, not prefixes)

The current PR works really well in my use case, becasue I'm setting the prefix in the constructor, so I can set an azure-safe prefix like this:

```python
class MySettings
    my_field: str

    @classmethod
        cls,
        az_key_vault_settings: AzureKeyVaultSettingsSource = (
            AzureKeyVaultSettingsSource(
                env_prefix="some-prefix-",
                settings_cls=settings_cls,
                url=kv_url,
                credential=DefaultAzureCredential(),
                snake_case_conversion=True,
            )
        )
        return (
            init_settings,
            env_settings,
            az_key_vault_settings,
            dotenv_settings,
        )
```

I can imagine that a different user might set a prefix like `snake_case_prefix` inside a `SettingsConfigDict` and try to use the same prefix with multiple sources (the would probably expect to use `some_prefix_my_field=foo` in an env var and `some-prefix-my-field` in a keyvault secret).  Under the current PR that doesn't work if the prefix is incompatible with Azure Keyault.  Therefore, I'm seeking feedback on whether we should convert the prefix as well.

FWIW I am weakly in favour of the current PR's behaviour - it makes it possible to use whatever prefix you like, but in some circumstances you need to override `settings_customise_sources` instead of a `SettingsConfigDict`.  Converting the prefix introduces some more awkardness in the implementation.

hramezani:
Thanks @cstarkers and @AndreuCodina

Issue #719, opened 2025-11-27
Feature Request: Per-field Secret Version Configuration for GCP Secret Manager


fede-bello:
### Problem

The GoogleSecretManagerSettingsSource currently hardcodes the secret version to 'latest'. Users cannot pin specific versions at the field level, which can be useful in some cases.

### Current behavior:
```python
def __getitem__(self, key: str) -> str | None:
    # ...
    self._loaded_secrets[key] = self._secret_client.access_secret_version(
        name=self._secret_version_path(key)  # Always uses 'latest'
    ).payload.data.decode('UTF-8')
```

### Proposed Solution
Add per-field version specification using `typing.Annotated`:
```python
from typing import Annotated
from pydantic import Field
from pydantic_settings import BaseSettings

class Settings(BaseSettings):
    database_password: Annotated[str, Field(json_schema_extra={"gcp_version": "3"})]
    service_token: Annotated[str, Field(json_schema_extra={"gcp_version": "latest"})]
```

Modify `__getitem__` to extract version from field metadata:

```python
def __getitem__(self, key: str) -> str | None:
    # ...
    version = self._get_field_version(key) or 'latest'
    self._loaded_secrets[key] = self._secret_client.access_secret_version(
        name=self._secret_version_path(key, version)
    ).payload.data.decode('UTF-8')

def _get_field_version(self, field_name: str) -> str | None:
    """Extract GCP secret version from field metadata."""
    # Parse json_schema_extra or annotations
    # Returns None if not specified (defaults to 'latest')
```

This maintains backward compatibility while enabling flexible version control per field.

hramezani:
Thanks @fede-bello, for this feature request.

Good idea. do you have any reason for suggesting using  `json_schema_extra`? what about a new annotation like `SecretVersion`?

```
database_password: Annotated[str, SecretVersion('1')]
```

ezwiefel:
FWIW - I like the idea of the annotated `SecretVersion` - that could be reusable across Azure, GCP, AWS, etc.

PR #763, opened 2026-01-23
feat(gcp): support SecretVersion annotation for per-field secret versioning

ezwiefel:
- Enable per-field versioning using `SecretVersion` annotation.
- Implement logic to prevent fallback to 'latest' when a specific version is requested but missing.
- Add comprehensive tests for versioning, case sensitivity, and duplicate aliases.
- Update documentation to reflect new functionality and `populate_by_name` requirement.

Resolves: #719

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -1158,7 +1158,7 @@ For `BaseModel` and `pydantic.dataclasses.dataclass` types, `CliApp.run` will in
 * `cli_implicit_flags=True`
 * `cli_kebab_case=True`
 
-### Asynchronous CLI Commands
+#### Asynchronous Commands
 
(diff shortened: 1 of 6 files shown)


hramezani:
Thanks @ezwiefel for the PR. 

There are some comments from Copilot. Please address them if they make sense

ezwiefel:
@hramezani - I made a few of the tweaks suggested... The CI tests are failing randomly after 4 seconds - first it was py3.12 on windows next it was py3.10 on windows... (At 4 seconds, I'm assuming that it's a actions issue not a code issue.)

hramezani:
Thanks @ezwiefel

Issue #766, opened 2026-01-28
Weird behavior between custom settings source and EnvSettingsSource


RonMcKay:
# The problem

While trying to setup a pydantic settings object with custom config sources I noticed something very odd. Running the minimal example below will give be the following result:

```bash
{
  "my_field": "custom-default"
}
```

What I don't get is:
1. Even though `MY_FIELD` is given in the `.env` file, the field validates to the default value
2. Debugging into `settings_customise_sources` and executing `env_settings()`, I saw that it returns `{'custom_config': 'my-field-value'}`. How does the `EnvSettingsSource` come to grab the value from the `.env` file and use it as the value for the `custom_config` key? This makes no sense. Even more strange is, that running `dotenv_settings()` in `settings_customise_sources` will return `{'custom_config': 'my-field-value'}`. So `EnvSettingsSource` is somehow interfering with the dotenv setup and replacing environment variables.

Am I missing something here or is there a bug around?

Edit: Reversing the order in the `validation_alias` returns the right, but why does the order matter? The type hint of `AliasChoices` clearly states, that the `first_choice` can be either `str` or `AliasPath`.

# Minimal failing example

`.env` File:
```
MY_FIELD="my-field-value"
```

Python setup:
```python
from typing import Any

from pydantic import AliasChoices, AliasPath, Field
from pydantic_settings import (
    BaseSettings,
    DotEnvSettingsSource,
    EnvSettingsSource,
    InitSettingsSource,
    PydanticBaseSettingsSource,
    SecretsSettingsSource,
    SettingsConfigDict,
)


class CustomConfigSource(PydanticBaseSettingsSource):
    """Custom configuration source example."""

    def get_field_value(self, field, field_name):
        return None, "", False

    def __call__(self) -> dict[str, Any]:
        return {"custom_config": {"my-custom-key": "my-custom-value"}}


class CustomBaseSettings(BaseSettings):
    """BaseSettings subclass that adds CustomConfigSource for demonstration."""

    model_config = SettingsConfigDict(
        extra="ignore",
        env_file=".env",
    )

    @classmethod
    def settings_customise_sources(
        cls,
        settings_cls: type,
        init_settings: InitSettingsSource,
        env_settings: EnvSettingsSource,
        dotenv_settings: DotEnvSettingsSource,
        file_secret_settings: SecretsSettingsSource,
    ):
        return (
            init_settings,
            CustomConfigSource(settings_cls),
            env_settings,
            dotenv_settings,
[...]

PR #767, opened 2026-01-29
fix: Only override preferred_key when no value was found

chbndrhnns:
When using AliasChoices with AliasPath as the first choice, the env source incorrectly used the first element of the AliasPath as the key in the returned dict, even when a different alias string actually matched.

Fixes https://github.com/pydantic/pydantic-settings/issues/766


hramezani:
Thanks @chbndrhnns

PR #776, opened 2026-02-13
Support reading .env files from FIFOs (e.g. 1Password Environments)

### LINKED ISSUE #771: FIFO Support for python-dotenv provider
In v1.2.1 `python-dotenv` added support for [reading .env from FIFOs (Unix)](https://github.com/theskumar/python-dotenv/releases/tag/v1.2.1). This feature was introduced to support reading `.env` files from 1Password Environments. I would like to leverage these feature with `pydantic-settings`, but there is an explicit check that is failing to allow this operation. 

https://github.com/pydantic/pydantic-settings/blob/0f8f951b89b137bd563ce74dac016ba2c4cdc7a2/pydantic_settings/sources/providers/dotenv.py#L102

I would like to make small change to allow for fifo to be passed to python-dotenv as well `if env_path.is_file()or env_path.is_fifo():`. 

Opening a discussion here before submitting a PR.

JacobHayes:
1Password Environments mounts `.env` files as named pipes (FIFOs). These were previously skipped because the `is_file()` check returns `False` for FIFOs. This adds an `is_fifo()` check so they are read normally.

https://developer.1password.com/docs/environments

Closes #771

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -663,6 +663,11 @@ validating them.
     If a filename is specified for `env_file`, Pydantic will only check the current working directory and
     won't check any parent directories for the `.env` file.
 
+!!! tip
+    Named pipes (FIFOs) are also supported as dotenv files. This is useful for tools like
+    [1Password Environments](https://developer.1password.com/docs/environments), which mount `.env` files as
+    named pipes to provide secrets on demand without writing them to disk.
+
 Even when using a dotenv file, *pydantic* will still read environment variables as well as the dotenv file,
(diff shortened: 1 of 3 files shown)


hramezani:
Thasnk @JacobHayes for the PR.

Would be great if you can udpate the dotenv source documentation in https://github.com/pydantic/pydantic-settings/blob/main/docs/index.md#dotenv-env-support

hramezani:
Thanks @JacobHayes

Issue #800, opened 2026-03-06
Behavioural change in env_prefix behaviour compared with v1


CaselIT:
I'm trying to update an application from pydantic v1 to v2 and I'm having difficulties with basic stuff.

I have multiple settings classes that load from different prefixes environment variables, either in the environment or in dot-env files.

Working example in v1:
```py
from pathlib import Path
import os
from pydantic import BaseSettings
class Example(BaseSettings):
    class Config:
        env_prefix = "PREFIX_"
        env_file = ".foo"
        allow_mutation = False
        extra='forbid'

    foo: str = 'a'
    bar: str = 'b'
    baz: str = 'c'

env = """
PREFIX_foo=dot
STUFF=x
baz=ko
DEBUG=false
"""
Path('.foo').write_text(env)
os.environ['xy'] = 'xy'
os.environ['bar'] = 'other'
os.environ['PREFIX_bar'] = 'env'
print(Example())
```
> foo='dot' bar='env' baz='c'

as expected foo comes from the file, bar from the env, baz is left at default

The v2 version is 
```py
from pathlib import Path
import os

from pydantic_settings import BaseSettings

class Example(BaseSettings):
    model_config = {'env_prefix': 'PREFIX_', 'env_file': '.foo'}
    
    foo: str = 'a'
    bar: str = 'b'
    baz: str = 'c'

env = """
PREFIX_foo=dot
STUFF=x
baz=ko
DEBUG=false
"""
Path('.foo').write_text(env)
os.environ['xy'] = 'xy'
os.environ['bar'] = 'other'
os.environ['PREFIX_bar'] = 'env'
print(Example())
```
this errors with
```
pydantic_core._pydantic_core.ValidationError: 2 validation errors for Example
stuff
  Extra inputs are not permitted [type=extra_forbidden, input_value='x', input_type=str]
    For further information visit https://errors.pydantic.dev/2.12/v/extra_forbidden
debug
  Extra inputs are not permitted [type=extra_forbidden, input_value='false', input_type=str]
    For further information visit https://errors.pydantic.dev/2.12/v/extra_forbidden
```

I need to add `'extra': 'ignore'` for it to work, but this seems like a bug to me. I would expect that the keys that don't match the prefix from the dotevn file are filtered out, like it happens from the environment variable.

What am I doing wrong?

I'm on windows, with python 3.12, but I don't think it os/python dependant. Pydantic versions are
pydantic                  2.12.5
pydantic_core             2.41.5
[...]

PR #832, opened 2026-04-01
Allow customisation of the dotevn setting source to filter variables

CaselIT:
This PR adds a new setting option for the dotenv setting source to filter what variables are provided to the model.

The setting `env_filtering` is added:
- the default of None keeps the current behaviour
- `'match_prefix'`: only the variable that match the prefix will be passed to the model. Useful when used in conjunction with `env_prefix` to "scope" a single dotenv file to a specific model.
- `'only_existing'`: only the variables that have a corresponding field will be passed to the model. When using this option the dotenv setting source will behave like the env settings source.

Fixes: #800

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -323,6 +323,10 @@ If you have multiple variables with the same structure they will be merged.
     Sub model has to inherit from `pydantic.BaseModel`, Otherwise `pydantic-settings` will initialize sub model,
     collects values for sub model fields separately, and you may get unexpected results.
 
+!!! note
+    The `env_nested_delimiter` option applies only to variables that point to declared fields.
+    If a variable points to an unknown field no conversion will take place.
+
 As an example, given the following environment variables:
 ```bash
 # your environment


hramezani:
Thanks @CaselIT for the PR.

Copilot left some comments. Please check them.

CaselIT:
the handling of env_nested_delimiter was already kinda problematic before this patch. I've fixed the reported comments, but I haven't fixed what I would argue is another bug. Let me know if you agree, otherwise I can look into fixing it too

CaselIT:
also I haven't found if you have a changelog or similar, if you want me to add a note let me know where should I

hramezani:
> also I haven't found if you have a changelog or similar, if you want me to add a note let me know where should I

we don't have a changelog file. We will generate it during release.

hramezani:
Could you please add a test for `match_prefix + extra='forbid' + nested delimiter` scenario?

CaselIT:
updated. let me know if you want me to do something regarding the other behaviour mentioned here https://github.com/pydantic/pydantic-settings/pull/832#discussion_r3022172986

hramezani:
> updated. let me know if you want me to do something regarding the other behaviour mentioned here [#832 (comment)](https://github.com/pydantic/pydantic-settings/pull/832#discussion_r3022172986)

It would be great if you could fix this. Better to fix it in separate PR

CaselIT:
I'll create an issue in the meantime. just to confirm, this should be `exp = {'a': {'x': 'bar', 'b': 'y'}, 'b': 'b', 'c': {'a': '1', 'b': '2'}}` independently from this PR, right?

hramezani:
> I'll create an issue in the meantime. just to confirm, this should be `exp = {'a': {'x': 'bar', 'b': 'y'}, 'b': 'b', 'c': {'a': '1', 'b': '2'}}` independently from this PR, right?

After checking more, I think this probably is not an issue. becuase `c` is not defined as a field. The merge is happening for `a` becase it is defined as a field in the model.

CaselIT:
You are probably right, the docs says 

> Sub model has to inherit from pydantic.BaseModel, Otherwise pydantic-settings will initialize sub model, collects values for sub model fields separately, and you may get unexpected results.

since there is no c BaseModel this is likely in line with "unexpected results"

Maybe an additional doc line may be worth adding?

hramezani:
> You are probably right, the docs says
> 
> > Sub model has to inherit from pydantic.BaseModel, Otherwise pydantic-settings will initialize sub model, collects values for sub model fields separately, and you may get unexpected results.
> 
> since there is no c BaseModel this is likely in line with "unexpected results"
> 
> Maybe an additional doc line may be worth adding?

yes, would be good

CaselIT:
done

hramezani:
Thanks @CaselIT 

Let's merge this. It will be available in the next release!

Please help us in case of issues after the release.

CaselIT:
great!

> Please help us in case of issues after the release.

will try, ping me if something comes up related to this

PR #692, opened 2025-10-12
Strip non-explicit default values.

kschwab:
Fixes #678

PR #852, opened 2026-04-28
fix: CliApp.run nested model_fields_set pollution with partial update

### LINKED ISSUE #851: CliApp.run still pollutes nested model_fields_set despite #692 fix
## Description

PR #692 fixed `nested_model_default_partial_update=True` polluting `model_fields_set` for direct `BaseSettings()` construction. However, the same bug persists when using `CliApp.run()` with CLI arguments.

When a CLI argument targets a single field in a nested model (e.g. `--training.num_steps 2`), **all** fields of that nested model end up in `model_fields_set` — not just the one the user actually provided.

## Reproduction

```python
import sys
from pydantic_settings import BaseSettings, SettingsConfigDict, CliApp

class NestedModel(BaseModel):
    field_a: str = "default_a"
    field_b: int = 42

class Settings(BaseSettings):
        nested_model_default_partial_update=True,
        cli_parse_args=False,
    )
    nested: NestedModel = NestedModel()

    def cli_cmd(self):
        print("fields_set:", self.nested.__pydantic_fields_set__)

sys.argv = ["test", "--nested.field_a", "override"]
CliApp.run(Settings)
```

**Actual output:**
```
fields_set: {'field_b', 'field_a'}
```

**Expected output:**
```
fields_set: {'field_a'}
```

## Root cause

`CliApp.run` → `_settings_build_values` merges CLI source with defaults via `deep_update`. The #692 strip logic (`{key: val for key, val in state.items() if key not in defaults or defaults[key] != val}`) only does shallow top-level comparison. For nested dicts, the entire nested dict differs from the default (because one field changed), so the whole dict — with all sub-fields — is kept. Pydantic then constructs the nested model with all fields passed → all end up in `model_fields_set`.

## Impact

Any library that checks `model_fields_set` to determine user intent (e.g. `reconcile`'s `@dependency` skip logic) breaks — derived fields that should be auto-computed are incorrectly treated as user-provided.

## Proposed fix

After constructing the model in `CliApp.run`, walk nested models and remove fields from `model_fields_set` that weren't present in the `CliSettingsSource` output. The CLI source already returns exactly the user-provided fields (`{'nested': {'field_a': 'override'}}`), so it's the correct ground truth.

See linked PR for implementation.

junjzhang:
Fixes #851

## Problem

When `nested_model_default_partial_update=True`, `CliApp.run()` marks **all** fields of a touched nested model as user-set in `model_fields_set`, even those the CLI didn't override. This is a gap in the #692 fix, which only handled the shallow top-level case.

**Root cause:** `_settings_build_values` merges CLI overrides with defaults via `deep_update`. For nested dicts, even a single changed sub-field makes the entire dict differ from the default, so the #692 strip logic keeps it wholesale. Pydantic then constructs the nested model with all fields passed → all end up in `model_fields_set`.

## Fix

After constructing the model in `CliApp.run`, walk nested models and trim `model_fields_set` using the `CliSettingsSource` output as ground truth — it returns exactly the fields the user provided.

```python
def _fix_nested_fields_set(model, cli_state):
    for field_name in type(model).model_fields:
        child = getattr(model, field_name, None)
        if not isinstance(child, BaseModel):
            continue
        if field_name in cli_state and isinstance(cli_state[field_name], dict):
            child.model_fields_set.intersection_update(cli_state[field_name].keys())
            _fix_nested_fields_set(child, cli_state[field_name])
        elif field_name not in cli_state:
            child.model_fields_set.clear()
```

## Tests

Two new tests:
- `test_cli_nested_model_default_partial_update_fields_set`: verifies `model_fields_set` contains only CLI-provided fields
- `test_cli_nested_partial_update_exclude_unset`: verifies `model_dump(exclude_unset=True)` only includes CLI-overridden nested fields


hramezani:
Thanks for the PR! The issue is valid, but I think the fix would be more robust if we address the root cause rather than applying a post-hoc mutation.

A few concerns with the current approach:

1. **Post-hoc mutation of `model_fields_set`** — Directly mutating pydantic's internal tracking set after construction is fragile and could break with future pydantic versions.

2. **Only applies to `CliApp.run()`** — If someone constructs a `BaseSettings` instance directly (not via `CliApp.run`) with a CLI source and `nested_model_default_partial_update=True`, the bug would still exist.

3. **Only considers `CliSettingsSource` as ground truth** — If multiple sources contribute to nested fields (e.g., an env var sets `nested.field_b` and CLI sets `nested.field_a`), the fix would incorrectly trim `field_b` from `model_fields_set` since it only looks at CLI state.

4. **Re-invokes the CLI source** — `cli_settings_source()` is called a second time to get the state, which could have side effects or performance implications.

### Suggested approach

Instead, I think we should fix the root cause in `_settings_build_values`. The strip logic on line 479:

```python
state = {key: val for key, val in state.items() if key not in defaults or defaults[key] != val}
```

only does a shallow top-level comparison. We could make this recursive — doing a deep comparison that strips matching default values at every nesting level. This would:

- Fix the problem at its source for **all** construction paths (not just `CliApp.run`)
- Correctly handle **multi-source** scenarios (env + CLI + config files)
- Avoid mutating `model_fields_set` after construction
- Let pydantic natively track `model_fields_set` correctly

junjzhang:
Thanks for the thorough review — all valid points. I've reworked the fix to address the root cause in `_settings_build_values` as you suggested.

### What changed

Replaced the shallow top-level strip:

```python
state = {key: val for key, val in state.items() if key not in defaults or defaults[key] != val}
```

with a recursive `_deep_strip_defaults(state, defaults, cls)` that walks nested dicts and strips individual values matching defaults. This:

- Fixes the problem at the source for **all construction paths**, not just `CliApp.run`
- Correctly handles **multi-source** scenarios (env + CLI + config files)
- **No post-hoc mutation** of `model_fields_set`
- **No re-invocation** of `cli_settings_source()`

### Handling required fields

One subtlety: for nested `BaseModel` types with required fields (no field-level defaults), we can't strip matching values from the init dict — pydantic needs them for construction. In those cases the function preserves them, maintaining existing behavior. The deep strip fully applies to nested `BaseModel` types where all fields have defaults (which is the common case for `nested_model_default_partial_update`).

For non-`BaseModel` nested types (pydantic/stdlib dataclasses), the function falls back to shallow comparison since we can't inspect their field defaults through the same API.

Updated `test_class_nested_model_default_partial_update` to reflect the corrected `exclude_unset` behavior for `NestedD` (BaseModel with all-optional fields).

hramezani:
Nice work on the rework! The recursive approach is much better.

One thing I noticed — it looks like the deep strip only handles one level of nesting. I tried this with a 3-level deep model and the second nested level still has the issue:

```python
from pydantic_settings import BaseSettings, SettingsConfigDict, CliApp

class DeepInner(BaseModel):
    x: int = 1
    y: int = 2

class Middle(BaseModel):
    deep: DeepInner = DeepInner()
    z: str = "hello"

class Settings(BaseSettings):
        nested_model_default_partial_update=True,
        cli_parse_args=False,
    )
    middle: Middle = Middle()

    def cli_cmd(self):
        pass

model = CliApp.run(Settings, cli_args=["--middle.deep.x", "99"])
print(model.middle.model_fields_set)       # {'deep'} ✓
print(model.middle.deep.model_fields_set)  # {'x', 'y'} — expected {'x'}
```

The first level is correctly stripped, but `_deep_strip_defaults` doesn't recurse into the inner nested dict. Could you make it fully recursive so it handles arbitrary nesting depth?

Issue #650, opened 2025-07-03
Allow setting env_nested_delimiter for CliSettingsSource


tmchartrand:
The delimiter is hardcoded to '.' for CliSettingsSource, but in some cases it would be helpful to allow the user to adjust that default!

https://github.com/pydantic/pydantic-settings/blob/9380bc653d0d5b712ea85230608854d898947ec8/pydantic_settings/sources/providers/cli.py#L234C13-L234C33

hramezani:
@kschwab what do you think?

kschwab:
Hi @tmchartrand, is there a strong use case for this? I think waiting to see if there is more demand for this feature might make more sense.

tmchartrand:
I imagine it's not especially common, but we work with a no-code app interface that passes parameters to python via command line args, and for some strange reason has character restrictions that exclude "." in parameter names. Using CliSettingsSource would otherwise be super convenient for parsing arguments, but nested parameters are currently impossible.

This seems like a very small and non-consequential change, I could try to submit a PR if it would be welcome.

jakubman1:
I can see a use-case for this in combination with the `CliMutuallyExclusiveGroup` class.

Suppose you have a mutually exclusive group for batch operations:

```py
class BatchOps(CliMutuallyExclusiveGroup):
    add: CliImplicitFlag[bool] = Field(False)
    delete: CliImplicitFlag[bool] = Field(False)
    set: CliImplicitFlag[bool] = Field(False)

class EditCmd(BaseModel):
    batch: BatchOpts
```

If you could set the delimiter to a dash, you can essentially "hide" the fact that it is a nested object and use `--batch-add` while taking an advantage of the `CliMutuallyExclusiveGroup`. If I want to achieve something like this right now, I have to specify the fields directly in the EditCmd class as `batch_add` and check for the mutual exclusivity in my `cli_cmd` function.

Issue #406, opened 2024-09-16
Support dash-separated CLI arguments and underscore-separated ENV variables


andlogreg:
When using multi-word named settings, I would like to:

1. for CLI, use dashes for word separation, as it is the most common approach I've seen in CLI programs
2. for ENV variables, use underscores for word separation as it is the standard (and dashes are not supported anyway)

Regarding 2., is given for "free" if you name the fields of your python models in snake case, as it is typically done.

For achieving 1., I tried something like this (see alias generator) which works:

```python
# test_config.py
from pydantic import Field
from pydantic_settings import BaseSettings, SettingsConfigDict
from pydantic import Field
from typing import List

class MySettings(BaseSettings):

    my_list: List[str] = Field(
        default_factory=list, description="A list argument"
    )

    my_int: int = Field(
        0,
        description="An int argument",
    )

    model_config = SettingsConfigDict(
        cli_parse_args=True,
        alias_generator=lambda x: x.replace("_","-")
    )



settings = MySettings()

print(settings)
```

And allows me to `python test_config.py --my-list=a,b,c --my-int=45`.

But this breaks the ENV variables functionality, since the names are also expected to have `-`, which is not supported afaik.
I understand this happens because aliases are also used to define the expected ENV variable names.

**Is there currently a way to achieve 1. and 2. simultaneously?** 

Maybe using different aliases for CLI arguments and ENV variables? 
Or using a different approach (ie, not aliases) to use `-` as the work separator in CLI?

hramezani:
Thanks @andlogreg for this issue.

I think you can use different aliases for env and CLI. You can use `AliasChoices` and define multiple aliases for one field. one for ENV and one for CLI.


take a look at https://docs.pydantic.dev/latest/concepts/pydantic_settings/#literals-and-enums

andlogreg:
> Thanks @andlogreg for this issue.
> 
> I think you can use different aliases for env and CLI. You can use `AliasChoices` and define multiple aliases for one field. one for ENV and one for CLI.
> 
> take a look at https://docs.pydantic.dev/latest/concepts/pydantic_settings/#literals-and-enums

Awesome! Indeed this seems to do the trick:

```python
(...)
class MySettings(BaseSettings):

    my_list: List[str] = Field(
        default_factory=list, description="A list argument",
        alias=AliasChoices("my_list", "my-list"),
    )

    my_int: int = Field(
        0,
        description="An int argument",
[...]

PR #860, opened 2026-05-10
Add support for showing environment variable names in CLI help text

brad-alexander:
## Motivation

`CliSettingsSource` can already derive CLI help from a `BaseSettings` model, but the help text does not show the
environment variables that can populate each field. This makes it harder to use the generated CLI help as an operations
or configuration-discovery reference for applications that support both command-line flags and environment variables.

I searched the upstream issue tracker for CLI help and environment-variable requests. I did not find an exact existing
issue for displaying env vars in `--help`; related discussions include env/alias behavior and CLI/env naming
interactions, such as #406 and #650.

## Behavior

This adds an opt-in `cli_show_env_vars` setting. It is disabled by default, so existing help output is unchanged unless
users explicitly enable it via `SettingsConfigDict`, `CliSettingsSource`, `BaseSettings` init kwargs, or `CliApp.run`.

When enabled, CLI help appends resolved env vars to each supported argument:

```text
usage: example.py [-h] [--foo str]

options:
  --foo str   Foo value (required) [env: MYAPP_FOO]
```

Multiple aliases are rendered in resolution order:

```text
--token str  (required) [env: TOKEN | LEGACY_TOKEN]
```

Nested models honor `env_nested_delimiter`:

```text
--database.url str  (required) [env: MYAPP_DATABASE__URL]
```

The implementation reuses `EnvSettingsSource._extract_field_info` for env-name resolution and does not change runtime
environment-variable loading behavior.

## Tests

Added coverage for:

- Default-off help output.
- `env_prefix` rendering.
- `AliasChoices` rendering.
- Nested `BaseModel` fields with `env_nested_delimiter`.
- Alias-only fields and positional argument skips.
- Case-sensitive and case-insensitive display.
- `CliSubCommand` help.
- `CliApp.run(..., cli_show_env_vars=True)` runtime override.

Focused test command:

```bash
uv run --group testing pytest tests/test_source_cli.py -k "cli_show_env_vars"
```


hramezani (commented):
Thanks for the PR. The feature is useful — showing env var names in CLI help is a common ask for 12-factor apps.

I tested the implementation locally against many scenarios (deep nesting, `AliasChoices`, `env_prefix_target`, discriminated unions, subcommands, `CliMutuallyExclusiveGroup`, `SecretStr`, `Optional`, kebab-case, implicit flags, etc.) and most work correctly. Nice job on threading `env_prefixes` through recursive calls.

A few things I'd like addressed:

---

### 1. Mutating shared state in `_get_field_env_var_names`

```python
original_env_prefix = self._env_settings_source.env_prefix
try:
    for env_prefix, force_env_prefix in env_prefixes:
        self._env_settings_source.env_prefix = env_prefix
        ...
finally:
    self._env_settings_source.env_prefix = original_env_prefix
```

Temporarily mutating `env_prefix` on the shared `_env_settings_source` and restoring it in a `finally` is fragile. If `_env_settings_source` is the **actual** `EnvSettingsSource` passed via `_env_settings_source=env_settings` from `_settings_init_sources`, you're mutating the source that will later be used for real env loading. Consider either:
- Working with a copy/snapshot of the prefix rather than mutating the source, or
- Passing the prefix as a parameter to `_extract_field_info` instead of monkey-patching the instance

---

### 2. Fallback `EnvSettingsSource` may diverge from actual config

In `CliSettingsSource.__init__` (cli.py line ~443):

```python
self._env_settings_source = (_env_settings_source if self.cli_show_env_vars else None) or (
    EnvSettingsSource(settings_cls) if self.cli_show_env_vars else None
)
```

When `_env_settings_source` is not passed (standalone `CliSettingsSource` construction), the fallback creates a bare `EnvSettingsSource(settings_cls)`. This picks up `model_config` values, so it's fine for most cases. But if a user customizes `EnvSettingsSource` at construction time (e.g. passing a different `env_prefix` or `env_nested_delimiter` to the source directly), the fallback won't reflect that.

Not a blocker, but worth documenting this limitation.

---

### 3. Missing test coverage

The tests cover the core scenarios well, but I'd like to see tests for:

- **Deep nesting (3+ levels)** — the current nested test only goes 1 level (`Database.url`). The implementation handles it correctly, but a test would prevent regressions.
- **`env_prefix_target`** — `'alias'`, `'variable'`, `'all'` all affect which env vars get the prefix. No test coverage for this.
- **Discriminated unions** — works correctly but untested.
- **`run_subcommand` with `cli_show_env_vars`** — the signature was updated but there's no test exercising that path.

---

### 4. Minor: docs example uses `sys.argv` manipulation

The docs example does:
```python
sys.argv = ['example.py', '--help']
Settings()
```

This is the existing pattern in the docs, but since `CliApp.format_help` exists and is cleaner, consider using it in the new example for consistency with the test style.

---

Overall this is solid work. The env prefix threading through recursive parser construction is well-designed, and the `env_var_names` property is a nice programmatic API.

brad-alexander:
Thanks for the thorough review, @hramezani , really appreciate the time you spent putting it through its paces. I pushed updates for each requested item:

1. **Avoid mutating the shared env source**

   I changed `_get_field_env_var_names` to work with a shallow copy of the configured env settings source before changing `env_prefix`, so the shared `_env_settings_source` is not monkey-patched during help generation:

   - [`pydantic_settings/sources/providers/cli.py` lines 1502-1514](https://github.com/brad-alexander/pydantic-settings/blob/cli-show-env-vars/pydantic_settings/sources/providers/cli.py#L1502-L1514)

2. **Document fallback `EnvSettingsSource` behavior**

   I added docs noting that direct `CliSettingsSource(..., cli_show_env_vars=True)` construction falls back to `EnvSettingsSource(settings_cls)` when `_env_settings_source` is not passed, and that source-level constructor overrides need to be passed explicitly:

   - [`docs/index.md` lines 1820-1823](https://github.com/brad-alexander/pydantic-settings/blob/cli-show-env-vars/docs/index.md#L1820-L1823)
   - Related fallback construction: [`pydantic_settings/sources/providers/cli.py` lines 443-445](https://github.com/brad-alexander/pydantic-settings/blob/cli-show-env-vars/pydantic_settings/sources/providers/cli.py#L443-L445)

3. **Add missing test coverage**

   I added tests for the scenarios you called out:

   - Deep nesting, 3 levels: [`tests/test_source_cli.py` lines 656-676](https://github.com/brad-alexander/pydantic-settings/blob/cli-show-env-vars/tests/test_source_cli.py#L656-L676)
   - `env_prefix_target='alias'`, `'variable'`, and `'all'`: [`tests/test_source_cli.py` lines 679-702](https://github.com/brad-alexander/pydantic-settings/blob/cli-show-env-vars/tests/test_source_cli.py#L679-L702)
   - Discriminated unions: [`tests/test_source_cli.py` lines 705-728](https://github.com/brad-alexander/pydantic-settings/blob/cli-show-env-vars/tests/test_source_cli.py#L705-L728)
   - `run_subcommand` with `cli_show_env_vars`: [`tests/test_source_cli.py` lines 814-835](https://github.com/brad-alexander/pydantic-settings/blob/cli-show-env-vars/tests/test_source_cli.py#L814-L835)

4. **Use `CliApp.format_help` in the docs example**

   I updated the new docs example to use `CliApp.format_help(Settings)` instead of mutating `sys.argv`:

   - [`docs/index.md` lines 1789-1813](https://github.com/brad-alexander/pydantic-settings/blob/cli-show-env-vars/docs/index.md?plain=1#L1789-L1813)

I have not worked much with `env_prefix_target` or discriminated unions before, so I would appreciate another look at those two test cases in particular. Please let me know if either one is off the mark and I can adjust it.

As a note:  it seems like without the [latest commit](https://github.com/pydantic/pydantic-settings/pull/860/changes/f00582905c34403381782c7e7be25e3ee051e53a) adding `cli_prog_name='example.py',`, the docs test failed only on  `test py3.14 on ubuntu-latest` and `test py3.14 on macos-latest` but curiously not on `testpy3.14 on windows-latest` or any previous python version.  Flagging in case this is unexpected.

hramezani:
Thanks for the update, @brad-alexander! All four items from my review look good now.

Could you also add tests for a few more edge cases?

- `CliMutuallyExclusiveGroup` with `cli_show_env_vars`
- `SecretStr` fields
- `Optional` fields
- Kebab-case (`cli_kebab_case=True`)

brad-alexander:
Appreciate the review again @hramezani!  I've added the additional tests requested:

- `CliMutuallyExclusiveGroup` with `cli_show_env_vars`.<sup><a href="https://github.com/brad-alexander/pydantic-settings/blob/2a69587c53b47376db117b8aef98458d80223eb4/tests/test_source_cli.py#L623-L650">link</a></sup>
- `SecretStr` fields <sup><a href="https://github.com/brad-alexander/pydantic-settings/blob/2a69587c53b47376db117b8aef98458d80223eb4/tests/test_source_cli.py#L653-L662">link</a></sup>
- `Optional` fields <sup><a href="https://github.com/brad-alexander/pydantic-settings/blob/2a69587c53b47376db117b8aef98458d80223eb4/tests/test_source_cli.py#L665-L674">link</a></sup>
- Kebab-case (`cli_kebab_case=True`) <sup><a href="https://github.com/brad-alexander/pydantic-settings/blob/2a69587c53b47376db117b8aef98458d80223eb4/tests/test_source_cli.py#L677-L691">link</a></sup>

pydantic_settings/sources/providers/cli.py
```diff
@@ -432,6 +440,10 @@ def __init__(
         self.cli_shortcuts = (
             cli_shortcuts if cli_shortcuts is not None else settings_cls.model_config.get('cli_shortcuts', None)
         )
+        self._env_settings_source = (_env_settings_source if self.cli_show_env_vars else None) or (
+            EnvSettingsSource(settings_cls) if self.cli_show_env_vars else None
+        )
```
hramezani: Minor: this `(x if flag else None) or (... if flag else None)` is correct but a bit hard to parse. A plain `if/else` reads more directly:

```suggestion
        if self.cli_show_env_vars:
            self._env_settings_source = _env_settings_source or EnvSettingsSource(settings_cls)
        else:
            self._env_settings_source = None
```

pydantic_settings/sources/providers/cli.py
```diff
@@ -432,6 +440,11 @@ def __init__(
         self.cli_shortcuts = (
             cli_shortcuts if cli_shortcuts is not None else settings_cls.model_config.get('cli_shortcuts', None)
         )
+        if self.cli_show_env_vars:
+            self._env_settings_source = _env_settings_source or EnvSettingsSource(settings_cls)
+        else:
+            self._env_settings_source = None
```
hramezani: The mypy failure comes from this `if/else`: mypy infers the attribute type from the first branch (`EnvSettingsSource`), so the `else: ... = None` is rejected (`cli.py:446: Incompatible types in assignment`). Declaring the type up front fixes it and keeps it readable:

```suggestion
        self._env_settings_source: EnvSettingsSource | None = None
        if self.cli_show_env_vars:
            self._env_settings_source = _env_settings_source or EnvSettingsSource(settings_cls)
```

Verified locally: mypy, `ruff check`, and `ruff format --check` all pass with this. Sorry for the churn from my earlier suggestion!

hramezani:
Thanks @brad-alexander

PR #862, opened 2026-05-12
Skip list_secrets call when case_sensitive=True

### LINKED ISSUE #861: GoogleSecretManagerSettingsSource: skip the list secrets step when case_sensitive=True
Currently GoogleSecretManagerSettingsSource will unconditionally `list_secrets()` which requires `secretmanager.secrets.list`. 

As far as I understand the only purpose of this list_secrets is to support the case ignore matching of the secret names. 

In principle, with the current code both IAM roles [Secret Manager Secret Accessor 
(roles/secretmanager.secretAccessor)](https://docs.cloud.google.com/iam/docs/roles-permissions/secretmanager#secretmanager.secretAccessor) + [Secret Manager Viewer 
(roles/secretmanager.viewer)](https://docs.cloud.google.com/iam/docs/roles-permissions/secretmanager#secretmanager.viewer) even when `case_sensitive=True`

If the list_secrets step is skipped then it will work with just [Secret Manager Secret Accessor 
(roles/secretmanager.secretAccessor)](https://docs.cloud.google.com/iam/docs/roles-permissions/secretmanager#secretmanager.secretAccessor).

On a related note maybe it would be nice to document that both `Secret Manager Secret Accessor` and `Secret Manager Viewer` are currently required for this to work.

ecerulm:
## Summary

Avoid `list_secrets()` when `case_sensitive=True`, lowering the required GCP IAM permissions from `roles/secretmanager.viewer` + `roles/secretmanager.secretAccessor` down to just `roles/secretmanager.secretAccessor`.

Closes #861

## Changes

- **Skip `list_secrets` when `case_sensitive=True`**: key lookup goes directly to `access_secret_version`, so no listing permission is needed.
- **Distinguish `NotFound` from `PermissionDenied`**: adds `_get_secret_value_or_raise()` which raises `KeyError` on `google.api_core.exceptions.NotFound` (secret does not exist) but returns `None` for other errors such as `PermissionDenied` (secret exists but access was denied). This preserves the correct `__getitem__` contract without requiring `list_secrets` to verify key existence.
- **Update tests**: mock `google.api_core.exceptions.NotFound` for missing secrets and `google.api_core.exceptions.PermissionDenied` for access-denied scenarios instead of using `Exception` for both.

## Permissions required

| `case_sensitive` | Permissions needed |
|---|---|
| `False` (default) | `roles/secretmanager.viewer` + `roles/secretmanager.secretAccessor` |
| `True` | `roles/secretmanager.secretAccessor` only |

Selected Reviewer: @dmontagu


tests/test_source_gcp_secret_manager.py
```diff
@@ -76,7 +77,7 @@ def mock_access_secret_version(name: str):
                 resp = mocker.Mock()
                 resp.payload.data.decode.return_value = secret_values[name]
                 return resp
-            raise Exception(f'Secret not found or access denied: {name}')
+            raise NotFound(f'Secret not found: {name}')
```
hramezani, line 80: Isn't it a breaking change?

hramezani:
Thanks @ecerulm for the PR.
Could you please add some tests for the new changes?

tests/test_source_gcp_secret_manager.py
ecerulm: No, those are the actual exceptions that the gcp actually raises. The previous test was just raising the general `Exception` now it raises the same exception it will be raised by gcloud sdk

ecerulm:
> Thanks @ecerulm for the PR. Could you please add some tests for the new changes?

I added three test cases for `caseSensitive=True` 

* `test_case_sensitive_getitem_skips_list_secrets`: checks that `list_secrets` is not called when the secret exists and we have the permissions to access it
* `test_case_sensitive_getitem_not_found_raises_keyerror`: checks that `list_secrets` is not called when the secret does not exist and check that KeyError is returned which is the same current behaviour (calling list_secrets)
* `test_case_sensitive_getitem_permission_denied_returns_none`: Checks that `list_secrets` operation is not called when the secret exists but we don't have permission to access it and also checks it returns `None` in this case which is the current behaviour.

hramezani:
Thanks @ecerulm for updating the PR.

@ezwiefel Could you please review the PR?

hramezani (approved):

hramezani:
Thanks @ecerulm

Issue #673, opened 2025-09-03
Strictness with BaseSettings


pycaw:
I fail to be able to use strict parsing with `BaseSettings`. I haven't found this behavior documented in the docs whatsoever. No mention of strict/strictness among the issues here either as far as I could see.

Given these definitions,
```py
import json as json

from pydantic import BaseModel, StrictBool, StrictInt, ValidationError
from pydantic_settings import BaseSettings

json_doc = """
{
    "mybool": 0
}
"""


class Model(BaseModel):
    mybool: bool


class Settings(BaseSettings):
    mybool: bool


class StrictSettings(BaseSettings):
    mybool: StrictBool
```

we will get `ValidationError`s with `Model` but not with `Settings`:

```py
try:
    data = Model.model_validate_json(json_doc, strict=True)
except ValidationError as e:
    assert e.error_count() == 1

data = Settings.model_validate_json(json_doc, strict=True)
```

Same with `model_validate`.
Even enabling `strict` on `model_config` doesn't have an effect
```py
Settings.model_config["strict"] = True

obj = json.loads(json_doc)

try:
    data = Model.model_validate(obj, strict=True)
except ValidationError as e:
    assert e.error_count() == 1

data = Settings.model_validate(obj, strict=True)
```

Using `Strict*` types changes things but that's not what I am looking for:
```py
try:
    data = StrictSettings.model_validate(obj)
except ValidationError as e:
    assert e.error_count() == 1
```

**My use case is that I want CLI arguments to be parsed normally and config files to be parsed strictly.**

_And I use yaml config files so that will be another problem with SettingsSource configuration but let's not get ahead of ourselves._

pycaw:
The more I think about the more it seems that `model_validate*` should not be used with `BaseSettings`. All examples construct `BaseSettings` instances by relying on the various sources. It is weird however why `model_validate*` methods sit there with options like `strict: bool` waiting to be misused. I only resorted to this way of doing it so that I can produce simplified examples. As I mentioned my use case revolves around CLI argument and config file parsing. I had gotten here from this test of mine where I tried to hack in strict parsing just for config files:
```py
from pydantic_settings import BaseSettings, JsonConfigSettingsSource, PydanticBaseSettingsSource

class Settings(BaseSettings):
    @classmethod
    def settings_customise_sources(
        cls,
        settings_cls: type[BaseSettings],
        init_settings: PydanticBaseSettingsSource,
        env_settings: PydanticBaseSettingsSource,
[...]

PR #872, opened 2026-05-26
Document strict validation behavior for BaseSettings

jameslcowan:
Relates to #673

BaseSettings loads values through settings sources during __init__, so passing strict=True to model_validate() / model_validate_json() does not behave the way many users expect.

This adds a short docs section explaining how to enable strict validation via SettingsConfigDict(strict=True) or Strict* field types, why model_validate* strict mode differs from plain BaseModel, and practical alternatives for mixed strict/lenient source requirements.

## Test plan
- [x] Docs-only change

Made with [Cursor](https://cursor.com)

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -138,6 +138,43 @@ print(Settings1())
 
 Check the [validation of default values](fields.md#validate-default-values) for more information.
 
+## Strict validation
+
+`BaseSettings` loads values from settings sources (environment variables, dotenv files, secrets, CLI, and so on). Those
+values are usually strings that must be coerced into the correct Python types.
+
+To reject invalid coercions, enable strict mode on the settings class:
+
+```py
+from pydantic_settings import BaseSettings, SettingsConfigDict
+
+
+class Settings(BaseSettings):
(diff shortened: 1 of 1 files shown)


hramezani:
Closing this — the docs change has factual errors that would mislead users.

Main issues:

1. **The core example is wrong.** `MYBOOL=0` with `strict=True` does NOT raise a validation error. It silently succeeds and returns `False`. You can verify this yourself in a shell. The reason is `_coerce_env_val_strict` pre-converts env strings using lax validation before strict mode ever sees them.

2. **Misleading about what `strict=True` actually does with env vars.** It doesn't work like standard Pydantic strict mode — most string values still get coerced just fine.

3. **Missing important detail about `model_validate`.** `SettingsConfigDict(strict=True)` actually does make `model_validate()` strict (without the parameter). The PR doesn't mention this.

4. **Issue #673 is a bug, not a docs gap.** Documenting broken behavior as if it's by design isn't the right fix here.

Please don't open PRs with AI-generated content without actually verifying that the code examples and claims are correct. Every example in a docs PR should be tested before submitting.

PR #874, opened 2026-05-28
fix: handle case-insensitive os.environ on Windows (closes #295)

### LINKED ISSUE #295: environment variables case sensitivity on windows
According to the [docs ](https://docs.pydantic.dev/latest/concepts/pydantic_settings/#case-sensitivity):

> On Windows, Python's os module always treats environment variables as case-insensitive, 
> so the case_sensitive config setting will have no effect - settings will always be updated ignoring case.

In my tests on windows 11:

- Case Sensitive: True, .env file    `works`
- Case Sensitive: False, .env file    `works`
- Case Sensitive: True, os.environ    `does NOT work`
- Case Sensitive: False, os.environ    `works`

The example given in the docs fails for me:

```python
class RedisSettings(BaseModel):
    host: str
    port: int

class Settings(BaseSettings, case_sensitive=True):
    redis: RedisSettings

os.environ["redis"] = '{"host": "localhost", "port": 6379}'
```
```
redis
    For further information visit https://errors.pydantic.dev/2.7/v/missing
```

But it works if I set `case_sensitive=False`.

botbikamordehai2-sketch:
## What
On Windows, `os.environ` is inherently case-insensitive and normalizes all keys to uppercase. When `case_sensitive=True` is configured, `parse_env_vars` preserves key case, but since `os.environ` has already uppercased the keys (e.g., `os.environ['redis']` is stored as `'REDIS'`), lookups for lowercase field names fail.

This caused `Settings(case_sensitive=True)` to fail to find environment variables on Windows, contradicting the documented behavior which states:
> On Windows, Python's os module always treats environment variables as case-insensitive, so the case_sensitive config setting will have no effect.

## Fix
Force `case_sensitive=False` when loading env vars from `os.environ` on Windows (`sys.platform == 'win32'`), since the OS itself cannot support case-sensitive environment variables. This makes the behavior match the documented expectation.

Closes #295

### DIFF
--- a/pydantic_settings/sources/providers/env.py
+++ b/pydantic_settings/sources/providers/env.py
@@ -72,7 +72,13 @@ class EnvSettingsSource(PydanticBaseEnvSettingsSource):
         self.env_vars = self._load_env_vars()
 
     def _load_env_vars(self) -> Mapping[str, str | None]:
-        return parse_env_vars(os.environ, self.case_sensitive, self.env_ignore_empty, self.env_parse_none_str)
+        # On Windows, os.environ is case-insensitive and normalizes keys to uppercase,
+        # so case_sensitive=True cannot be respected at the OS level.
+        # We force case_sensitive=False to ensure lookups work correctly on Windows.
+        import sys
+
+        case_sensitive = self.case_sensitive if sys.platform != 'win32' else False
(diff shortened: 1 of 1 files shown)


hramezani:
The PR doesn't have any change

PR #880, opened 2026-06-02
fix: move SecretsManagerClient type import under TYPE_CHECKING

### LINKED ISSUE #879: AWS Secrets Manager source unnecessarily requires types-boto3 at runtime
The AWS provider imports `SecretsManagerClient` from `types_boto3_secretsmanager.client` only for type checking purposes:
https://github.com/pydantic/pydantic-settings/blob/e20044d155ad86d047a004dbccef4a7aa6532a03/pydantic_settings/sources/providers/aws.py#L18-L33

This causes inconvenience for people (myself included) which prefer to keep type stubs as dev dependencies, because `pydantic-settings` requires to install them anyway.

Since `types-boto3[secretsmanager]` is already a part of a `linting` dependency group, a trivial change is required to move the `SecretsManagerClient` under `TYPE_CHECKING` guard and drop `types-boto3[secretsmanager]` dependency from `aws-secrets-manager` extra.

If you are interested, I'd be glad to contribute.

gavin913-lss:
## Problem

`SecretsManagerClient` from `types_boto3_secretsmanager` is imported at runtime inside `import_aws_secrets_manager()`, but it's only used for type annotations. This forces users to install `types-boto3[secretsmanager]` even when they don't need type stubs.

Fixes #879

## Fix

1. **`pydantic_settings/sources/providers/aws.py`**:
   - Move `SecretsManagerClient` import under `TYPE_CHECKING` guard
   - Remove it from the runtime import function
   - Since `from __future__ import annotations` is already present, the type annotation is lazy and doesn't require runtime availability

2. **`pyproject.toml`**:
   - Remove `types-boto3[secretsmanager]` from `aws-secrets-manager` extra
   - Keep it in the `linting` dependency group for type checking

## Testing

The type annotation `_secretsmanager_client: SecretsManagerClient` on `AWSSecretsManagerSettingsSource` still works because `from __future__ import annotations` makes it lazy.

### DIFF
--- a/pydantic_settings/sources/providers/aws.py
+++ b/pydantic_settings/sources/providers/aws.py
@@ -21,7 +21,6 @@ def import_aws_secrets_manager() -> None:
 
     try:
         from boto3 import client as boto3_client
-        from types_boto3_secretsmanager.client import SecretsManagerClient  # noqa: F401
     except ImportError as e:  # pragma: no cover
         raise ImportError(
             'AWS Secrets Manager dependencies are not installed, run `pip install pydantic-settings[aws-secrets-manager]`'
--- a/pydantic_settings/sources/providers/aws.py
+++ b/pydantic_settings/sources/providers/aws.py
@@ -15,6 +15,7 @@ if TYPE_CHECKING:
 
 
 boto3_client = None
+SecretsManagerClient = None
 
 
 def import_aws_secrets_manager() -> None:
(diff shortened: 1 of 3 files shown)


hramezani:
Thanks @gavin913-lss for the PR

Please fix the lint error

gavin913-lss:
Fixed! The issue was ruff formatting. I've applied ruff format and pushed the fix.

hramezani:
> Fixed! The issue was ruff formatting. I've applied ruff format and pushed the fix.

You made it worse. Please fix the lint problem and check

gavin913-lss:
Done! Removed the runtime import of types_boto3_secretsmanager from import_aws_secrets_manager() entirely. The TYPE_CHECKING import already provides the type for mypy, and the runtime import was causing ImportError since types-boto3 was removed from the extras.

hramezani:
Thanks @gavin913-lss

PR #882, opened 2026-06-04
Support using a table header as root when reading a TOML file

### LINKED ISSUE #881: `TomlConfigSettingsSource` should support `toml_table_header`
I have an app that can be configured using env vars and/or a TOML file. I have all the env vars prefixed with `APP_`, but it seems logical to me that in the TOML configuration file, the root settings also are in the `[app]` table. I happen to dislike TOML files without table headers, it seems in general clearer to me, and more logical that the TOML config names matches the env var names.

Thus, I would like to be able to use the `pyproject_toml_table_header` logic from the `PyprojectTomlConfigSettingsSource` in `TomlConfigSettingsSource` too. I have a working piece of code, but need to open create a proper PR, and add some tests.

```python
class Settings(BaseSettings):
    model_config = SettingsConfigDict(env_prefix="APP_", env_nested_delimiter="_")

    log_level: str = "INFO"

    @classmethod
        cls,
        return init_settings, env_settings, TomlConfigSettingsSource(settings_cls, toml_table_header=("app",))
```
```shell
# Using env vars:
APP_LOG_LEVEL = "OTHER"
```
```toml
# Or using toml config:
[app]
log_level = "OTHER"
```
Any ideas, ~~or should I just open a PR~~?

Edit: that was easier than expected. PR opened.

whyscream:
Support `toml_table_header` for regular TOML files.

It can be set from `model_config` or from `settings_customise_sources`.

Resolves #881

### DIFF
--- a/tests/test_source_toml.py
+++ b/tests/test_source_toml.py
@@ -23,7 +23,7 @@ except ImportError:
 
 def test_repr() -> None:
     source = TomlConfigSettingsSource(BaseSettings, Path('config.toml'))
-    assert repr(source) == 'TomlConfigSettingsSource(toml_file=config.toml)'
+    assert repr(source) == 'TomlConfigSettingsSource(toml_file=config.toml, toml_table_header=())'
 
 
 @pytest.mark.skipif(sys.version_info <= (3, 11) and tomli is None, reason='tomli/tomllib is not installed')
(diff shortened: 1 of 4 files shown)


tests/test_source_toml.py
```diff
@@ -158,3 +158,61 @@ def settings_customise_sources(
 
     s = Settings()
     assert s.model_dump() == {'hello': 'world', 'nested': {'foo': 3, 'bar': 2 if deep_merge else 0}}
+
```
hramezani, line 161: Could you add a test for the case where toml_table_header references a key that doesn't exist in the TOML file? Currently self.toml_data.get(key, {}) silently returns {}, which would result in missing/default field values. It'd be good to have a test that documents this behavior explicitly.

hramezani:
Thanks @whyscream for the PR.

`toml_table_header` should also be added to the `_settings_warn_unused_config_keys` check in `main.py`.
Currently line 560 only checks for ('toml_file',):
                                                                                                                                                       
`warn_if_not_used(TomlConfigSettingsSource, ('toml_file',))`

It should be:                                                                                                                                        

`warn_if_not_used(TomlConfigSettingsSource, ('toml_file', 'toml_table_header'))`
                            
Without this, if a user sets toml_table_header in model_config but forgets to add a TomlConfigSettingsSource to their sources, they won't get the expected warning.

tests/test_source_toml.py
whyscream: @hramezani Nice catch! I added a test for the missing table header, and an explicit error, based on how it's done with `yaml_config_section`.

hramezani (approved):

hramezani:
Thanks @whyscream

PR #886, opened 2026-06-12
Don't complain about missing table headers in a TOML file that isn't there

whyscream:
In PR #882 I added support for TOML table headers.

I noticed that I broke one thing: If you set up a `TomlConfigSettingsSource` and then pass `toml_file=None`, no error is raised: it's okay to not have a config file after all. Configuring a  `toml_table_header` broke this: the code expects that a file is read and that the header should be in there, otherwise an error is raised. The possibility of having no config file was lost.

With this PR, it works like this:
- no toml file: no error
- toml file exists and table header found: no error
- toml file exists but table header is missing: error


hramezani (approved):

hramezani:
Thanks @whyscream 

I didn't see Copilot reviews and merged the PR. Could you please create another PR and address them?

PR #895, opened 2026-06-30
Add explicit callout for the case where the unprefixed value is in dotenv

martinky24:
This MR adds an explicit documentation note for an edge case we ran into while using pydantic-settings, where unrecognized settings throw an error (good!), but unrecognized settings (due to lacking prefix) do not throw an error (error prone). If this is actually a bug in the code I am happy to instead open a fix for that.

This example below demonstrates the issue, and hopefully highlights the unintuitive behavior. 

```py3


class Settings(BaseSettings):
        env_prefix="APP_",
        env_file=".env",
    )

    debug: bool = False
    port: int = 8000
    database_url: str


if __name__ == "__main__":
    settings = Settings()
```

with this `.env`

```sh
# APP_DEBUG=true
DEBUG=true
APP_PORT=8080
APP_DATABASE_URL=postgresql://localhost/mydb
# ABC=1
```

Uncomment `ABC` and the script throws a ValidationError, but very unintuitively, `DEBUG` is silently ignored, even though it's not in the model either.

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -728,11 +728,12 @@ class Settings(BaseSettings):
     So if you provide extra values in a dotenv file, whether they start with `env_prefix` or not,
     a `ValidationError` will be raised.
 
-    There is one exception: when an `env_prefix` is set, an unprefixed dotenv entry whose name matches a field name
-    (e.g. `debug` for a `debug` field, while the prefix is `app_`) is silently skipped rather than treated as an
-    extra value. It neither populates the field — only the prefixed name `app_debug` does that — nor raises a
-    `ValidationError`. Note also that the `ValidationError` above is only raised under the default `extra='forbid'`;
-    with `extra='ignore'` the extra values are dropped instead.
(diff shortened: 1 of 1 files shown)


hramezani (approved):

hramezani:
Thanks @martinky24

martinky24:
@hramezani Thanks a ton for helping to get this over the line!

I thought about this a bit more the last few days. Is a doc fix the _right_ fix here? I'm not convinced this isn't a bug that shouldn't be fixed in code. I'd be happy to make that change if so... 

I had hoped to bring this up before it got merged, I was a few hours too late!

hramezani:
@martinky24, this is intended behaviour. We had some discussion around this(unfortunately, I can't remember the details and it needs checking for issues and PRs) and we decided to have it like this.

I am still open to changing it, but we probably have to do it in the next major release(I don't know when).
I think we can have a config flag for that if we want to change it to keep the existing behaviour

martinky24:
@hramezani  Thank you very much for the prompt reply! I really appreciate the help.

PR #902, opened 2026-07-07
Support Traversable for json/toml/yaml config file sources (#299)

### LINKED ISSUE #299: Typing for multiple PathTypes (and support Traversable) for SettingsConfigDict json_file, toml_file, yaml_file, etc
There are issues with the typing for the file options for `SettingsConfigDict`:

1. They are typed as `PathType | None` despite documentation indicating they accept lists (which indeed they do).

https://github.com/pydantic/pydantic-settings/blob/bcbfd17affa5a90663c9afc5a7dc67b281a728dc/pydantic_settings/main.py#L43-L46

The documentation indicates this at: https://docs.pydantic.dev/latest/concepts/pydantic_settings/#other-settings-source

2. `PathType` could, but does not not include `importlib.resources.abc.Traversable`:

https://github.com/pydantic/pydantic-settings/blob/bcbfd17affa5a90663c9afc5a7dc67b281a728dc/pydantic_settings/sources.py#L87 

I should note that if one provides a `Traversable`, it _does_ currently work, indicating the subset of `Path` functionality used by `pydantic-settings` is within the `Traversable` subset of `Path`. 

In the example below, I pass a `Traversable`, and I have verified in a test that it does indeed load the packaged resource file just fine.

---

### Motivation

A somewhat useful pattern is to package a default configuration file _inside_ a Python package. For example:

```python
class FoobarSettings(BaseSettings):
    data_dir: Path

        env_prefix="foobar_",
        toml_file=[
            resources.files("foobar").joinpath("foobar.defaults.toml"),
            "foobar.toml"
        ],
    )
    
    @classmethod
            cls,
        default_sources = super().settings_customise_sources(
            settings_cls, init_settings, env_settings, dotenv_settings, file_secret_settings
        )
        return TomlConfigSettingsSource(settings_cls), *default_sources
```

And then have in `src/foobar/foobar.defaults.toml`:

```toml
data_dir = ".foobar/data"
```

**This currently _works_.** However, it will not type-check.

hramezani:
## Summary

Closes #299.

Adds typing and runtime support for passing a `Traversable` (e.g. `importlib.resources.files(...).joinpath(...)`) as `json_file`, `toml_file`, or `yaml_file`. This enables loading a configuration file that is packaged inside a distribution — including resources that don't live on the filesystem, such as a file inside a zip/wheel — without a `cast` to `Path`.

### Background

Issue #299 raised two points:

1. **File options don't type-check against lists** — already resolved: `PathType` includes `Sequence[...]`.
2. **`PathType` should include `Traversable`** — still a real typing gap. The issue also noted that a `Traversable` "currently works" at runtime. That is now only partially true: a **genuine non-`Path` `Traversable`** (a zip/wheel-packaged resource) still works for JSON/YAML, but **regressed for TOML** — `TomlConfigSettingsSource._any_file_exists` wraps every path in `Path(path)`, which raises `TypeError` for an object that isn't `os.PathLike`.

### Changes

- Introduce a dedicated `ConfigFileSourceType` alias (`Path | str | Traversable | Sequence[Path | str | Traversable]`) and apply it to the config-file fields of `SettingsConfigDict` and to the `Json`/`Toml`/`Yaml` `*ConfigSettingsSource` constructors (and the shared `_read_files`/`_read_file`). `PathType` — used by `secrets_dir` — is intentionally left unchanged so directory handling keeps its `Path`-only semantics.
- Fix `TomlConfigSettingsSource._any_file_exists` to query a non-`Path` `Traversable` directly via `is_file()` instead of coercing through `Path(...)`. This also fixes a latent bug where a single `str` path was iterated character-by-character.
- Version-guard the `Traversable` import (`importlib.resources.abc` on 3.11+, `importlib.abc` on 3.10), since `importlib.abc.Traversable` was removed in 3.12+.
- Export `ConfigFileSourceType` from the package.
- Add a docs note under the config-file sources section.

### Tests

- New zip-packaged (genuine non-`Path`) `Traversable` regression tests for the TOML source (plain load and `toml_table_header`) and the YAML source.
- The JSON source already had `TestTraversableSupport`; updated its now-stale "not added in annotation" comment.

Verified locally: `mypy` clean (no new errors over baseline), `ruff` clean, full test suite green (excluding pre-existing Windows-only symlink test failures unrelated to this change).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -2810,6 +2810,12 @@ The files are merged shallowly in increasing order of priority. To enable deep m
 !!! warning
     The `deep_merge` option is **not available** through the `SettingsConfigDict`.
 
+!!! note
+    In addition to `str` and `pathlib.Path`, the `json_file`, `toml_file`, and `yaml_file` options also accept a
+    [`Traversable`][importlib.resources.abc.Traversable], such as the result of `importlib.resources.files(...).joinpath(...)`.
+    This makes it possible to load a configuration file that is packaged inside your distribution, including cases where the
+    resource does not live on the filesystem (e.g. inside a zip/wheel).
+
 ```py
 from pydantic import BaseModel
 
--- a/pydantic_settings/main.py
+++ b/pydantic_settings/main.py
@@ -21,6 +21,7 @@ from .exceptions import SettingsError
 from .sources import (
     ENV_FILE_SENTINEL,
     CliSettingsSource,
+    ConfigFileSourceType,
     DefaultSettingsSource,
     DotenvFiltering,
     DotEnvSettingsSource,
@@ -72,9 +73,9 @@ class SettingsConfigDict(ConfigDict, total=False):
     cli_kebab_case: bool | Literal['all', 'no_enums'] | None
     cli_shortcuts: Mapping[str, str | list[str]] | None
     secrets_dir: PathType | None
-    json_file: PathType | None
+    json_file: ConfigFileSourceType | None
     json_file_encoding: str | None
-    yaml_file: PathType | None
+    yaml_file: ConfigFileSourceType | None
     yaml_file_encoding: str | None
     yaml_config_section: str | None
     """
@@ -105,7 +106,7 @@ class SettingsConfigDict(ConfigDict, total=False):
     To use the root table, exclude this config setting or provide an empty tuple.
     """
 
-    toml_file: PathType | None
+    toml_file: ConfigFileSourceType | None
 
     toml_table_header: tuple[str, ...]
     """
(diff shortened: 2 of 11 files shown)


hramezani:
Thanks for the review! Addressed in 450bf45:

- **Skip condition off-by-one (both TOML tests):** Changed `sys.version_info <= (3, 11)` → `< (3, 11)` for the two new `Traversable` tests, matching the guard in `import_toml()` (where `tomli` is only required below 3.11). In practice `sys.version_info` always carries a micro version, so `<= (3, 11)` never actually matched a `3.11.x` release, but `< (3, 11)` is the clearer and correct form.
- **`ConfigFileSourceType` top-level export (suppressed note):** Kept the export at `pydantic_settings.sources` only, for consistency with the sibling aliases `PathType` and `DotenvType`, which are likewise not re-exported from the top-level `pydantic_settings` package. "Export from the package" in the description referred to `pydantic_settings.sources`.

hramezani:
@copilot resolve the merge conflicts in this pull request

PR #908, opened 2026-07-10
docs: recommend async settings loading pattern

### LINKED ISSUE #479: Recommend a way to load/reload settings in async environment

w3lld1:
## Summary
- document the recommended `asyncio.to_thread` pattern for constructing settings from async code
- call out synchronous file-backed sources and how to reload a mutable settings instance without blocking the event loop
- include a tested dotenv-backed example before the existing in-place reload section

Fixes #479

## Validation
- `uv run --group testing pytest 'tests/test_docs.py::test_docs_examples' -q -k '3249 or Async or 3271'`
- `uv run --group testing --extra azure-key-vault pytest tests/test_docs.py -q`

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -3319,13 +3319,14 @@ from pathlib import Path
 
 from pydantic_settings import BaseSettings, SettingsConfigDict
 
-Path('.env').write_text('api_key=secret\n', encoding='utf-8')
+env_path = Path('example.env')
+env_path.write_text('api_key=secret\n', encoding='utf-8')
 
 
 class Settings(BaseSettings):
     api_key: str
 
-    model_config = SettingsConfigDict(env_file='.env')
+    model_config = SettingsConfigDict(env_file=env_path)
 
 
 async def load_settings() -> Settings:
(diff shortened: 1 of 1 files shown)


hramezani:
Thanks @w3lld1

PR #910, opened 2026-07-12
fix: parse enum names through nested annotations

Sanjays2402:
With `env_parse_enums=True`, enum names inside nested wrappers such as `Optional[Annotated[MyEnum, ...]]` were left as strings and failed validation. Resolve enum members recursively through annotation arguments, while preserving existing direct and union handling; the regression test fails before the source change and passes with it.

### DIFF
--- a/pydantic_settings/sources/utils.py
+++ b/pydantic_settings/sources/utils.py
@@ -250,10 +250,15 @@ def _strip_annotated(annotation: Any) -> Any:
 
 
 def _annotation_enum_val_to_name(annotation: type[Any] | None, value: Any) -> str | None:
-    for type_ in (annotation, get_origin(annotation), *get_args(annotation)):
+    for type_ in (annotation, get_origin(annotation)):
         if _lenient_issubclass(type_, Enum):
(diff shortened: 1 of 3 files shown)


hramezani:
Thanks @Sanjays2402  for the fix! I pushed one follow-up commit (`761bfb9`) to the branch.

`_annotation_enum_val_to_name` — the twin of the function you fixed — had the exact same one-level lookup limitation. It only inspected the annotation, its origin, and its *direct* args, so an enum wrapped as `Optional[Annotated[MyEnum, ...]]` was not resolved. This shows up in CLI help rendering of enum defaults, which fell back to the raw value (e.g. `(default: 1)`) instead of the member name (`(default: kiwi)`).

The commit applies the same recursive-descent pattern you used for `_annotation_enum_name_to_val`, plus a regression test (`test_cli_enum_default_through_nested_annotation`) that fails before the change and passes with it. This keeps the two helpers symmetric so both the env-parsing and CLI-help paths handle nested annotations consistently.
@

hramezani:
Thanks @Sanjays2402

PR #912, opened 2026-07-13
Fix dotenv extras being claimed by complex fields sharing a name prefix

### LINKED ISSUE #911: Dotenv extras silently dropped when their name shares a complex field's prefix
### Description

`DotEnvSettingsSource` decides whether a `.env` variable "belongs" to a complex field (a `dict` or nested model) using a bare string prefix check:

    env_name.startswith(field_env_name)

This matches any variable that merely *starts with* the field's name, not just variables actually nested under it via `env_nested_delimiter`. As a result, a variable like `dbx_token` is wrongly treated as belonging to a field named `db`, and is silently swallowed instead of being kept as an extra input.

With `extra='forbid'`, this means an unexpected variable that happens to share a name prefix with a complex field bypasses validation entirely — no `ValidationError` is raised, even though it should be treated as a forbidden extra input.

### Example Code

```python

# .env contains: dbx_token=secret

class Settings(BaseSettings):
    model_config = SettingsConfigDict(env_file='.env', extra='forbid')
    db: dict = {}

Settings()  # does NOT raise -- but 'dbx_token' is an unrelated extra field
            # that should trigger "Extra inputs are not permitted"
```

With `extra='allow'` instead, `dbx_token` is silently dropped rather than
appearing in the model:

```python
class Settings(BaseSettings):
    model_config = SettingsConfigDict(env_file='.env', extra='allow')
    db: dict = {}

s = Settings()
print(s.model_dump())  # {'db': {}} -- dbx_token vanished entirely
```

ritsth:
Fixes #911

The extras pass in `DotEnvSettingsSource.__call__` marked a var as "used" by a complex field on a bare `startswith(field_env_name)` — so `dbx_token` was claimed by a `db: dict` field and silently dropped (`extra='allow'`), or escaped the forbidden-extra error (`extra='forbid'`). This is the complex-field counterpart of the similar-name case already tested for simple fields.

The fix requires the nested delimiter boundary (`db<delimiter>...`), which is exactly what `explode_env_vars` consumes, so legitimately nested vars (`db__host`) are still not duplicated as extras.

Added three regression tests (no delimiter, delimiter boundary, forbid); all fail before the fix. Full suite, ruff, and mypy pass.

Used an AI assistant to find and fix this; I reviewed and verified everything myself.

### DIFF
--- a/pydantic_settings/sources/providers/dotenv.py
+++ b/pydantic_settings/sources/providers/dotenv.py
@@ -150,7 +150,13 @@ class DotEnvSettingsSource(EnvSettingsSource):
                                 and _union_is_complex(field.annotation, field.metadata, self._init_state)
                             )
                         )
-                        and env_name.startswith(field_env_name)
+                        # A var only belongs to a complex field when it sits at the nested
+                        # delimiter boundary (`db<delim>...`, matching what explode_env_vars
+                        # consumes). A bare name-prefix match (e.g. `dbx_token` vs field `db`)
+                        # must not claim the var, otherwise it is silently dropped instead of
+                        # being kept as an extra.
+                        and bool(self.env_nested_delimiter)
+                        and env_name.startswith(f'{field_env_name}{self.env_nested_delimiter}')
                     ):
                         env_used = True
                         break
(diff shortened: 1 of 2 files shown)


hramezani (approved):

hramezani:
Thanks @ritsth

PR #914, opened 2026-07-16
fix: coerce empty yaml_config_section to an empty mapping instead of crashing

chuenchen309:
## Change

A YAML section that is present but empty (e.g. `nested:` with no children — a common "this section is all defaults" pattern) is parsed by PyYAML as `None`. `_traverse_nested_section` returns that `None` verbatim, and it is then handed to `InitSettingsSource`, which does `set(init_kwargs.keys())` and raises:

```
AttributeError: 'NoneType' object has no attribute 'keys'
```

An empty *whole file* already works, because `_read_file` coerces it with `yaml.safe_load(...) or {}`. An empty *section* should behave the same way and fall back to defaults, rather than crashing with an opaque error from deep in the init machinery.

## Reproduction

```python
# config.yaml contains just:  nested:
class Settings(BaseSettings):
    foo: str = "default"
    model_config = SettingsConfigDict(yaml_file="config.yaml", yaml_config_section="nested")

    @classmethod
    def settings_customise_sources(cls, settings_cls, *args):
        return (YamlConfigSettingsSource(settings_cls),)

Settings()  # AttributeError: 'NoneType' object has no attribute 'keys'
```

## Fix

Apply the same `or {}` coercion `_read_file` already uses, to the traversed section result.

## Test

Added `test_yaml_config_section_empty_section` (present-but-empty section falls back to defaults). Full `tests/test_source_yaml.py` passes (19), and `ruff format/check` + `mypy` are clean on the touched file.

### DIFF
--- a/pydantic_settings/sources/providers/yaml.py
+++ b/pydantic_settings/sources/providers/yaml.py
@@ -61,8 +61,11 @@ class YamlConfigSettingsSource(InitSettingsSource, ConfigFileSourceMixin):
         self.yaml_data = self._read_files(self.yaml_file_path, deep_merge=deep_merge)
 
         if self.yaml_config_section is not None:
-            self.yaml_data = self._traverse_nested_section(
-                self.yaml_data, self.yaml_config_section, self.yaml_config_section
+            # An empty section (`section:` with no children) parses to `None`; coerce it to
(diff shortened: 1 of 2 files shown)


hramezani (approved):

hramezani:
Thanks @chuenchen309

PR #917, opened 2026-07-18
fix: read secret files as UTF-8 instead of the locale encoding

### LINKED ISSUE #916: UTF-8 secret files can be silently decoded with the Windows locale encoding
### Summary

`SecretsSettingsSource` reads each secret with `Path.read_text()` without specifying an encoding. On Windows systems whose locale encoding is not UTF-8, a UTF-8 secret file can therefore be decoded using the active Windows code page.

For some byte sequences this does not raise `UnicodeDecodeError`; it silently produces a different string. The same implicit `read_text()` behavior also appears in `NestedSecretsSettingsSource`.

Relevant code:

- [`SecretsSettingsSource` in v2.14.2](https://github.com/pydantic/pydantic-settings/blob/v2.14.2/pydantic_settings/sources/providers/secrets.py#L119-L122)
- [`NestedSecretsSettingsSource` in v2.14.2](https://github.com/pydantic/pydantic-settings/blob/v2.14.2/pydantic_settings/sources/providers/nested_secrets.py#L201-L203)

### Environment

- `pydantic-settings==2.14.2` from PyPI
- `pydantic==2.13.4`
- Python `3.12.10`
- Windows 11 (`10.0.22621`)
- Windows locale encoding: `cp936`
- Python UTF-8 mode: disabled

### Minimal reproduction

```python
import locale
import sys
import tempfile
from pathlib import Path

class Settings(BaseSettings):
    token: str


with tempfile.TemporaryDirectory() as directory:
    secrets_dir = Path(directory)
    expected = "p\u00e4ssw\u00f6rd-\U0001f510"
    (secrets_dir / "token").write_bytes(expected.encode("utf-8"))

    actual = Settings(_secrets_dir=secrets_dir).token
    print("locale:", locale.getpreferredencoding(False))
    print("utf8_mode:", sys.flags.utf8_mode)
    print("expected:", ascii(expected))
    print("actual:  ", ascii(actual))
    assert actual == expected
```

Run normally on a Windows installation using a non-UTF-8 locale:

```powershell
py -3.12 repro.py
```

Observed output:

```text
locale: cp936
utf8_mode: 0
expected: 'p\xe4ssw\xf6rd-\U0001f510'
actual:   'p\u76f2ssw\u679ard-\u9983\u653c'
AssertionError
```

Control run:

```powershell
py -3.12 -X utf8 repro.py
```

The control succeeds and returns the original value.

I reproduced the mismatch twice with the PyPI `v2.14.2` package and twice with the current `main` checkout (`1d0ea26`). An expanded local check confirms the same result for both `SecretsSettingsSource` and `NestedSecretsSettingsSource`; both sources return the original value in the `-X utf8` control run.

The relevant existing test suites still pass:

- `tests/test_settings.py -k secrets`: `14 passed, 6 skipped`
- `tests/test_source_nested_secrets.py`: `36 passed`

The current tests do not appear to cover a file written explicitly as UTF-8 and then read under a non-UTF-8 Windows locale.

Running the expanded reproduction with `-X warn_default_encoding` also emits Python's `EncodingWarning` at both `read_text()` calls. This matches the Python documentation: `Path.read_text()` delegates its optional encoding behavior to `open()`, whose default text encoding is locale-specific. See [pathlib](https://docs.python.org/3.12/library/pathlib.html#pathlib.Path.read_text) and [Text Encoding](https://docs.python.org/3.12/library/io.html#text-encoding).

### Expected behavior

Loading a text secret should not silently change its Unicode value based on the host's locale encoding. Possible approaches might be:

1. define UTF-8 as the default encoding for secret files; or
2. add an explicit secret-file encoding setting, similar to `env_file_encoding`, and use it in both standard and nested secret sources.

I realize changing the default may have backward-compatibility implications for existing locale-encoded files, so an explicit configuration option may be preferable. If locale-dependent decoding is intentional, documenting that behavior would also remove the current ambiguity.

I searched open and closed issues and pull requests using `encoding`, `UTF-8`, `secrets encoding Windows`, `secrets UTF-8`, `secret file encoding`, `read_text secrets`, and `cp936 secrets`, and did not find a matching report.

dchaudhari7177:
## Summary

Fixes #916. `SecretsSettingsSource` and `NestedSecretsSettingsSource` both called `Path.read_text()` without an encoding, so secret files were decoded with the **platform locale encoding**. On Windows systems whose code page is not UTF-8, a UTF-8 secret is misread.

Both sources now read with `encoding='utf-8'`. This matches the dotenv provider, which already defaults to UTF-8 (`encoding or 'utf8'`). (The JSON and YAML providers expose a `*_file_encoding` option but fall back to the locale encoding when it is unset, so they share the same latent issue — that is out of scope here and tracked separately.)

## Reproduced locally

I'm on Windows 11 with `cp1252`, `utf8_mode: 0`, so I could reproduce both failure modes on real hardware rather than simulating them.

**Silent corruption** — every UTF-8 byte happens to map in cp1252, so nothing raises and the field just gets the wrong value:

```
locale:   cp1252
expected: 'caf\xe9-p\xe4ssw\xf6rd'
actual:   'caf\xc3\xa9-p\xc3\xa4ssw\xc3\xb6rd'   <- silently doubly encoded
```

**Hard failure** — with a byte undefined in the code page (the issue's `\U0001f510` case), it raises instead:

```
UnicodeDecodeError: 'charmap' codec can't decode byte 0x90 in position 14
  -> SettingsError: error getting value for field "token" from source "SecretsSettingsSource"
```

The issue's exact reproduction script now prints `MATCH` on this machine, and `NestedSecretsSettingsSource.load_secrets()` round-trips too.

## Tests

Four tests, two per source: a plain UTF-8 round-trip, and one under a simulated non-UTF-8 locale.

The second kind needs care, because **the tests must also fail on a UTF-8 CI host**. The real lever is the platform locale, which a test can't portably change: CPython resolves `open()`'s default encoding in C, so patching `locale.getpreferredencoding` has no effect (verified on 3.12 — the read still used the true cp1252).

So the new `non_utf8_default_encoding` fixture makes an **encoding-less** `Path.read_text()` fall back to cp1252 while leaving explicit-encoding reads untouched. That is exactly the bug's precondition, and it reproduces on any platform.

Verified the tests genuinely catch the bug by reverting only the source change and re-running — all four fail, covering both modes:

| test | failure without the fix |
| --- | --- |
| `test_secrets_path_utf8` | `SettingsError` |
| `test_secrets_path_utf8_under_non_utf8_locale` | `AssertionError` (silent corruption) |
| `test_load_secrets_reads_utf8` | `UnicodeDecodeError` |
| `test_load_secrets_utf8_under_non_utf8_locale` | `AssertionError` (silent corruption) |

## Verification

- `make lint` → all checks passed, 36 files already formatted
- Full suite → **609 passed, 12 skipped**

Two pre-existing issues, both confirmed on a clean tree by stashing and unrelated to this change:

- 6 `test_symlink*` / `test_cyclic*` failures in `test_source_nested_secrets.py` — `OSError` because creating symlinks on Windows needs elevation
- `mypy pydantic_settings` reports `utils.py:35: Cannot call function of unknown type`

## Note on scope

I kept this to a hardcoded `utf-8` rather than adding a configurable `secrets_dir_encoding` alongside `json_file_encoding` / `yaml_file_encoding`. UTF-8 is the right default for secret files (Docker and Kubernetes secrets are UTF-8), and adding public config felt like your call rather than mine. Happy to add it if you'd prefer the symmetry.


hramezani (approved):

hramezani:
Thanks @dchaudhari7177

Issue #732, opened 2025-12-17
Settings: Parsing comma-separated list of strings


vinnybod:
### Initial Checks

- [x] I have searched Google & GitHub for similar requests and couldn't find anything
- [x] I have read and followed [the docs](https://docs.pydantic.dev) and still think this feature is missing

### Description

I'd like to be able to set a `list[str]` variable with a comma-separated list. (eg `CORS_ALLOW_ORIGINS=http://localhost:8080,http://example.com`. I am running into this problem because trying to work with escaped quotes is quite [problematic in github actions](https://github.com/pydantic/pydantic/issues/1458#issuecomment-831292347).

Related (old) issue from pydantic v1 - https://github.com/pydantic/pydantic/issues/1458

This appeared to be a solved problem in https://github.com/pydantic/pydantic/commit/fe7c9da6c19d602907aa236bec3cde86495c8c18, but the recommended approach no longer works in v2 because `parse_env_var` was removed in pydantic v2.

I searched through the v2 documents and haven't found a straightforward recommendation, and it seems like this should be something that doesn't require so much effort to workaround.

Searching github I found a few implementation examples, all slightly different.
https://github.com/kfstorm/sonarr-metadata-rewrite/blob/7485c70708111f871759dad71ce9b593222eea07/src/sonarr_metadata_rewrite/config.py#L15
https://github.com/mozdowski/ssh-control-mcp/blob/63043a05a9906eb42082fb92ac15c5407bd650aa/ssh_server_mcp/config.py#L19
https://github.com/dthuy0224/File2Learning/blob/d3198cd7aabba8b3d891372569151dded93c3078/backend/app/core/config.py#L15

### Affected Components

- [ ] [Compatibility between releases](https://docs.pydantic.dev/changelog/)
- [x] [Data validation/parsing](https://docs.pydantic.dev/concepts/models/#basic-model-usage)
- [ ] [Data serialization](https://docs.pydantic.dev/concepts/serialization/) - `.model_dump()` and `.model_dump_json()`
- [ ] [JSON Schema](https://docs.pydantic.dev/concepts/json_schema/)
- [ ] [Dataclasses](https://docs.pydantic.dev/concepts/dataclasses/)
- [ ] [Model Config](https://docs.pydantic.dev/concepts/config/)
- [ ] [Field Types](https://docs.pydantic.dev/api/types/) - adding or changing a particular data type
[...]

Issue #570, opened 2025-03-26
Comma separated string as list[str]


vitalik:
What's the recommended way to create field type that gets some string form env and then splits it by comma? 

trying this:

```python
from typing import Annotated, Any
from pydantic_settings import BaseSettings
from pydantic import BeforeValidator


def comma_sep_list(value: Any) -> list[str]:
    if not value:
        return []
    if isinstance(value, str):
        return value.split(',')
    return value


CommaSepList = Annotated[list[str], BeforeValidator(comma_sep_list)] # !!!!


class Settings(BaseSettings):
    SOME_VALUES: CommaSepList = []


settings = Settings()

print(repr(settings.SOME_VALUES))

```

trying out:
```
$ SOME_VALUES=1,2,three python main.py
```

gives an error:

```
Traceback (most recent call last):
  File "/private/tmp/polyzoary_multitarian/.venv/lib/python3.12/site-packages/pydantic_settings/sources.py", line 610, in __call__
    field_value = self.prepare_field_value(field_name, field, field_value, value_is_complex)
                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/private/tmp/polyzoary_multitarian/.venv/lib/python3.12/site-packages/pydantic_settings/sources.py", line 827, in prepare_field_value
    raise e
  File "/private/tmp/polyzoary_multitarian/.venv/lib/python3.12/site-packages/pydantic_settings/sources.py", line 824, in prepare_field_value
    value = self.decode_complex_value(field_name, field, value)
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/private/tmp/polyzoary_multitarian/.venv/lib/python3.12/site-packages/pydantic_settings/sources.py", line 333, in decode_complex_value
    return json.loads(value)
           ^^^^^^^^^^^^^^^^^
  File "/opt/homebrew/Cellar/python@3.12/3.12.7_1/Frameworks/Python.framework/Versions/3.12/lib/python3.12/json/__init__.py", line 346, in loads
    return _default_decoder.decode(s)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^
[...]

Issue #291, opened 2024-05-22
Comma delimited string as list of strings


romanzdk:
I use to define settings class like e.g.

```python
class CoolSettings(pydantic_settings.BaseSettings):
    COOL_LIST: list[str] = ['xy', 'ab']
```

Would it be possible to pass `COOL_LIST` env var as comma separated string? I.e. I would define my env var as `COOL_LIST=abc,def` and it would be parsed to list of strings? Of course I would have to specify the separator? I recon `COOL_LIST='["abc","def"]'` works but is not very user friendly to write. Or ofc custom parser could be written.

hramezani:
Thanks @romanzdk for this issue. Right now it is not possible to load a list from comma-separated env value. you can create a [custom source class](https://docs.pydantic.dev/latest/concepts/pydantic_settings/#adding-sources), inherit from [EnvSettingsSource](https://github.com/pydantic/pydantic-settings/blob/6d25cee4bb7a6db592ca0da123c53f3d775cd1e1/pydantic_settings/sources.py#L443) and override the [prepare_field_value](https://github.com/pydantic/pydantic-settings/blob/6d25cee4bb7a6db592ca0da123c53f3d775cd1e1/pydantic_settings/sources.py#L492C9-L492C28)

tcztzy:
I wrote a decorator to patch original `EnvSettingSource`
```python
from typing import cast
from pydantic_settings import (
    BaseSettings,
    EnvSettingsSource,
    PydanticBaseSettingsSource,
)

def parse_comma_separated_list(self: EnvSettingsSource):
    origin_prepare_field_value = self.prepare_field_value

    def prepare_field_value(
        field_name: str,
        field: FieldInfo,
        value: Any,
        value_is_complex: bool,
    ) -> Any:
        if (
            isinstance(value, str) and
            get_origin(annotation := field.annotation) is list and
            isinstance(extra := field.json_schema_extra, dict) and
            (delimiter := extra.get("delimiter")) is not None
        ):
            if not isinstance(delimiter, str):
                raise SettingsError(
                    f"Invalid delimiter for comma-separated list: {delimiter} (type: {type(delimiter)})"
                )
            item_type, = get_args(annotation)
            validate = TypeAdapter(item_type).validate_python
            return [validate(item) for item in value.split(delimiter)]
        return origin_prepare_field_value(field_name, field, value, value_is_complex)
    
    self.prepare_field_value = prepare_field_value

[...]

Issue #80, opened 2023-06-21
error parsing value for field <...> from source "EnvSettingsSource"


fredrikaverpil:
Hi 👋

I just saw the [FastAPI 0.100.0 prerelease announcement](https://github.com/tiangolo/fastapi/releases/tag/0.100.0-beta1), the [Pydantic v2 migration guide](https://docs.pydantic.dev/dev-v2/migration/) and wanted to give this all a try with one of my project's test suites with the latest and greatest.

I ended up getting this error:

```python
self = EnvSettingsSource(env_nested_delimiter=None, env_prefix_len=0)

    def __call__(self) -> dict[str, Any]:
        data: dict[str, Any] = {}
    
        for field_name, field in self.settings_cls.model_fields.items():
            try:
                field_value, field_key, value_is_complex = self.get_field_value(field, field_name)
            except Exception as e:
                raise SettingsError(
                    f'error getting value for field "{field_name}" from source "{self.__class__.__name__}"'
                ) from e
    
            try:
                field_value = self.prepare_field_value(field_name, field, field_value, value_is_complex)
            except ValueError as e:
>               raise SettingsError(
                    f'error parsing value for field "{field_name}" from source "{self.__class__.__name__}"'
                ) from e
E               pydantic_settings.sources.SettingsError: error parsing value for field "ui_languages" from source "EnvSettingsSource"
```

And this is my code that triggered the error when calling `LanguageSettings().ui_languages`:

```python
from pathlib import Path
from pydantic import Field
from pydantic_settings import BaseSettings


def comma_separated_string_to_set(raw_value: str) -> set[str]:
    if not raw_value:
        msg = f"{raw_value} must be a comma separated string"
        raise ValueError(msg)
    return {item.strip() for item in raw_value.split(",")}


class LanguageSettings(BaseSettings):
    ui_languages: set[str] = Field(env="UI_LANGUAGES")

    class Config:
        env_file = Path("envs/test.env")

        @classmethod
        def parse_env_var(cls, field_name: str, raw_val: str) -> Any:
            if field_name == "ui_languages":
                return comma_separated_string_to_set(raw_val)
```

Here's my poetry config:

```toml
[tool.poetry.dependencies]
fastapi = {extras = ["all"], version = "^0.100.0b1", allow-prereleases = true}
pydantic = {version = "^2.0b3", allow-prereleases = true}
pydantic-settings = {version = "^2.0b1", allow-prereleases = true}
```

[...]

PR #919, opened 2026-07-21
docs: document JSON parsing of complex env values + comma-separated recipe

hrithvikakb:
## Change Summary

Complex fields (`list`, `set`, `dict`, sub-models) are populated from environment variables by JSON-decoding the value. The docs already cover the *escape hatches* (`NoDecode`, `enable_decoding=False`, `ForceDecode`) but never state the failure that motivates them: a value that is **not valid JSON** raises a `SettingsError`.

This is a recurring source of confusion — e.g. `CORS_ORIGINS=https://a.com,https://b.com` for a `list[str]` field fails rather than being read as a comma-separated list, and the error (`error parsing value for field ... from source "EnvSettingsSource"`) doesn't obviously point to "the value must be JSON". It comes up repeatedly (#732, #570, #291, #80, …), each time resolved by pointing at `NoDecode` + a validator.

This PR documents both the failure and the canonical fix:

- A `!!! warning` where complex-type JSON parsing is first introduced, linking forward to *Disabling JSON parsing*.
- A runnable example opening *Disabling JSON parsing* that shows the actual `SettingsError`.
- A reusable `Annotated` type recipe (`NoDecode` + `BeforeValidator`) for accepting comma-separated values across multiple fields — the pattern maintainers already hand out in the issues above.

Docs-only; no code changes.

## Test plan

- [x] `uv run pytest tests/test_docs.py` → 72 passed (the new examples are executed and their output verified by `pytest-examples`).

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -318,6 +318,12 @@ the same way it would be if passed directly to the initialiser (as a string).
 Complex types like `list`, `set`, `dict`, and sub-models are populated from the environment by treating the
 environment variable's value as a JSON-encoded string.
 
+!!! warning
+    Because these values are JSON-decoded, an environment variable whose value is not valid JSON raises a
+    [`SettingsError`](../api/pydantic_settings.md#pydantic_settings.SettingsError). For example, `NUMBERS=1,2,3`
+    for a `numbers: list[int]` field fails, because `1,2,3` is not valid JSON. See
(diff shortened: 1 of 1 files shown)


hramezani (approved):

hramezani:
Thanks @hrithvikakb

PR #922, opened 2026-07-24
feat: support additional fields with pyproject source

gjask:
I use pydantic-settings to create CLI tool that can read configuration from pyproject.toml But some values are already present in standard project table (e.g. `[project].name`). So I tried expanding the source with support for specifiing some additional fields.

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -3038,6 +3038,53 @@ field = "default-table"
 field = "some-table"
 ```
 
+#### Loading additional fields from outside the table
+
+Sometimes you want to reuse a value declared elsewhere in "pyproject.toml" (such as the project version) without moving it into your settings table.
+`SettingsConfigDict(pyproject_toml_additional_fields=tuple[tuple[str, ...], ...])` accepts a tuple of field paths; the leaf key of each path is merged into the settings alongside the values loaded from `pyproject_toml_table_header`.
+Paths that are not present in the file are silently skipped.
+
+```python
+from pydantic_settings import (
+    BaseSettings,
+    PydanticBaseSettingsSource,
+    PyprojectTomlConfigSettingsSource,
+    SettingsConfigDict,
+)
+
+
+class AdditionalFieldsSettings(BaseSettings):
+    """Example pulling `version` from the `[project]` table into the settings."""
+
+    model_config = SettingsConfigDict(
+        pyproject_toml_additional_fields=(('project', 'version'),)
+    )
+
+    field: str
+    version: str
+
+    @classmethod
+    def settings_customise_sources(
+        cls,
+        settings_cls: type[BaseSettings],
+        init_settings: PydanticBaseSettingsSource,
+        env_settings: PydanticBaseSettingsSource,
+        dotenv_settings: PydanticBaseSettingsSource,
+        file_secret_settings: PydanticBaseSettingsSource,
+    ) -> tuple[PydanticBaseSettingsSource, ...]:
+        return (PyprojectTomlConfigSettingsSource(settings_cls),)
+```
+
+Given the following "pyproject.toml", this results in `AdditionalFieldsSettings(field='default-table', version='1.2.3')`:
+
+```toml
+[project]
+version = "1.2.3"
+
+[tool.pydantic-settings]
+field = "default-table"
+```
+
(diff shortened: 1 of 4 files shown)


hramezani:
Thanks for the PR, and sorry for the slow response.

The underlying need is real — reusing `[project].version` or `[project].name` in a CLI tool's settings instead of duplicating them under `[tool.your-tool]` is a reasonable thing to want, and I don't think you're wrong to have hit it.

That said, I don't think I want to take this as a config key, for a few reasons:

**It's a new public API on one source only.** `pyproject_toml_additional_fields` lands on `PyprojectTomlConfigSettingsSource` but not on `TomlConfigSettingsSource`, YAML, or JSON, which all have the same table-header/section concept. Either the idea is general and belongs on all of them, or it's pyproject-specific special-casing — and I'd rather not add the latter to `SettingsConfigDict`, which is already a large surface.

**Keying on the leaf name causes silent collisions.** Because the merge is `toml_data.update(additional_fields)` after descending into the table header, an additional path whose leaf matches a key already in your settings table will silently overwrite it — e.g. `(('project', 'name'),)` beats a `name` set under `[tool.pydantic-settings]`, with no warning. Two additional paths ending in the same leaf clobber each other too, depending on order.

**There's no way to rename.** You get the leaf key or nothing, so your model's field names are forced to match unrelated parts of `pyproject.toml`. If your field is `project_version`, you can't map `[project].version` onto it.

**Missing paths are silently skipped**, which makes typos undebuggable — `(('projekt', 'version'),)` just quietly yields nothing.

The good news is that your case is already solvable today without any new API, by subclassing the source and overriding `__call__`:

```python
from pydantic_settings import BaseSettings, PyprojectTomlConfigSettingsSource


class ProjectAwarePyprojectSource(PyprojectTomlConfigSettingsSource):
    def __call__(self):
        values = super().__call__()
        project = self._read_files(self.toml_file_path).get('project', {})
        if 'version' in project:
            values.setdefault('version', project['version'])
        return values


class Settings(BaseSettings):
    field: str
    version: str

    @classmethod
    def settings_customise_sources(cls, settings_cls, **kwargs):
        return (ProjectAwarePyprojectSource(settings_cls),)
```

Given this `pyproject.toml`:

```toml
[project]
version = "1.2.3"

[tool.pydantic-settings]
field = "default-table"
```

that yields `Settings(field='default-table', version='1.2.3')`. Using `setdefault` means your settings table wins over the outside value; swap it for a plain assignment if you want the opposite. Renaming works the same way — `values.setdefault('project_version', ...)` maps `[project].version` onto a differently-named field — and a file with no `[project]` table just leaves the field alone rather than erroring.

That's a handful of lines in your own code, it's fully general, and it lets you decide for yourself how collisions and renaming behave.

If we did want to support this in the library eventually, I think the shape would be a **mapping** rather than a tuple of paths — `{'version': ('project', 'version')}`, field name → path — which fixes both the renaming and the collision problems, and could then be applied consistently across all the file-based sources. That's a larger design discussion than this PR, though, and not something I want to commit to right now.

So I'm going to close this one — but to be clear, closing here is about demand and design, not about the quality of your work. If more people run into this and ask for it, I'm happy to reopen this PR or revisit the idea properly; a couple of "+1, I need this too" comments on a linked issue would be exactly the signal that changes my mind. If you'd like, open an issue describing the use case (feel free to reference this PR) so there's somewhere for that to accumulate.

I do appreciate the work, including the docs and tests — please don't read this as a signal not to open PRs here.

gjask:
Thanks for the reply. I fully get the need for more design here and overall agree with your resolution. I just felt that I should answer some of those points which I cannot fully agree with. They can be used as basis for future work if needed.

> It's a new public API on one source only

Here I just followed already present concepts. No other source actually trims the data the way the `PyprojectTomlConfigSettingsSource` does. So the change is not applicable elsewhere. But it is logical supplement to `pyproject_toml_table_header` which currently lives only on `SettingsConfigDict`. So this seemed like most consistent approach.

> Keying on the leaf name causes silent collisions.

That is valid point. It shoul probably be at least well documented.

> There's no way to rename.

Yes. Same as you don't get to rename any fields in any sources. So it felt like adding rename would add clunky complexity. But a valid feature request I think.

> Missing paths are silently skipped

I would say this is intended. Validation is performed on the Settings model not on the Source. Otherwise you couldn't have multiple sources at the same time.

PR #926, opened 2026-08-04
fix: raise ValidationError for non-JSON env values on strict fields

shuvamk:
On `main` (ae25d70):

```python
import os

from pydantic import StrictBool

os.environ['MY_BOOL'] = 'maybe'


class Settings(BaseSettings):
    my_bool: StrictBool | None = None


Settings()
```

raises `json.decoder.JSONDecodeError: Expecting value: line 1 column 1 (char 0)`, rewrapped as `SettingsError: error parsing value for field "my_bool" from source "EnvSettingsSource"`. Pydantic's own error is lost.

`_coerce_env_val_strict` falls back to `json.loads(value)` inside a `try` whose handler only catches `ValidationError`, so a value that is not valid JSON escapes the helper and `PydanticBaseSettingsSource.__call__` rewraps it as a `SettingsError`.

Catching `json.JSONDecodeError` there as well returns the raw string, which is already what the handler does when strict validation fails — the error is deferred to model instantiation. With the change, the snippet above ends with:

```
my_bool
  Input should be a valid boolean [type=bool_type, input_value='maybe', input_type=str]
```

The same path is reachable with `strict=True` on a plain field, with `Literal[<IntEnum member>]` and no model config at all, and through `env_nested_delimiter`.

`test_env_strict_coercion_non_json_value` sits with the existing strict-coercion tests. I reverted the one-line source change and re-ran it: it fails inside the first `pytest.raises(ValidationError)` block with `SettingsError: error parsing value for field "my_bool" from source "EnvSettingsSource"`, and passes again with the change restored.

`make test`, `make lint` and `make mypy` are clean locally, including the without-PyYAML run.

An alternative is to wrap only the `json.loads` call in its own `try` and fall through to the existing `raise`, which is narrower than widening the outer handler. Happy to switch if you prefer that.

### DIFF
--- a/tests/test_settings.py
+++ b/tests/test_settings.py
@@ -3949,6 +3949,40 @@ def test_env_strict_coercion_optional_strict_types(env):
     assert s.my_int == 42
 
 
+def test_env_strict_coercion_non_json_value(env):
+    from pydantic import StrictBool
+
+    class Settings(BaseSettings):
+        my_bool: StrictBool | None = None
+
+    env.set('MY_BOOL', 'maybe')
+    with pytest.raises(ValidationError) as exc_info:
+        Settings()
+    assert exc_info.value.errors(include_url=False) == [
+        {
+            'type': 'bool_type',
+            'loc': ('my_bool',),
+            'msg': 'Input should be a valid boolean',
+            'input': 'maybe',
+        }
+    ]
+
+    class StrictSettings(BaseSettings, strict=True):
+        my_int: int = 0
+
+    env.set('MY_INT', 'not-a-number')
+    with pytest.raises(ValidationError) as exc_info:
+        StrictSettings()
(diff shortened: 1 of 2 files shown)


tests/test_settings.py
```diff
@@ -3949,6 +3949,40 @@ class IntSettings(BaseSettings):
     assert s.my_int == 42
 
 
+def test_env_strict_coercion_non_json_value(env):
+    from pydantic import StrictBool
```
hramezani: please move the import to the top of the file

hramezani:
Thanks @shuvamk for the PR, I left a small comment please address it

hramezani (approved):

hramezani:
I fixed the comment
Thanks @shuvamk

Issue #923, opened 2026-07-24
Nested models are not merged if name/alias differs between sources


aeisenbarth:
When using multiple settings sources, nested settings are usually merged in a way so that the lower priority source provides values missing from the higher priority source. If a nested model is used with an alias, not all sources are taken into account during merge.

## Use case

Aliases seemed like a convenient approach for having a user-friendly configuration file with section names written out in full length, and environment variables abbreviated (since they would become quite long with multiple nesting levels). 
```yaml
service1:
  database:
    authentication:
      some_self_describing_parameter: 1
```
and a shorter environment variable like `S1_DB_AUTH_SDPARAM`.

In order to keep the example short, I use a dotenv file in place of a yaml configuration file.

## Example

In the example, the nested model `submodel` is aliased as `sub`.
Environment variables (higher priority source) using the alias

```.env
SUB_VAR1 = "var1 from env"
```

should be merged with a lower priority file:

```.env
SUBMODEL_VAR1='var1 from dotenv'
SUBMODEL_VAR2='var2 from dotenv'
```
to achieve:
```python
SettingsWithAlias(
    submodel=SubModel(
        var1='var1 from env',
        var2='var2 from dotenv'
    )
)
```

```python
import logging
import os
from pathlib import Path
from pydantic import BaseModel, Field
from pydantic_settings import BaseSettings, SettingsConfigDict

for var in ["SUBMODEL_VAR1", "SUBMODEL_VAR2", "SUB_VAR1", "SUB_VAR2"]:
    os.environ.pop(var, None)

class SubModel(BaseModel):
    var1: str | None = None
    var2: str | None = None

class SettingsWithoutAlias(BaseSettings):
    submodel: SubModel | None = Field(default=None)
    model_config = SettingsConfigDict(env_nested_delimiter="_")

dotenv_file = Path(".env").resolve()
dotenv_file.write_text("SUBMODEL_VAR1='var1 from dotenv'\nSUBMODEL_VAR2='var2 from dotenv'")

os.environ["SUBMODEL_VAR1"] = "var1 from env"

# Create settings where submodel.var1 comes from the higher priority source, submodel.var2 from the lower priority source.
settings_without_alias = SettingsWithoutAlias(_env_file=dotenv_file)
print(settings_without_alias)
# actual: SettingsWithoutAlias(submodel=SubModel(var1='var1 from env', var2='var2 from dotenv'))
assert settings_without_alias.submodel.var1 == "var1 from env"
[...]

PR #933, opened 2026-08-08
docs: clarify alias + nested model env lookup with populate_by_name

pranjalm37:
## Summary

Related to #923.

I investigated #923 (nested model fields not merging across sources when the field uses an `alias`) expecting to find a source-merge-ordering bug. It isn't one: a nested field with an `alias` is looked up, in **every** source (env, dotenv, etc.), using only that alias as the prefix. I confirmed via `_extract_field_info`/`explode_env_vars` that the field's own name is never considered anywhere in the resolution pipeline unless `populate_by_name`/`validate_by_name` is set.

So when the reporter's dotenv file used `SUBMODEL_*` naming for a field aliased `SUB`, those vars were never visible to that field from any source — they weren't "lost in a merge," they were unmatched from the start and fell through as unrelated top-level extras. Setting `populate_by_name=True` (making both the alias and the plain field name valid lookup prefixes) resolves their exact repro correctly. I verified this and left a comment on #923 explaining it.

Since this is working-as-designed alias behavior rather than a bug, I don't think a code change is warranted — changing default alias-resolution semantics here could have surprising effects for other alias users, and isn't my call to make unilaterally. What I think *is* worth fixing is that this is a genuinely easy trap to fall into (I fell into it myself while investigating) and wasn't documented anywhere.

## Change

Adds a `!!! note` with a small runnable example under the alias section in `docs/index.md`, showing the gotcha and the `populate_by_name=True` fix, right where someone setting an alias on a nested field would actually be reading.

## Test plan

- [x] Verified the example's printed output matches real execution exactly.
- [x] `pytest tests/test_docs.py` → my example (`docs/index.md:186-214`... shifts to `:190-...` in the current diff) passes both the `pytest-examples` black-formatting check and its actual runtime output assertion. The only failures are 3 pre-existing, unrelated `ModuleNotFoundError: azure` examples further down the file (missing optional `azure` extra in my local env, not something this PR touches).
- [x] `docs/index.md` isn't part of the repo's `ruff` lint target (`Makefile`'s `sources = pydantic_settings tests`), so no lint step applies beyond the doc-example check above.

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -169,17 +169,19 @@ There are two ways to do this:
 Check the [`Field` aliases documentation](fields.md#field-aliases) for more information about aliases.
 
 !!! note
-    When a nested model field has an alias, environment variables for its sub-fields are looked up using
-    *only* that alias as the prefix — the field's own name is not considered, even for values coming from a
-    dotenv file or other sources. If different sources for the same nested field use different naming (e.g. one
-    source uses the alias, another uses the field name), only the alias-based values will be found; the rest
-    aren't recognized as belonging to that field at all. What then happens to them depends on your `extra`
-    setting: with the default `extra='forbid'`, an unmatched dotenv entry is kept as an extra input and raises
-    a `ValidationError`, while a plain environment variable (not from a dotenv file) is simply not picked up.
-    With `extra='ignore'` both are dropped silently. Either way, the underlying issue is the same: those values
-    were never associated with the field's alias in the first place. Set `populate_by_name=True` (or
-    `validate_by_name=True`) so that both the alias *and* the field's own name are recognized, letting
-    mixed-naming sources merge correctly:
+    Setting an alias on a nested model field *narrows* which environment variables can reach it: its
+    sub-fields are then looked up using only that alias as the prefix, in every source. The field's own name
+    is not an accepted name at all, including for dotenv files and other non-env sources.
+
+    This matters most when sources disagree on naming. If a higher-priority source uses the alias and a
+    lower-priority one uses the field name, only the alias-based values are found — the others were never
+    associated with the field, so there is nothing to merge and the lower-priority values appear to be
+    ignored. What happens to them depends on `extra`: with the default `extra='forbid'` an unmatched dotenv
+    entry is kept as an extra input and raises a `ValidationError`, while a plain environment variable is
+    simply not picked up; with `extra='ignore'` both are dropped silently.
+
+    To accept both spellings, widen the set of accepted names — either globally with `populate_by_name=True`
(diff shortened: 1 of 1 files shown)


hramezani:
This is good, and it's the resolution I want for #923 — thanks for chasing it down and for the two follow-up commits tightening the example cleanup and walking back the "silently missed" wording. That distinction between an unmatched *dotenv* entry (kept as an extra input → `ValidationError` under the default `extra='forbid'`) and an unmatched *plain env var* (simply not picked up) is exactly the sort of thing that makes this trap hard to diagnose, and you got it right. I verified all of it against the library rather than reading it off the PR description:

| claim | result |
|---|---|
| default `extra` is `'forbid'` | ✅ |
| `forbid` + unmatched dotenv entry → `ValidationError` | ✅ `extra_forbidden` at `('submodel_var2',)` |
| `forbid` + unmatched plain env var → not picked up | ✅ `submodel=None`, no error |
| `extra='ignore'` → both dropped silently | ✅ |
| `populate_by_name=True` merges both sources | ✅ output matches exactly |

I've pushed one commit (`93351eb`) rather than sending you round again for it:

**Added an `AliasChoices` example.** The note documented `populate_by_name=True` but not `AliasChoices`, and for the use case that actually motivated the issue — a config file spelling section names out in full (`service1.database.authentication`) alongside abbreviated env vars (`S1_DB_AUTH_SDPARAM`) — `AliasChoices('SUB', 'SUBMODEL')` is usually the better tool. It says per field which spellings are accepted, instead of globally accepting every field name, which is a meaningful difference once a settings class has fields you *don't* want reachable by their own name. Verified it produces byte-identical output to the `populate_by_name` version.

**Restructured the prose.** The original was one dense 11-line paragraph that walked through every failure mode before reaching the fix. Same facts, now split into three: what an alias does (it *narrows* which variables can reach the field), what goes wrong when sources disagree, then the fix. No claims were changed or dropped.

`pytest tests/test_docs.py` → 74 passed (the new example is executed and its `#>` output asserted against real execution); full suite 695 passed, 5 skipped; no leftover `example.env` and no `SUB_*` env leakage into other tests.

Have a look and push back if you'd rather word any of it differently — it's your PR and I'm happy to defer on phrasing. Otherwise I'll merge once CI is green.

For context on the other thread: I closed #934, which attempted this as a behaviour change. It fixed the reported case but inverted alias precedence when both spellings appear in one source, and leaked values between fields whose names and aliases collide. Documenting the existing semantics is the right call here.

hramezani:
@macroscope-app review

hramezani (approved):

hramezani:
Thanks @pranjalm37

PR #934, opened 2026-08-08
Ensure nested models aliased by plain strings are also reachable by their field name (#923)

### LINKED ISSUE #923: Nested models are not merged if name/alias differs between sources

MRoshaan:
Fixes #923

## Change

A nested model field aliased by a plain string (`Field(alias='SUB')`) is currently only reachable under the alias-based prefix in every settings source. A lower-priority source that uses the field's real name (e.g. a `.env` file with `SUBMODEL_VAR1=...`) then never contributes to that field: its values fall through to top-level extras and are never merged into the aliased nested model.

This PR makes such aliased complex fields **also resolvable by their real field name**, so both sources produce a value for the same field and the existing `deep_update` merge combines them:

```python
class SubModel(BaseModel):
    var1: str | None = None
    var2: str | None = None

class SettingsWithAlias(BaseSettings):
    submodel: SubModel | None = Field(alias='SUB', default=None)
    model_config = SettingsConfigDict(env_nested_delimiter='_')

# env:  SUB_VAR1=var1 from env
# dotenv: SUBMODEL_VAR1=var1 from dotenv, SUBMODEL_VAR2=var2 from dotenv
# result: submodel=SubModel(var1='var1 from env', var2='var2 from dotenv')
```

The change is scoped narrowly in `_extract_field_info` to nested `BaseModel` / pydantic-dataclass fields (including `Optional[...]`). Scalar fields keep their existing alias-only input behavior.

## Notes for reviewers

- The merge path in `_settings_build_values` is untouched: sources still hand back plain dicts keyed by the (uniquely normalized) alias, so `deep_update` merges as before.
- `populate_by_name=True` / explicit `AliasChoices` already worked — this instead makes the plain-string-alias case behave the same way automatically, without requiring those settings.
- Covered by tests in `tests/test_precedence_and_merging.py` for both merged split-across-sources and dotenv-only field-name sourcing, plus a guard that scalar aliases are not broadened to their field name.

## Checklist

- [x] Tests added / updated
- [x] `ruff check` and `ruff format --check` pass
- [x] `mypy pydantic_settings` passes for the changed module
- [x] Full test suite passes (613 passed, 14 skipped)

### DIFF
--- a/pydantic_settings/sources/base.py
+++ b/pydantic_settings/sources/base.py
@@ -14,6 +14,7 @@ from pydantic._internal._typing_extra import (  # type: ignore[attr-defined]
     get_origin,
 )
 from pydantic._internal._utils import deep_update, is_model_class
+from pydantic.dataclasses import is_pydantic_dataclass
 from pydantic.fields import FieldInfo
 from typing_inspection.introspection import is_union_origin
 
@@ -250,6 +251,16 @@ def _unwrap_optional_annotation(annotation: Any) -> Any:
     return annotation
 
 
+def _nested_model_annotation(annotation: Any) -> bool:
+    """Whether ``annotation`` describes a nested model or pydantic dataclass.
+
+    ``Optional[T]`` is unwrapped so that a field aliased by a plain string keeps being
+    reachable by its real field name even when the field is optional (see #923).
+    """
+    annotation = _unwrap_optional_annotation(annotation)
+    return is_model_class(annotation) or is_pydantic_dataclass(annotation)
+
+
 def _has_discriminator(field_info: FieldInfo) -> bool:
     """Check if a field uses a discriminated union (via Annotated or Field(discriminator=...))."""
     if field_info.discriminator is not None:
@@ -472,6 +483,17 @@ class PydanticBaseEnvSettingsSource(PydanticBaseSettingsSource):
                 field_info.append((field_name, self._apply_case_sensitive(env_prefix + field_name), True))
             else:
                 field_info.append((field_name, self._apply_case_sensitive(env_prefix + field_name), False))
+        elif isinstance(v_alias, str) and _nested_model_annotation(
+            _strip_annotated(_resolve_type_alias(field.annotation))
+        ):
+            # A nested model field aliased by a plain string must also be reachable by
+            # its real field name so that lower-priority sources (e.g. a config file or
+            # another env source) which use the field name instead of the alias can
+            # populate -- and be merged with -- the same nested model (see #923).
+            # This is scoped to complex model/dataclass fields in order to leave scalar
+            # alias input behavior unchanged.
+            name_prefix = self.env_prefix if self.env_prefix_target in ('variable', 'all') else ''
+            field_info.append((field_name, self._apply_case_sensitive(name_prefix + field_name), False))
 
         return field_info
 
--- a/tests/test_precedence_and_merging.py
+++ b/tests/test_precedence_and_merging.py
@@ -3,7 +3,7 @@ from __future__ import annotations as _annotations
 from pathlib import Path
 from typing import Literal
 
-from pydantic import AliasChoices, AliasGenerator, AnyHttpUrl, Field
+from pydantic import AliasChoices, AliasGenerator, AnyHttpUrl, BaseModel, Field
 from pydantic.alias_generators import to_camel
 
 from pydantic_settings import (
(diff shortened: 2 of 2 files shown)


hramezani:
Thanks for picking this up, @MRoshaan — and for keeping it focused rather than reaching for the merge path. I checked it out and ran it against the original repro plus some surrounding cases. It does fix the reported case, but I found two regressions that I don't think we can take, and I've landed on docs as the right resolution here. Details below so the work isn't a black box.

## It does fix the narrow case

The issue's exact snippet gives `SubModel(var1='var1 from env', var2='var2 from dotenv')` on your branch, and the full suite passes (697 passed, 5 skipped).

## Regression 1: alias precedence inverts

When both spellings are present in the *same* source, the alias should win. On this branch it doesn't:

```python
class Settings(BaseSettings):
    submodel: SubModel | None = Field(alias='SUB', default=None)
    model_config = SettingsConfigDict(env_nested_delimiter='_', extra='ignore')

# env: SUB_VAR1='from ALIAS'  SUBMODEL_VAR1='from FIELDNAME'
```

```
2.15.0 : submodel=SubModel(var1='from ALIAS',     var2=None)
this PR: submodel=SubModel(var1='from FIELDNAME', var2=None)
```

The reason the scalar path is safe but this one isn't: `EnvSettingsSource.get_field_value` (`env.py:110`) iterates `_extract_field_info` and **breaks on first match**, so appending the field-name entry after the alias is harmless there. The nested-explosion path (`env.py:288`) instead collects *all* entries into a set of prefixes and merges what each one yields — no first-match semantics, so the appended field-name entry silently overrides the explicit alias.

## Regression 2: cross-field value leakage

Because the field name is registered unconditionally, one field's name can collide with another field's alias. Distinct types, no shared model:

```python
class Settings(BaseSettings):
    db:      Db    | None = Field(alias='PRIMARY', default=None)
    primary: Cache | None = Field(alias='DB',      default=None)
    model_config = SettingsConfigDict(env_nested_delimiter='_', extra='ignore')

# env: PRIMARY_HOST=h1  PRIMARY_PORT=5432
```

```
2.15.0 : db=Db(host='h1', port=5432)  primary=None
this PR: db=Db(host='h1', port=5432)  primary=Cache(host='h1', ttl=None)
```

`PRIMARY_*` now populates both fields. That's a silent wrong-value bug, which is worse than the silent drop the issue was originally about. Neither regression is covered by the added tests.

## The underlying design problem

This is the part that decides it for me, and it isn't really about the two bugs above — those are symptoms.

As I noted [above](https://github.com/pydantic/pydantic-settings/issues/923#issuecomment-5166247482), `alias='SUB'` making `SUB` the only accepted input name is ordinary pydantic alias behaviour, not something settings-specific. This PR widens the accepted-name set in `_extract_field_info` so the existing merge happens to fire — which is exactly what `populate_by_name=True` already does explicitly, except applied unconditionally and only to nested models.

The result is that `Field(alias='SUB')` would silently mean "SUB or submodel" for nested models but strictly "SUB" for scalars. Your own test asserts `Settings(submodel={'var1': 'x'})` still raises `extra_forbidden` — so init kwargs stay alias-only while env/dotenv don't. Same field, two different alias semantics depending on which source the value arrives from. I don't think we can ship that divergence from pydantic proper, and the scoping guard I'd otherwise ask for is "gate it on `populate_by_name`/`validate_by_name`" — at which point the change is redundant with what already works.

## Where I'm landing

Docs. #933 already calls out the gotcha, and I'd rather invest there than in behaviour change. What's genuinely missing from the docs — and what I don't think anyone could work out from first principles — is that setting an alias on a nested field **narrows** which source keys can reach it, so a lower-priority source using the field's real name contributes nothing and looks silently ignored. Paired with the two working spellings (`populate_by_name=True`, or `AliasChoices('SUB', 'SUBMODEL')` when you want each level to accept both explicitly), that covers the real use case in the issue.

If we ever do want the merge itself to work across differing spellings, I still think it needs the pre-merge key-to-field resolution I sketched in the issue — handling `AliasPath`, `AliasChoices`, per-source alias handling and case-insensitivity — rather than widening accepted names. That's a much bigger change and I'm not asking you to take it on here.

So I'm going to close this in favour of the docs route. Genuinely sorry to send you away empty-handed after real work — the investigation was sound and the `_extract_field_info` scoping instinct was the right place to look. If you'd like to take the docs change instead, that one I'd merge happily.

One unrelated thing I turned up while testing, so it doesn't get lost: `env_prefix` combined with a nested alias looks broken independently of this PR. `APP_SUB_VAR1` with `env_prefix='APP_'` and `alias='SUB'` gives `submodel=None` on 2.15.0 *and* on your branch. The alias branch uses `env_prefix_target in ('alias', 'all')` while explosion uses the `'variable'` target. I'll open a separate issue for it.

MRoshaan:
Thanks for the thorough review, @hramezani — the two regressions are clearly demonstrated, and your point about diverging from pydantic's alias semantics when the accepted-name set is widened unconditionally is well taken. The `_extract_field_info` scoping instinct was the right lead, but you're right that the cleaner resolution is documentation.

I'll close this PR and take the docs route instead: #933 already covers the gotcha with `populate_by_name=True`, so I'll add the missing piece you identified — that an alias on a nested field narrows which source keys can reach it, and the explicit `AliasChoices('SUB', 'SUBMODEL')` spelling for when you want each level to accept both. Thanks again for the detailed write-up.

MRoshaan:
Closing in favor of the docs route (per maintainer feedback). The docs addition covering the AliasChoices spelling will be opened as a separate PR.

MRoshaan:
Follow-up: opened #939 with the docs change covering the narrowing behavior and the `AliasChoices('SUB', 'SUBMODEL')` spelling. Thanks again for the guidance.

PR #938, opened 2026-08-14
fix: do not leak SecretStr values in ValidationError

### LINKED ISSUE #937: SecretsSettingsSource leaks to ValidationError messages
When a ValidationError occurs during instantiation of the settings class, the error message contains all values including those that were supposed to be instantiated as `SecretStr`.

## Cause

When instantiating a Pydantic model, a potential validation error will contain the input values.

SecretsSettingsSource and NestedSecretsSettingsSource pass an exploded dictionary as input values:

https://github.com/pydantic/pydantic-settings/blob/77384c93cd7643264f799a741d38d949ea622a71/pydantic_settings/main.py#L262

This is a plain dictionary containing secrets as strings. Therefore, they are printed in the exception message.

## Test case

```python
import pytest
from pathlib import Path
from pydantic import SecretStr, ValidationError
from pydantic_settings import BaseSettings, SettingsConfigDict, PydanticBaseSettingsSource

def test_pydantic_settings_prints_secrets(tmp_path: Path):
    from pydantic import SecretStr, ValidationError
    from pydantic_settings import BaseSettings, SettingsConfigDict, PydanticBaseSettingsSource

    the_password = "super_secret"
    (tmp_path / "password").write_text(the_password)

    class MyModel(BaseSettings):
        the_username: str
        password: SecretStr
        model_config = SettingsConfigDict(secrets_dir=tmp_path)

        @classmethod
            cls,
            return (
                init_settings,
                file_secret_settings,
            )

    # Correctly reading all required fields
    my_model = MyModel(the_username="my_user") # ok
    assert my_model.password.get_secret_value() == the_password

    # Required fields missing (not necessarily an application bug like here, but they could be missing from a configuration file).
    that_it_does_not_contain_password = rf"^((?!{the_password}).)*$"
    with pytest.raises(ValidationError, match=that_it_does_not_contain_password):
        MyModel()
```

## Actual behavior

The validation error contains secrets in plain text:

```text
pydantic_core._pydantic_core.ValidationError: 1 validation error for MyModel
the_username
  Field required [type=missing, input_value={'password': 'super_secret'}, input_type=dict]
    For further information visit https://errors.pydantic.dev/2.13/v/missing
```

## Expected behavior

The validation error should not show secrets in plain text.

Currently SecretsSettingsSource (and NestedSecretsSettingsSource) postpone Pydantic class instantiation to the above cited line. Probably, they should instead already wrap values in SecretStr. 

A possible problem would be mis-use of secret sources for non-secret values. For example if a secret file contains a JSON structure (as mentioned in issue https://github.com/pydantic/pydantic-settings/issues/154#issuecomment-1697588043) and if the JSON structure contains fields that are another type than SecretStr.

DSeaStar:
## Summary

Fixes #937.

`SecretsSettingsSource` loaded secret files as plain strings and passed them into `BaseModel.__init__`. When validation failed for another reason (e.g. a required field missing), `ValidationError` included the secret in `input_value`:

```
input_value={'password': 'super_secret'}
```

This wraps `SecretStr` / `SecretBytes` fields at load time so pydantic redacts them in the error. Fields that are not secret types (including JSON-shaped secrets used as nested models) stay plain strings.

## Test plan

- [x] `test_secrets_source_does_not_leak_secretstr_in_validation_error` — the #937 case
- [x] `test_secrets_path` / `test_secrets_path_utf8` / `test_secrets_path_json` — existing secrets_dir behavior unchanged

### DIFF
--- a/pydantic_settings/sources/providers/secrets.py
+++ b/pydantic_settings/sources/providers/secrets.py
@@ -6,11 +6,16 @@ import logging
 import os
 import warnings
 from pathlib import Path
+from types import UnionType
 from typing import (
     TYPE_CHECKING,
     Any,
+    Union,
+    get_args,
+    get_origin,
 )
 
+from pydantic import SecretBytes, SecretStr
 from pydantic.fields import FieldInfo
 
 from pydantic_settings.utils import _settings_debug_enabled, logger, path_type_label
@@ -22,6 +27,33 @@ from ..utils import InitState
 
 if TYPE_CHECKING:
     from pydantic_settings.main import BaseSettings
+
+
+def _secret_type(annotation: Any) -> type[SecretStr | SecretBytes] | None:
+    """Return SecretStr/SecretBytes if *annotation* is one of those types (or Optional/Annotated of them)."""
+    if annotation is SecretStr:
+        return SecretStr
+    if annotation is SecretBytes:
+        return SecretBytes
+    origin = get_origin(annotation)
+    if origin is Union or origin is UnionType:
+        args = [arg for arg in get_args(annotation) if arg is not type(None)]
+        if len(args) == 1:
+            return _secret_type(args[0])
+    if origin is not None and getattr(origin, '__name__', None) == 'Annotated':
+        args = get_args(annotation)
+        if args:
+            return _secret_type(args[0])
+    return None
+
+
+def _wrap_secret_value(field: FieldInfo, value: str) -> Any:
+    secret_cls = _secret_type(field.annotation)
+    if secret_cls is SecretStr:
+        return SecretStr(value)
+    if secret_cls is SecretBytes:
+        return SecretBytes(value.encode())
+    return value
 
 
 class SecretsSettingsSource(PydanticBaseEnvSettingsSource):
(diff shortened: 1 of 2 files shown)


hramezani:
Thanks for putting this together, @DSeaStar — and sorry for the timing, since you'd clearly already done the work by the time we wrote up the analysis on #937.

We've closed #937 as a duplicate of [pydantic#9139](https://github.com/pydantic/pydantic/issues/9139): the leak isn't specific to `SecretsSettingsSource`, and we don't think it can be properly fixed at the source level. I checked this branch out and ran it to be sure the reasoning holds against the real implementation rather than a hypothetical:

```
A) reported case (#937)   fixed  ✅
B) error on secret field  still leaks  ❌
C) env / dotenv / json    still leaks  ❌
D) before-validators      regressed  ❌
```

**B — the secret field's own errors still leak.** Wrapping at load time doesn't help when the failure is on the secret itself, since the raw value is still the validation input:

```python
class B(BaseSettings):
    password: SecretStr = Field(min_length=100)
    model_config = SettingsConfigDict(secrets_dir=tmp)
```
```
Value should have at least 100 items after validation, not 5 [type=too_short, input_value='short', ...]
```

**C — other sources are untouched.** `secrets_dir` is only one of the paths into the same plain dict; env, dotenv, and file sources leak exactly as before. Extending this approach would mean repeating the wrapping in every source and still not fixing B.

**D — this breaks `mode='before'` validators**, which is the High-severity finding from the automated review, and it does reproduce:

```python
@field_validator("token", mode="before")
def check(cls, v):
    return v.startswith("prefix_")   # AttributeError: 'SecretStr' object has no attribute 'startswith'
```

Any existing code doing string work in a before-validator on a secret-file field would break on upgrade. There's also the Critical `TypeAliasType` gap, where a secret behind an alias silently stays unwrapped — the failure mode being that it looks fixed while still leaking.

So this trades a partial fix for a real regression, and the remaining vectors mean users could reasonably believe secrets are safe when they aren't. That "looks fixed but isn't" property is what makes us want to keep it upstream, where redaction can happen at error-rendering time and cover every source and every error type at once. Closing on those grounds — no reflection on the quality of the patch, which was clean and well-tested for the case it targeted.

If you'd like to keep pushing on this, [pydantic#9139](https://github.com/pydantic/pydantic/issues/9139) is the place, and a repro showing the settings angle would genuinely strengthen the case there. The `hide_input_in_errors=True` workaround is on #937, with the caveat that it doesn't cover `.errors()` / `.json()`.

PR #939, opened 2026-08-15
docs: show AliasChoices for nested fields with an alias

MRoshaan:
## Summary

Follow-up to #923 and the closed #934. When a nested model field has an alias, that alias **narrows** which
source keys can reach it ,the field's sub-keys are only looked up under the alias prefix, so a lower-priority
source using the field's real name contributes nothing and looks silently ignored.

#933 already covers the gotcha and the `populate_by_name=True` fix. This PR adds the complementary spelling:
using `AliasChoices('SUB', 'SUBMODEL')` to make each nesting level accept both the alias and the field's own
name explicitly, with a runnable example showing the env + dotenv merge working.

## Change

- `docs/index.md`: short explanation of the narrowing behavior plus a `AliasChoices` example that merges
  `SUB_VAR1` (env, higher priority) with `SUBMODEL_VAR2` (dotenv, lower priority).

## Test plan

- Verified the example output on `2.15.0` (installed from this branch's fork): prints
  `{'submodel': {'var1': 'var1 from env', 'var2': 'var2 from dotenv'}}`.
- Confirmed the example passes the `pytest-examples` black-format check (`string_normalization=False`,
  line-length 88) unchanged.

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -168,6 +168,41 @@ There are two ways to do this:
 
 Check the [`Field` aliases documentation](fields.md#field-aliases) for more information about aliases.
 
+When a nested model field has an alias, that alias **narrows** which source keys can reach it: the field's
+sub-keys are looked up only under the alias prefix, so a lower-priority source that uses the field's real name
+contributes nothing and its values are silently ignored. When you want each nesting level to accept *both* the
+alias and the field's own name explicitly, use [`AliasChoices`][pydantic.AliasChoices] instead of a single alias:
+
+```py
+import os
+from pathlib import Path
+
+from pydantic import AliasChoices, BaseModel, Field
+
+from pydantic_settings import BaseSettings, SettingsConfigDict
+
+
+class SubModel(BaseModel):
+    var1: str | None = None
+    var2: str | None = None
+
+
+class Settings(BaseSettings):
+    model_config = SettingsConfigDict(env_nested_delimiter='_')
+    submodel: SubModel | None = Field(
+        validation_alias=AliasChoices('SUB', 'SUBMODEL'), default=None
+    )
+
+
+dotenv_file = Path('example.env')
+dotenv_file.write_text("SUBMODEL_VAR2='var2 from dotenv'")
+os.environ['SUB_VAR1'] = 'var1 from env'
+
+print(Settings(_env_file=dotenv_file).model_dump())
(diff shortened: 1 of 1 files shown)


hramezani:
Thanks @MRoshaan — and apologies, this overlap is my fault rather than yours.

When I closed #934 I pointed you at the docs route and specifically named `AliasChoices` as the better fit for the reporter's use case, without mentioning that I'd just pushed that same example to #933. You then wrote it independently, and the two examples turned out byte-identical — I diffed them, right down to the line-wrapped `Field(...)` call and the cleanup lines. That's a duplicated effort I caused by not saying what was already in flight.

Two things from your version were genuinely better than mine, and I've pushed them to #933 in `2c0d685` crediting you:

- **The `[`AliasChoices`][pydantic.AliasChoices]` cross-reference.** This file already uses that mkdocs style at line 95 and I'd left mine as a plain code span. Yours links to the API docs; mine didn't.
- **The "each nesting level" framing.** You noted `AliasChoices` lets each level accept both spellings explicitly, which is precisely the property that matters for the multi-level config in #923 (`service1.database.authentication` vs `S1_DB_AUTH_SDPARAM`). My phrasing only made the per-field-vs-global distinction and left that implicit.

So I'm closing this as a duplicate of #933, which is the superset — it additionally documents the `extra='forbid'` vs `extra='ignore'` behaviour and the dotenv-vs-plain-env asymmetry that makes this trap hard to diagnose in the first place.

For what it's worth, your read on this issue has been correct throughout. The narrowing analysis in #934 was right about the mechanism even though the fix didn't work out, and you landed on the same documentation shape I did. Sorry for the wasted round trip on the sequencing — that one's on me.

hramezani:
Closing as a duplicate — #933 is merged as e2f3d01, and both of the things your version did better are on `main` now:

- the `[`AliasChoices`][pydantic.AliasChoices]` cross-reference (`docs/index.md:185`)
- the "each nesting level" framing (`docs/index.md:220`)

They went in as 2c0d685 on #933 with credit to you.

To restate what I said above so it isn't buried: the duplication here was caused by me. When I closed #934 I pointed you at the docs route and named `AliasChoices` specifically, without mentioning I'd already pushed that example to #933 — so you wrote it independently and the two came out byte-identical. Thanks for the work on this, and sorry for the wasted round trip.

PR #944, opened 2026-08-21
fix(types): include DotenvFiltering in __all__ in sources.types

aoright:
### Summary
- Add `DotenvFiltering` to `__all__` in `pydantic_settings.sources.types` to match its definition and re-export in `pydantic_settings.sources`.


hramezani (approved):

hramezani:
Thanks @aoright

PR #945, opened 2026-08-21
fix: use self.cli_prefix instead of cli_prefix in CliSettingsSource constructor

RomainL972:
Currently, the `CliSettingsSource` constructor uses the variable `cli_prefix` in its constructor (line 426) where it might be None, instead of using `this.cli_prefix`.

This causes an error with the following code:

```python
class SomethingSettings(BaseSettings):
    url: str

    model_config = SettingsConfigDict(cli_parse_args=True, cli_prefix="some_prefix")

print(CliApp.format_help(SomethingSettings))
```


hramezani:
Thanks @RomainL972 
I pushed a commit and added a test for it.

hramezani (approved):

PR #947, opened 2026-08-27
fix: treat empty JSON config file as empty mapping instead of crashing

sneha4175:
## Problem

`JsonConfigSettingsSource` crashes on an empty (or whitespace-only) JSON config file, while the other file-based sources handle it gracefully:

```python
JsonConfigSettingsSource._read_file  # json.load('') -> JSONDecodeError
YamlConfigSettingsSource._read_file  # yaml.safe_load(...) or {}  -> {}
TomlConfigSettingsSource._read_file  # tomllib.load(empty)        -> {}
```

An empty file passes the `is_file()` check in `ConfigFileSourceMixin._read_files` and is treated as a real source, so `json.load()` is called on empty content and raises:

```python
from pydantic_settings.sources import JsonConfigSettingsSource

# empty config.json on disk
class Settings(BaseSettings):
    model_config = SettingsConfigDict(json_file='config.json')
    @classmethod
    def settings_customise_sources(cls, settings_cls, *args, **kwargs):
        return (JsonConfigSettingsSource(settings_cls),)

Settings()  # json.decoder.JSONDecodeError: Expecting value: line 1 column 1 (char 0)
```

YAML has a dedicated `test_yaml_empty_file`; JSON had no such coverage and crashed.

## Root cause

`_read_file` in `pydantic_settings/sources/providers/json.py` called `json.load(json_file)` with no guard for empty content.

## Fix

Read the file content and fall back to an empty mapping when it is empty or whitespace-only, so an empty JSON config falls back to defaults — matching `YamlConfigSettingsSource` (`... or {}`) and `tomllib`. This is the JSON analog of #914, which applied the same "empty → empty mapping" robustness fix to the YAML source.

## Testing

- Added `test_json_empty_file` (parametrized over empty and whitespace-only content), mirroring `test_yaml_empty_file`. It fails before the fix (`JSONDecodeError`) and passes after.
- `uv run pytest` — 723 passed, 5 skipped.
- `uv run ruff check` / `ruff format --check` — clean.
- `make mypy` — success, no issues.


hramezani (approved):

hramezani:
Thanks @sneha4175

PR #949, opened 2026-08-28
Add CliVariadicArg for named variadic CLI options

### LINKED ISSUE #739: Named arguments cannot be variadic.
Positional arguments can be made variadic as of #518, but this is not possible for named arguments.

```py
class RootCommand(
    BaseSettings,
    cli_parse_args=True,
    cli_implicit_flags='toggle',
):
    param: list[str] = Field(default_factory=list)

    def cli_cmd(self):
        print('------pydantic-settings-------')
        print(self)
        print('------pydantic-settings help-------')
        CliSettingsSource[argparse.ArgumentParser](RootCommand).root_parser.print_help()


    CliApp.run(RootCommand, cli_args=args)


def mainA():
    print('------------argparse-------------')
    parser = argparse.ArgumentParser()
    parser.add_argument('--param', type=str, nargs='*')
    print(parser.parse_args())
    print('------------argparse help-------------')
    parser.print_help()


if __name__ == '__main__':
    mainA()
    print('----------------------------------')
    main()
```

```
python app.py --param a,b
------------argparse-------------
Namespace(param=['a,b'])
------------argparse help-------------
usage: app.py [-h] [--param [PARAM ...]]

options:
  -h, --help           show this help message and exit
  --param [PARAM ...]
----------------------------------
------pydantic-settings-------
param=['a', 'b']
------pydantic-settings help-------
usage: app.py [-h] [--param list[str]]

options:
  -h, --help         show this help message and exit
  --param list[str]  (default factory: list)
```

```
$ python app.py --param a b
------------argparse-------------
Namespace(param=['a', 'b'])
------------argparse help-------------
usage: app.py [-h] [--param [PARAM ...]]

options:
  -h, --help           show this help message and exit
  --param [PARAM ...]
----------------------------------
usage: app.py [-h] [--param list[str]]
app.py: error: unrecognized arguments: b
```

gyanu2507:
Positional lists already use `nargs='*'`, but a named `list[str]` field is still a repeated option (`--param a --param b`). There was no way to accept `--param a b c`.

`CliVariadicArg` is the named equivalent. Default list handling is unchanged so existing CLIs keep `action=append`.

Fixes #739.


pydantic_settings/sources/providers/cli.py
```diff
@@ -1199,9 +1205,12 @@ def _add_parser_args(  # noqa: C901
...
         if is_append_action:
-            kwargs['action'] = 'append'
-            if _annotation_contains_types(field_info.annotation, (dict, Mapping), is_strip_annotated=True):
-                self._cli_dict_args[kwargs['dest']] = field_info.annotation
+            if _CliVariadicArg in field_info.metadata:
+                kwargs['nargs'] = '*'
+            else:
+                kwargs['action'] = 'append'
+                if _annotation_contains_types(field_info.annotation, (dict, Mapping), is_strip_annotated=True):
+                    self._cli_dict_args[kwargs['dest']] = field_info.annotation
```
hramezani: Three issues come out of this branch, all because `nargs='*'` is set without `action='append'`, and because the `_cli_dict_args` registration moved inside the `else`.

**1. `CliPositionalArg[CliVariadicArg[...]]` silently breaks (high).**

`_convert_positional_arg` branches on `kwargs.get('action') == 'append'` to decide `nargs='+'`/`'*'`. For a variadic-wrapped positional that check is now `False`, so control falls to `elif not is_required: kwargs['nargs'] = '?'`, clobbering the `'*'` set here:

```py
class S(BaseSettings, cli_parse_args=True):
    param: CliPositionalArg[CliVariadicArg[list[str]]] = Field(default_factory=list)

CliApp.run(S, cli_args=['a', 'b', 'c'])
# usage: t.py [-h] [PARAM]
# error: unrecognized arguments: b c
```

Baseline `CliPositionalArg[list[str]]` on `main` correctly returns `{'param': ['a', 'b', 'c']}`. The required variant also loses the argparse-level requirement — `main` raises argparse's "the following arguments are required: PARAM" via `nargs='+'`, this branch raises a pydantic `Field required` instead.

Either set `action='append'` here as well, or teach `_convert_positional_arg` to respect an already-set `nargs`.

**2. dict/Mapping fields produce a list instead of a dict (high).**

Moving `self._cli_dict_args[kwargs['dest']] = ...` into the `else` means a variadic dict field is never registered, so `_get_merge_parsed_list_types` falls back to `merge_type = list` and merges the values into a JSON array:

```py
class S(BaseSettings, cli_parse_args=True):
    param: CliVariadicArg[dict[str, int]] = Field(default_factory=dict)

CliApp.run(S, cli_args=['--param', '{"a":1}', '{"b":2}'])
# ValidationError: Input should be a valid dictionary [input_value=[{'a': 1}, {'b': 2}]]
```

Same for unions containing a dict — `CliVariadicArg[Union[dict[str, int], list[int], None]]` with `--param '{"a":1}'` fails both union members, where the unwrapped annotation on `main` gives `{'param': {'a': 1}}`.

`is_append_action` includes `dict`/`Mapping`, so these annotations do reach this branch. Either keep the `_cli_dict_args` registration for the variadic path too, or reject `CliVariadicArg` on Mapping fields with a `SettingsError`.

**3. Silently ignored on non-collection fields (medium).**

This is gated behind `if is_append_action`, so `CliVariadicArg[str]` / `CliVariadicArg[int]` fall through with no error and no effect:

```py
class S(BaseSettings, cli_parse_args=True):
    param: CliVariadicArg[str] = 'x'

CliApp.run(S, cli_args=['--param', 'a'])   # {'param': 'a'} — annotation had no effect
```

Every other `Cli*` annotation validates its target (`_verify_cli_flag_annotations` raises for non-bool `CliImplicitFlag`, `_sort_arg_fields` raises for bad `CliSubCommand`). Someone typo'ing `CliVariadicArg[str]` for `CliVariadicArg[list[str]]` currently gets no feedback — worth raising a `SettingsError` when `_CliVariadicArg` is present but the annotation isn't a list/set/Sequence.

pydantic_settings/sources/providers/cli.py
```diff
@@ -1199,9 +1205,12 @@ def _add_parser_args(  # noqa: C901
...
```
hramezani, line 1217: Minor, and partly pre-existing: `CliVariadicArg` is also silently ignored on `AliasPath` fields, because `_add_parser_alias_paths` unconditionally overwrites `kwargs` with `action='append'`, discarding the `nargs` set just above.

```py
class S(BaseSettings, cli_parse_args=True):
    param: CliVariadicArg[list[str]] = Field(
        default_factory=list, validation_alias=AliasPath('p', 0)
    )

CliApp.run(S, cli_args=['--p', 'a', 'b'])
# usage: t.py [-h] [--p list]
# error: unrecognized arguments: b
```

The structure predates this PR, but the new annotation now claims to support something that doesn't work here. Either support it or note the limitation in the docs.

docs/index.md
```diff
@@ -1136,6 +1136,9 @@ not required, set the `is_required` flag to `False` to disable raising an error
 !!! note
     `CliSubCommand` and `CliPositionalArg` are always case sensitive.
 
+List fields are repeated options by default (`--param a --param b`). Wrap the field in
+`CliVariadicArg` to accept remaining values after a single option (`--param a b c`).
+
```
hramezani: This paragraph lands inside "Subcommands and Positional Arguments", directly above a `CliSubCommand`/`CliPositionalArg` example it has nothing to do with — `CliVariadicArg` is a named-option feature, so it reads as though it applies to the example below it.

Two behaviours users will hit are also worth documenting:

- `nargs='*'` is greedy and swallows a following subcommand name:
  ```py
  CliApp.run(S, cli_args=['--param', 'a', 'b', 'sub', '--y', '3'])
  # error: argument {sub}: invalid choice: '3'
  ```
- Repeating the option **replaces** rather than appends (`['--param', 'a', 'b', '--param', 'c']` → `['c']`), which is the opposite of the `action='append'` behaviour this paragraph contrasts against.

tests/test_source_cli.py
```diff
@@ -1901,6 +1902,15 @@ class MainOptional(MainRequired):
     assert CliApp.run(MainRequired, cli_args=['7', '8', '9']).model_dump() == {'values': [7, 8, 9]}
 
 
+def test_cli_variadic_named_arg():
+    class Settings(BaseSettings):
+        model_config = SettingsConfigDict(cli_parse_args=True)
+        param: CliVariadicArg[list[str]] = Field(default_factory=list)
+
+    assert CliApp.run(Settings, cli_args=[]).model_dump() == {'param': []}
```
hramezani, line 1916: This covers only the happy path. Worth adding cases for the combinations that currently fail silently: positional + variadic, dict/Mapping annotations, non-collection annotations, required fields, and repeated options.

The new `CliVariadicArg is not outermost annotation` error added in `cli.py` is also currently untested.

hramezani (commented):
Thanks for the PR! The core idea works, but the `nargs='*'` branch bypasses two code paths that the `action='append'` branch was feeding, which makes a few annotation combinations fail silently. Details inline — the first two are the ones I'd consider blocking.

All examples below were run against this branch, with `main` as the baseline for comparison.

hramezani (commented):
Re-reviewed at `34a5224` — all previous comments are addressed. I re-ran every reproduction from the first review: positional + variadic (optional and required), dict and union-with-dict fields, non-collection rejection, and the docs/tests updates all check out, and the full test suite passes locally. Thanks for the thorough follow-up!

One small new edge case found while re-testing, noted inline — not blocking from my side.

pydantic_settings/sources/providers/cli.py
```diff
@@ -1193,8 +1199,15 @@ def _add_parser_args(  # noqa: C901
...
 
     def _convert_append_action(self, kwargs: dict[str, Any], field_info: FieldInfo, is_append_action: bool) -> None:
+        if _CliVariadicArg in field_info.metadata and not is_append_action:
+            raise SettingsError(
+                f'CliVariadicArg requires a list, set, dict, Sequence, or Mapping type for {kwargs["dest"]}'
+            )
         if is_append_action:
-            kwargs['action'] = 'append'
+            if _CliVariadicArg in field_info.metadata:
+                kwargs['nargs'] = '*'
```
hramezani, line 1212: One remaining rough edge: `nargs='*'` legitimately permits zero values, but passing the bare option dies with an obscure error:

```py
class Settings(BaseSettings, cli_parse_args=True):
    param: CliVariadicArg[list[str]] = Field(default_factory=list)

CliApp.run(Settings, cli_args=['--param'])
# SettingsError: error parsing value for field "param" from source "CliSettingsSource"
# (underlying: json.decoder.JSONDecodeError: Expecting value: line 1 column 1 (char 0))
```

argparse stores `[]` for the dest, `_merge_parsed_list([])` merges it to `''`, and that empty string then fails JSON decoding in `decode_complex_value`. It should produce `[]` (i.e. `--param` with no values behaves like the option was given an empty list).

This can't happen on `main` since `action='append'` requires one value per occurrence (`argument --param: expected one argument`), so it's new surface introduced by `nargs='*'`. Likely a small fix in the merge path — return `'[]'` when the parsed list is empty — plus a test for `cli_args=['--param']`.

pydantic_settings/sources/providers/cli.py
hramezani: All three addressed in the update — verified at 34a5224 by re-running the reproductions: `CliPositionalArg[CliVariadicArg[list[str]]]` parses correctly (and the required variant restores the argparse-level `the following arguments are required: PARAM` error, matching the `main` baseline), dict/union-with-dict fields merge into dicts again, and `CliVariadicArg[str]` now raises a clear `SettingsError`. Thanks!

pydantic_settings/sources/providers/cli.py
hramezani: Addressed — the `AliasPath` limitation is now documented ("CliVariadicArg is ignored on AliasPath fields"), which was one of the two options suggested. Fine to leave actual support as a future improvement.

docs/index.md
hramezani: Addressed — the docs now live in their own "Variadic named options" section, and both caveats (repeat-replaces and greedy `nargs='*'` swallowing a following subcommand) are documented. Verified both behaviors still match the docs at 34a5224.

tests/test_source_cli.py
hramezani: Addressed — the new tests cover positional + variadic (optional and required), dict fields, non-collection rejection, repeated options, and the not-outermost error. One remaining uncovered case noted in my new comment on `_convert_append_action`: `cli_args=['--param']` with zero values.

pydantic_settings/sources/providers/cli.py
gyanu2507: Fixed in 9553d8e.

`_merged_list_to_str` now returns `'[]'` when nothing was consumed. The empty case has no `cli_arg` to infer decoding from, so `is_use_decode` stayed `None` and the function fell through to the un-bracketed `''`.

`--param` with no values now gives `[]`, and dict-annotated fields already merged to `{}` through the other branch, so `--options` on its own works too. Test added covering both.

Reproductions I re-ran alongside it, to make sure the empty path didn't shift anything: `CliPositionalArg[CliVariadicArg[list[str]]]` (empty and with values), `list[int]`, a required variadic, and `Union[dict[str, str], list[str]]`. The rest of the suite is unchanged — the only failures locally are `test_source_json.py::TestTraversableSupport` (2), which fail the same way on the branch point.

gyanu2507:
@hramezani the remaining `cli_args=['--param']` case is covered now — `_merged_list_to_str` returns `'[]'` for an empty variadic, with `test_cli_variadic_named_arg_no_values`. Ready for another look.

pydantic_settings/sources/providers/cli.py
```diff
@@ -1199,9 +1205,12 @@ def _add_parser_args(  # noqa: C901
 
     def _convert_append_action(self, kwargs: dict[str, Any], field_info: FieldInfo, is_append_action: bool) -> None:
         if is_append_action:
-            kwargs['action'] = 'append'
-            if _annotation_contains_types(field_info.annotation, (dict, Mapping), is_strip_annotated=True):
-                self._cli_dict_args[kwargs['dest']] = field_info.annotation
+            if _CliVariadicArg in field_info.metadata:
+                kwargs['nargs'] = '*'
```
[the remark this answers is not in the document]
gyanu2507: `CliVariadicArg[dict[str, str]]` goes through the existing dict parser — `test_cli_variadic_dict_arg` checks `--options a=b c=d` → `{'a': 'b', 'c': 'd'}`. Also merged latest `main`.

gyanu2507:
All of the review notes on this PR should be covered on the current tip, including the bare `--param` → `[]` case from the last thread.

Happy to adjust further if anything still looks off.

hramezani:
Thanks @gyanu2507 for the update. I will check it later

pydantic_settings/sources/providers/cli.py
hramezani: Fixed in 9553d8e — verified at fef4748: bare `--param` → `[]`, bare `--options` → `{}`, and an explicit empty-string value (`--param ''`) still correctly gives `['']`. Also confirmed the early return in `_merged_list_to_str` doesn't change behavior for the shared append path (`--param '[]'` behaves identically to `main`). Full test suite passes. LGTM, thanks!

hramezani (approved):

hramezani:
Thanks @gyanu2507

PR #953, opened 2026-08-28
docs: add nested model alias example for env_prefix_target

MohammedAlkindi:
Follow-up to the note in #934: `env_prefix` plus a string alias on a nested model looks broken at first glance (`APP_SUB_VAR1` with `env_prefix='APP_'` and `alias='SUB'` gives `submodel=None`), but this is the documented `env_prefix_target='variable'` default: the alias matches unprefixed (`SUB_VAR1`), and `env_prefix_target='all'` or `'alias'` is what makes the prefixed spelling match.

The `env_prefix_target` docs show only a scalar example, and no test covers the nested case. This adds a nested worked example beside the existing one and a test parametrized over `'variable'`, `'alias'` and `'all'`.

## Testing

- `pytest tests/test_settings.py`: 247 passed, 7 skipped (the new test runs 3 cases).
- Full suite on Windows: 580 passed, 70 skipped, 6 failed; all six are pre-existing symlink-privilege failures in `test_source_nested_secrets.py`, identical on clean `main` (1ef0864).
- `ruff check` and `ruff format --check`: clean.
- The docs-example harness skips on Windows, so I checked the new snippet against pytest-examples' black and ruff configs locally and verified its printed output against a live run.

Drafted with AI assistance; I ran the new test and the test_settings.py suite on Windows with Python 3.13.13 and checked the example's output lines myself.

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -317,6 +317,53 @@ print(TargetVarSettings().model_dump())
 
 1. `env_prefix` will be ignored and the value will be read from `FooAlias` environment variable.
 
+The same applies to an alias on a nested model field:
+
+```py
+import os
+
+from pydantic import BaseModel, Field
+
+from pydantic_settings import BaseSettings, SettingsConfigDict
+
+
+class SubModel(BaseModel):
+    var1: str | None = None
+
+
+class NestedSettings(BaseSettings):
+    # default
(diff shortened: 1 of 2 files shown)


hramezani (approved):

hramezani:
Thanks @MohammedAlkindi

PR #955, opened 2026-09-01
fix: read JSON and YAML config as UTF-8, not the locale encoding

MohammedAlkindi:
`json_file_encoding` and `yaml_file_encoding` default to `None`, so `file_path.open(encoding=None)` falls back to the locale encoding. On a non-UTF-8 Windows code page a UTF-8 config decodes to mojibake, or raises `UnicodeDecodeError` when a byte has no mapping.

Measured here with `locale.getpreferredencoding(False)` = `cp1252`, on bytes `{"name": "caf\xc3\xa9"}`:

```
open(encoding=None)    -> '{"name": "cafÃ©"}'
open(encoding="utf-8") -> '{"name": "café"}'
```

This completes d2fdeda (#917), which fixed exactly this class for the secrets sources and left these two readers behind. Everything else already pins UTF-8: `dotenv.py:84` uses `encoding or 'utf8'`, and `secrets.py:137` and `nested_secrets.py:206` both pass `encoding='utf-8'`. Both changes here mirror the dotenv form, so the explicit `*_file_encoding` override still wins; it had no test coverage before and now has some.

Four tests, two per provider, covering the UTF-8 default and the override. Both default tests fail on the unpatched tree with `UnicodeDecodeError` while both override tests pass, so they discriminate the two paths.

Suite: 645 passed before, 649 after, 6 failed either side. Those 6 are pre-existing `test_symlink*` failures in `test_source_nested_secrets.py` from `WinError 1314` on this machine. `ruff check` and `ruff format --check` clean.

No behaviour change on a UTF-8 locale, which is why CI never caught it.


hramezani (approved):

hramezani:
Thanks @MohammedAlkindi

PR #901, opened 2026-07-06
Warn when using fields not fully resolved at instantiation

Viicos:
Fixes https://github.com/pydantic/pydantic-settings/issues/775, closes https://github.com/pydantic/pydantic-settings/pull/855.

https://github.com/pydantic/pydantic-settings/pull/855 had a wrong approach, at it was scoped per source, which would result in duplicate warnings.

Issue #193, opened 2023-11-27
Feature Request: Make it possible to specify environment variables for fields again


palle-k:
With Pydantic 1.x, it was possible to set an env name to populate a field from, using `Field(..., env="SOME_ENV_VAR")`. This behavior was replaced in Pydantic-Settings with `Field(...validation_alias="some_var_name")`. However, the new approach does not allow specific environment variables to be loaded in nested settings:

```python
class NestedModel(BaseModel):
    foo: Annotated[str, Field(..., validation_alias="foo")]  # does not actually load "FOO", instead loads "NESTED__FOO"
    bar: str


class Settings(BaseSettings):
    nested: NestedModel

    model_config = SettingsConfigDict(env_nested_delimiter="__")
```

Here, I would like to populate `NestedModel.foo` from the environment variable `FOO`, while `NestedModel.bar` should be populated using the environment variable `NESTED__BAR`. However, as settings sources pre-process env variables using the nested delimiter, `NestedModel` cannot access `FOO` in any way. 

I would appreciate, if this behavior was possible again, i.e. if there was a way to source a field from any environment variable in a nested model, not just ones that are provided by applying `env_nested_delimiter`, like it was in Pydantic 1.x.

---

### Attempted Workarounds

Two workarounds have been attempted, each having its own set of drawbacks.

#### Using BaseSettings for the nested model class

This workaround was able to target individual environment variables, as this created another "top-level", where I had access to any env variable. However, this approach has the drawback that `model_config` (and potentially `settings_customize_sources`) need to be duplicated across both the nested and root settings and that the nested model by default attempts to populate all fields itself from the environment directly, which can result in some conflicts if a field from the top level scope is named like a field from the nested scope (e.g. if there's an env var `FOO` and another env var `NESTED__FOO`)

#### Changing settings sources

This approach is rather messy, as multiple sources may needs to be modified to include the updated behavior. Furthermore, as the settings sources only have access to the top level fields, this approach is rather messy - e.g. when hooking into `explode_env_vars` to inject top-level variables into nested dictionaries and it suffers from the same drawback as the other approach.

DanielHabenicht:
@palle-k was this implemented or did you find a better workaround than using `BaseSettings`?

palle-k:
@DanielHabenicht we switched to using settings sources and using BaseSettings for the nested models.

DanielHabenicht:
Alright, as it is not implemented are you interested in reopening the issue?
[...]

PR #956, opened 2026-09-03
fix: warn when a field is declared with the v1 env kwarg

aktanazat:
## Change Summary

`Field(..., env='VAR')` was the pydantic v1 spelling for binding a field to an
environment variable. In v2 `env` is not a known `Field` argument, so pydantic
stores it in `json_schema_extra` and every settings source ignores it. The field
is never read from `VAR`.

Today this fails in two ways, neither of which names the cause:

- A required field raises `Field required [type=missing]` while the variable is
  set, which points the reader at the environment instead of the declaration.
- An optional field keeps its default and nothing fails, so the process runs on
  the wrong value.

Reproduced on `main` (a7134cd) with pydantic 2.13.4:

```python
import os
from pydantic import Field

os.environ['BACKEND_API_URL'] = 'http://from-env'


class RequiredCase(BaseSettings):
    api_url: str = Field(..., env='BACKEND_API_URL')


class OptionalCase(BaseSettings):
    api_url: str = Field(default='default-sentinel', env='BACKEND_API_URL')


try:
    RequiredCase()
except Exception as e:
    print(e)
print('resolved:', OptionalCase().api_url)
```

```
1 validation error for RequiredCase
api_url
    For further information visit https://errors.pydantic.dev/2.13/v/missing
resolved: default-sentinel
```

The only signal is pydantic's own `PydanticDeprecatedSince20`: "Using extra
keyword arguments on `Field` is deprecated ... (Extra keys: 'env')". It names the
key but not the consequence, and it says nothing about the environment binding
being dropped. This has come up before: #171 was reported with exactly this
symptom and answered by hand with "use `alias`", with no code change. #193 asks
for the v1 behaviour back, which this PR does not do.

This PR reports the dropped binding where the sources already report field
declarations they cannot honour. `_warn_if_field_info_incomplete` (added in #901)
runs on every field of every source and deduplicates through a shared
per-resolution `InitState` set, so the new check goes there and the helper is
renamed to `_warn_if_field_info_unsupported`.

After the change, both cases warn once per field per resolution:

```
IgnoredEnvKwargWarning: Field 'api_url' is declared with `env='BACKEND_API_URL'`,
the pydantic v1 spelling. It is stored in `json_schema_extra` and ignored, so the
field is not read from that environment variable. Use
`validation_alias='BACKEND_API_URL'` instead.
```

Warning rather than error follows the precedent in
`pydantic_settings/exceptions.py`: `IncompleteFieldDefinitionWarning` is a
`UserWarning` for a declaration the sources cannot resolve, and resolution
continues. `env` in `json_schema_extra` can also be a deliberate schema key, so
raising would break working code.

Reviving v1 `env=` support is not proposed. The behaviour stays the same, it
stops being silent.

## Related issues/PRs

- #171: same symptom, closed with a manual answer and no code change.
- #193: feature request to restore v1 per-field env names. Out of scope here.
- #901: added `IncompleteFieldDefinitionWarning` and the per-field warning hook
  this PR extends.

### DIFF
--- a/pydantic_settings/__init__.py
+++ b/pydantic_settings/__init__.py
@@ -1,4 +1,4 @@
-from .exceptions import IncompleteFieldDefinitionWarning, SettingsError
+from .exceptions import IgnoredEnvKwargWarning, IncompleteFieldDefinitionWarning, SettingsError
 from .main import BaseSettings, CliApp, SettingsConfigDict
 from .sources import (
     CLI_SUPPRESS,
@@ -53,6 +53,7 @@ __all__ = (
     'EnvSettingsSource',
     'ForceDecode',
     'GoogleSecretManagerSettingsSource',
+    'IgnoredEnvKwargWarning',
     'IncompleteFieldDefinitionWarning',
     'InitSettingsSource',
     'JsonConfigSettingsSource',
--- a/pydantic_settings/exceptions.py
+++ b/pydantic_settings/exceptions.py
@@ -8,3 +8,12 @@ class IncompleteFieldDefinitionWarning(UserWarning):
     A field definition is incomplete when its annotation contains unresolved forward references,
     in which case settings sources may fail to correctly resolve its value.
     """
+
+
+class IgnoredEnvKwargWarning(UserWarning):
+    """Warning emitted when a field is declared with the pydantic v1 `env` keyword argument.
+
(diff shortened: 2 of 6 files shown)


hramezani:
Thanks @aktanazat, for the PR.

I think it is too late for this warning. We introduced Pydantic V2 and pydantic-settings in 2023.

PR #959, opened 2026-09-04
fix: strip a case-normalized env_prefix from dotenv extras

MohammedAlkindi:
With `extra='allow'` and an `env_prefix` that is not already lower-case, the prefix is never stripped from extra dotenv vars, so the field lands under its full prefixed name.

`self.env_vars` comes from `parse_env_vars()`, which lower-cases every key when `case_sensitive` is false, the default. `self.env_prefix` is stored raw and never case-normalized, so `env_name.startswith(self.env_prefix)` cannot match and the `else` branch keeps the unstripped name.

Given a `.env` containing `XXX__FOO=1` and `XXX__EXTRA_VAR=extra_value`, on main:

```python
class Settings(BaseSettings):
    model_config = SettingsConfigDict(extra='allow', env_file=p, env_prefix='XXX__')
    foo: str = ''

Settings().model_dump()
```

returns `{'foo': '1', 'xxx__extra_var': 'extra_value'}`, where the extra was expected under `extra_var`.

Declared fields are unaffected, since those go through `_extract_field_info`, which already applies `_apply_case_sensitive`. Only the extras branch was reading the raw prefix. The fix reuses that same helper, so a case-sensitive config is untouched.

One test added, failing before. Suite is 7 pre-existing failures either side with an identical set by name, all symlink and toml cases this machine cannot run; passed goes 644 to 645. `ruff check` and `ruff format --check` clean.


hramezani (approved):

hramezani:
Thanks @MohammedAlkindi

PR #961, opened 2026-09-05
Fix nested secrets precedence for case-insensitive names

### LINKED ISSUE #960: Nested secrets directory precedence changes with filename casing
### Bug

`NestedSecretsSettingsSource` does not always honor last-directory-wins precedence when filename matching is case-insensitive.

Create three secret directories in this order:

```text
first/TOKEN  -> first
second/token -> second
third/TOKEN  -> third
```

Configure `secrets_dir=[first, second, third]` and use a settings model with `token: str`, returning `NestedSecretsSettingsSource(file_secret_settings)` from `settings_customise_sources`.

Expected: `Settings().token == 'third'` with the default case-insensitive matching.

Actual on current main: `Settings().token == 'second'`.

The raw dictionaries are merged before their keys are case-normalized. Replacing the original `TOKEN` value in the third directory does not change that key's insertion position, so the earlier lowercase `token` is processed last during normalization and wins incorrectly.

Normalizing each directory's names before merging preserves the documented priority. I added a regression covering the three-directory case and a case-sensitive control; the former fails before the fix and both pass afterward. The full project checks also pass.

### Reproduction

```python
from pathlib import Path
from tempfile import TemporaryDirectory
from pydantic_settings import BaseSettings, NestedSecretsSettingsSource, SettingsConfigDict

with TemporaryDirectory() as root:
    directories = [Path(root) / str(i) for i in range(3)]
    for directory, filename, value in zip(directories, ['TOKEN', 'token', 'TOKEN'], ['first', 'second', 'third']):
        directory.mkdir()
        (directory / filename).write_text(value)

    class Settings(BaseSettings):
        token: str
        model_config = SettingsConfigDict(secrets_dir=directories)

        @classmethod
        def settings_customise_sources(cls, settings_cls, init_settings, env_settings, dotenv_settings, file_secret_settings):
            return (NestedSecretsSettingsSource(file_secret_settings),)

    print(Settings().token)
```

Only temporary files with synthetic values are used. No credentials or cloud services are needed. Prepared with OpenAI Codex assistance and verified locally.

### Environment

```text
pydantic 2.13.4
pydantic-core 2.46.4
pydantic-settings 2.15.0 (current main source)
Python 3.13.9
macOS arm64
```

mehuljariwala:
Fixes #960.

Normalize each secrets directory's filenames before combining its values with earlier directories. Merging raw names first can violate last-directory-wins precedence: directories containing `TOKEN=first`, `token=second`, and `TOKEN=third` currently yield `second` when matching is case-insensitive. Updating the original `TOKEN` key does not move its insertion position, so the older lowercase spelling wins during the final normalization.

Add a public Settings regression for that ordering and a case-sensitive control. As in the existing case-sensitive environment tests, the fixture uses a plain environment mapping so Windows environment normalization does not override the requested control setting. The former fails before the change; both pass afterward. Empty directory lists and all other source tests continue to pass.

Validation: `make` passed Ruff checks and formatting, mypy for all 22 source files, and the full test suite (750 passed, 5 skipped), then generated the coverage report. Python 3.13.9, Pydantic 2.13.4. Only synthetic temporary files were used; no real secret directories or cloud services were accessed.

Prepared with OpenAI Codex assistance; the failure and verification were executed locally. The initial Windows CI failure exposed the inherited environment fallback in the test fixture; the fixture now isolates that behavior while leaving the production change focused on directory merging.

### DIFF
--- a/pydantic_settings/sources/providers/nested_secrets.py
+++ b/pydantic_settings/sources/providers/nested_secrets.py
@@ -1,7 +1,6 @@
 import os
 import warnings
 from collections.abc import Iterator
-from functools import reduce
 from pathlib import Path
 from typing import TYPE_CHECKING, Any, Literal, Optional
 
@@ -120,19 +119,17 @@ class NestedSecretsSettingsSource(EnvSettingsSource):
         self.env_parse_none_str = None  # update manually because of None
 
         # update parent members
-        if not len(self.secrets_paths):
-            self.env_vars = {}
-        else:
-            secrets = reduce(
-                lambda d1, d2: dict((*d1.items(), *d2.items())),
-                (self.load_secrets(p) for p in self.secrets_paths),
+        secrets: dict[str, str | None] = {}
+        for path in self.secrets_paths:
+            secrets.update(
+                parse_env_vars(
+                    self.load_secrets(path),
+                    self.case_sensitive,
+                    self.env_ignore_empty,
+                    self.env_parse_none_str,
+                )
             )
-            self.env_vars = parse_env_vars(
-                secrets,
-                self.case_sensitive,
-                self.env_ignore_empty,
-                self.env_parse_none_str,
-            )
+        self.env_vars = secrets
 
     def validate_secrets_path(self, path: Path) -> None:
         if not path.exists():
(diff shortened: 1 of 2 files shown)


mehuljariwala:
please review

Issue #946, opened 2026-08-26
Allow marking a field as not loadable from settings sources (model constant)


albertopasqualetto:
### Problem

Some fields are contracts the code depends on, not per-environment configuration.

Today there are only two states: it's a field (and any source can override it), or it isn't (and it disappears from the model).

### Current workaround and what it costs

`ClassVar` removes the attribute from `model_fields`, so every source ignores it — `_get_model_fields()` returns only `model_fields`. It works, but the value is then absent from `model_dump()`, absent from the JSON schema, and no longer settable via init kwargs.

Note `Field(exclude=True)` does not help: it only affects serialisation, not loading.

### Proposal

An annotation marking a field as excluded from external sources, keeping it a normal field otherwise.
This is similar to what proposes #521 

Two variants worth separating:

1. **exclude from specific sources** (e.g. not env, but still init or secrets)
2. **model constant** — only settable in the class definition

While the annotation is not implemented, I suggest to add in the documentation the existance of the workaround in a note.

Issue #521, opened 2025-01-11
Disable loading field from certain sources via annotation


jvllmr:
In one of my newer projects I plan to use `pydantic-settings` paired with `pydanclick` for a declarative way to load settings from cli args and other sources such as `pyproject.toml` at the same time.

When thinking about it I notices there could be cases where I don't want users to configure things from a specific source. For instance a `--clean` or `--clear` for creating a clean build in a build tool should only be possible via cli args in my opinion. Taking in this arg from other sources could lead to potentially destructive behaviour.

I think using an annotation would be the best way to implement this. I imagined it as something like this:

```python
from pydantic_settings import BaseSettings, SettingsFieldOpts
from typing import Annotated

class BuildCmdSettings(BaseSettings):
    clean: Annotated[bool, SettingsFieldOpts(disable_sources=["PyprojectTomlConfigSettingsSource"])] = False
```
SettingsFieldOpts would be a new type which could hold more options for a field if needed in the future.

jvllmr:
Also selecting resources and ignoring all other resource could be helpful as well.

Geo5:
If you are fine with grouping arguments a bit differently, then i think it is possible to achieve what you want to do by creating separate `BaseSettings` with different `settings_customise_sources` implementations. I think you then also need to set `extra="ignore"` in the `model_config` to be able to load multiple independent sources, at least from CLI.

PR #962, opened 2026-09-06
Exclude fields from setting sources

BenYang12:
I added NoExternalSources to prevent external settings sources from populating marked fields, 
while preserving constructor inputs, defaults, validation, serialization, and JSON Schema.

Skip excluded fields before environment decoding, prevent dotenv extras from
reintroducing them, and filter file and custom source results without mutating source
data or removing unrelated alias-path siblings. Generated CLI arguments and CLI
serialization omit marked fields.

This implements the external-source exclusion variant discussed in #946. It applies to
fields directly on the settings model; it does not implement immutable constants or per-
source selection. Related: #521.

  Verification:
  - Final full suite: 770 passed, 5 skipped.
  - Regression tests: 21 failed without the implementation; all 22 passed with it.
  - Ruff, formatting, mypy, and repository pre-commit checks passed.


hramezani:
Thanks for putting this together, @BenYang12 — clearly a lot of careful work here, especially the test coverage. I'm going to close it, though, as I don't think this is the right approach. Reasoning, so it's useful for later:

**The annotation can't grow into what #946 asks for.** #946 requests exclusion from *specific* sources and a true model constant. This implements a third thing — all-or-nothing exclusion from every source except init and defaults — and `class NoExternalSources: ...` is a bare, non-parameterized marker. Adding per-source selection (#521) or constant semantics later would mean a second annotation and a second interacting code path in the same places.

**Filtering on raw source keys creates the alias problem the PR then defends against.** The central filter operates on aliases, alias paths, and env names rather than resolved field names. That's what produced the two issues flagged in review, and fa8739e responds by *detecting* overlaps and raising `SettingsError` — ~100 lines of alias arithmetic in `sources/base.py`. Two fields sharing a `validation_alias` is legal today (this PR's own `test_unmarked_shared_aliases_are_unchanged` confirms it), and under this change that becomes a hard error as soon as one field is marked. A new opt-in annotation shouldn't turn a previously-valid model into a startup failure. Filtering on resolved field names instead — as `EnvSettingsSource` already does in its own loop — makes the overlap question disappear along with all that supporting code.

**Enforcement is spread wider than needed.** `_settings_build_values` already filters every non-init source, which is most of the feature. On top of that there's handling in `EnvSettingsSource.__call__`, two loops in `DotenvSettingsSource.__call__`, and four sites in `cli.py`. Some is genuinely source-specific (CLI parser generation, `CliApp.serialize()`, dotenv extras), but env and the generic sources are already covered centrally. The `_is_init_source` flag threaded through `main.py` is the part I'd most want gone — it lets one call site tell apart two uses of the same class when the caller already knows which is which.

**Minor:** the metadata checks use `field.metadata`, while the existing `NoDecode` check in the same file uses `_get_field_metadata(field)`, which also unwraps `Annotated` inside type aliases.

PR #965, opened 2026-09-10
Add inherited dotenv file fallbacks

### LINKED ISSUE #71: feature request: Load multiple .env files from different paths, or additionally support .ini files
Currently only one .env file is supported according to the source code, but sometimes the submodules of the project will have some custom variables that need to be set, and all these variables are not very well managed in one .env file, and it is easy to make changes to the wrong situation. For this reason, I think it should be managed in separate .env files

Consider multiple .env source reads, or dynamic .ini reads (this part of the dynamic allows the user to stitch the path to the ini file based on the module path)

nickita-khylkouski:
Closes #71.

Adds an opt-in env_file_inherit setting so subclasses can retain parent dotenv files as fallbacks while their own files and explicit runtime files take precedence. The implementation handles multiple inheritance, duplicate paths, disabling dotenv loading, and existing source precedence.

Tests:
- 821 tests passed, 5 skipped
- Ruff check and format passed
- mypy passed
- documentation examples passed
- package build passed


hramezani:
Thanks for the PR and for the thorough work here, @nickita-khylkouski — the tests and docs are clearly carefully done.

That said, we don't want to add this feature to `pydantic-settings`.

First, on #71 itself: the original request was support for loading multiple `.env` files, and that has been supported for a long time — `env_file` accepts a list or tuple, with later files taking precedence. See [the docs](https://pydantic.dev/docs/validation/latest/concepts/pydantic_settings/#dotenv-env-support). So the issue this PR references is already resolved, and we'll close it separately.

What this PR implements is a different thing: having a subclass inherit its parent classes' `env_file` as fallbacks rather than replacing it. We'd rather not add a new public config option for this, because:

- It's already achievable today by listing the files explicitly: `model_config = SettingsConfigDict(env_file=['parent.env', 'child.env'])`. The inheritance option mainly saves repeating a path.
- It adds a permanent public knob (`env_file_inherit` in `SettingsConfigDict`, as a class kwarg, as `_env_file_inherit`, and as a `DotEnvSettingsSource` parameter) plus MRO-walking behaviour that we'd have to maintain and reason about alongside Pydantic's own config merging.
- For the cases where the explicit list isn't enough, users can build a [custom settings source](https://pydantic.dev/docs/validation/latest/concepts/pydantic_settings/#customise-settings-sources) — subclassing `DotEnvSettingsSource` and computing the file list from the MRO is a good fit and keeps the logic in the codebase that needs it.

Closing for those reasons. We do appreciate the effort that went into this, and sorry that it went into a feature we're not taking.

nickita-khylkouski:
No problem!

PR #969, opened 2026-09-11
fix: preserve last-directory nested secrets precedence for mixed-case names

### LINKED ISSUE #960: Nested secrets directory precedence changes with filename casing

cpruijsen:
Each directory is now passed through `parse_env_vars` before the merge, so mixed-case names collide on the same key and a later directory overwrites. That is the same order DotEnvSettingsSource already uses for multiple env files. Fixes #960.

When `secrets_dir` is a list, NestedSecretsSettingsSource is documented as last-directory-wins. Under the default case-insensitive matching, NestedSecretsSettingsSource merged raw filenames first, then lowercased keys. Python dicts keep a key's insertion position when its value is replaced, so directories `TOKEN=first`, `token=second`, `TOKEN=third` produced `token=second` instead of `third`.

Parse then `dict.update`, matching `_read_env_files`. Alternative was keeping `reduce` and wrapping `parse_env_vars` in the generator. Can switch if you would rather keep `reduce`.


MohammedAlkindi (commented):
The fix is right, and the mechanism is worth stating plainly: `parse_env_vars` lowercases keys when case-insensitive, so merging raw filenames first left the earlier `TOKEN` holding the slot and a later `token` only overwrote its value under the old key. Normalizing each directory before `update` puts both under one key, so the last directory wins. Matching `_read_env_files` is the right precedent to follow.

Two things.

There is a second PR for the same issue, #967, opened about four hours earlier. It makes the same per-directory `parse_env_vars` move, keeping `reduce` rather than the explicit loop. Worth consolidating rather than carrying both forward.

Your test docstring says `EnvSettingsSource._load_env_vars` force-disables `case_sensitive` on Windows, which is why you monkeypatch `_environ_is_case_insensitive` to get a meaningful control. If that override does reach this source, then `secrets_case_sensitive=True` is equally unreachable on Windows in production, and the parametrized `True` case asserts behavior no Windows user can actually observe. Is applying an `os.environ` rule to secret files, which are not environment variables, intended here?

---
_Generated by [Claude Code](https://claude.ai/code)_

PR #972, opened 2026-09-16
Fix CLI append-action detection for collection subtypes

mmdverse:
The append-action check compares the annotation to `(list, set, dict, Sequence, Mapping)` by exact origin, so anything else falls through. `CliVariadicArg[tuple[str, ...]]` fails with `CliVariadicArg requires a list, set, dict, Sequence, or Mapping type`, which is odd since a tuple is a Sequence. `frozenset`, `typing.Tuple`, `MutableSequence`, `MutableMapping` hit the same. Plain fields aren't registered as append actions either, so repeated options aren't merged and you have to pass JSON.

Match subclasses of Sequence/Set/Mapping instead. str/bytes/bytearray are left out, they're sequences too but they should keep taking a single value.

```python
class Config(BaseSettings):
    files: CliVariadicArg[tuple[str, ...]] = ()
    tags: frozenset[str] = frozenset()
```

`--files a b` raised before, now gives `('a', 'b')`. `--tags x --tags y` gives `frozenset({'x', 'y'})`.

Tests for named/positional variadic, plain and bare collection subtypes, and str/bytes staying scalar.


hramezani (approved):

hramezani:
Thanks @mmdverse

PR #973, opened 2026-09-16
Fix abstract Set annotations not being treated as complex

mmdverse:
`set` and `frozenset` are complex annotations, but `collections.abc.Set` and `MutableSet` are missing from that tuple, so env vars for those fields never get decoded:

```python
class Settings(BaseSettings):
    fruits: Set[str] = frozenset()
```

`FRUITS='["empire", "honeycrisp"]'` gives `ValidationError: Input should be a valid frozenset`, while the same field typed as `frozenset[str]` parses fine. `typing.AbstractSet` is the same origin, so it's affected too.

Add `Set` to the tuple in `_annotation_is_complex_inner()`.

Unit test for the detection + an env var test.


hramezani (approved):

hramezani:
Thanks @mmdverse

PR #976, opened 2026-09-17
Preserve CliVariadicArg values in CliApp.serialize argparse/env styles

glaziermag:
`CliApp.serialize(..., list_style='argparse')` repeats the option for each list item, but `CliVariadicArg` replaces earlier occurrences, so parsing the emitted arguments silently keeps only the last item. Emit a variadic option once, followed by all its values; the same grouping also preserves dictionary entries with `dict_style='env'`.

```python
from pydantic_settings import CliApp, CliVariadicArg

class Settings(BaseModel):
    param: CliVariadicArg[list[str]]

settings = Settings(param=['a', 'b'])
args = CliApp.serialize(settings, list_style='argparse')
loaded = CliApp.run(Settings, cli_args=args)
print(args)
print(loaded.param)
assert loaded.param == settings.param
```

Before: `['--param', 'a', '--param', 'b']` parses as `['b']`.
After: `['--param', 'a', 'b']` parses as `['a', 'b']`.

The [variadic-option documentation](https://github.com/pydantic/pydantic-settings/blob/e4191877f2954d7b9857a407c8bacb3958940818/docs/index.md#L1167-L1177) explicitly describes repeated options replacing previous values. The serializer now respects the parser's `nargs` setting when choosing whether to repeat a flag.

Validation: the reproduction failed identically twice on the base commit. Four of six added list/dictionary-style regression cases failed before the fix; all six pass after it. The full local suite passes (793 passed, 5 skipped; Python 3.12.12, Pydantic 2.13.5). CI's second local test mode, without PyYAML, also passes (754 passed, 29 skipped). All pre-commit hooks pass; the changed CLI module has 100% statement/branch coverage.

AI assistance: prepared and tested by Codex.


hramezani (approved):

hramezani:
Thanks @glaziermag

PR #977, opened 2026-09-17
Honor cli_exit_on_error=False in CliSubCommand parsers

glaziermag:
With `cli_exit_on_error=False`, a `CliSubCommand` missing a required positional argument still exits with status 2 instead of raising `SettingsError`. Child `_CliInternalArgParser` instances keep their default error policy because only the root parser receives the configured value. This change propagates that policy to internal subcommand parsers, including nested commands.

Minimal reproduction on `e4191877f2954d7b9857a407c8bacb3958940818`:

```python
from pydantic_settings import BaseSettings, CliPositionalArg, CliSubCommand, SettingsError

class Child(BaseModel):
    value: CliPositionalArg[str]

class Settings(BaseSettings, cli_exit_on_error=False):
    child: CliSubCommand[Child]

try:
    Settings(_cli_parse_args=['child'])
except SettingsError as exc:
    print(f'SettingsError: {exc}')
```

Before the fix, this prints argparse usage and exits with status 2; the `except` block is never reached. After the fix, it prints `SettingsError: error parsing CLI: the following arguments are required: VALUE` and completes normally. The documented [CLI error policy](https://github.com/pydantic/pydantic-settings/blob/e4191877f2954d7b9857a407c8bacb3958940818/docs/index.md#L1796-L1817) promises that this option controls whether the internal parser exits or raises `SettingsError`.

Validation on Python 3.12.12:

- Eight regression cases cover direct/nested subcommands, model configuration/`CliApp.run` overrides, and both error policies. Four cases fail with `SystemExit` before the fix; all eight pass afterward.
- `make test`: 795 passed, 5 skipped.
- `make lint mypy`: passed.

Prepared with Codex; the reproduction and tests above were run locally.


hramezani (approved):

hramezani:
Thanks @glaziermag

PR #979, opened 2026-09-19
Avoid mutating field metadata when resolving type aliases

jakezwang:
Reading a field annotated with a named `Annotated` type alias appends the alias metadata to the shared `FieldInfo`. Creating a subclass or rebuilding the model afterward then runs its validators twice, changing the resulting values.

Keep the combined metadata separate and consult it for both `NoDecode` and `ForceDecode`. Add regressions for inheritance and model rebuilding with a transforming validator.

Validation: full suite passes on Python 3.10 (797 tests) and 3.14 (798 tests), and without PyYAML (759 tests). Pre-commit hooks, type checks, and package builds pass.


hramezani:
Reviewed and verified locally — nice catch, this is a real silent-corruption bug.

I confirmed the problem reproduces on `main`: because `_get_field_metadata` did `metadata = field.metadata` followed by `metadata += meta`, it mutated the shared `FieldInfo` list in place, so the alias metadata got appended again on every instantiation:

```
first:  [2]
metadata after first: [AfterValidator(...), ForceDecode]   # leaked into FieldInfo
derived: [4]                                               # validator ran twice
```

On this branch it's `derived: [2]` with `metadata == []`. Full suite passes here too (798 passed, 5 skipped).

Worth calling out that the two source changes are **both required** and correctly coupled — I checked that applying only the `utils.py` de-mutation without the `base.py` change breaks `ForceDecode` inside a type alias:

```
ValidationError: Input should be a valid list [type=list_type, input_value='[1,2]']
```

because `decode_complex_value` was previously checking raw `field.metadata`, which only ever worked *by accident* thanks to the mutation. Good that you spotted that rather than fixing just one half.

The test is well built: parametrizing over both inheritance and `model_rebuild(force=True)`, and using a transforming validator so the double application is actually observable rather than idempotent.

One nit, which I've pushed as a follow-up commit rather than asking you to round-trip: `metadata = metadata + meta` is a slightly oblique way to express "don't mutate", and is an easy target for a future "simplification" back to `+=`, which would silently reintroduce this exact bug. Switched it to unpacking with a comment explaining why:

```python
if typing_objects.is_annotated(origin):
    _, *meta = get_args(annotation)
    # Build a new list rather than extending in place: `field.metadata` is shared
    # across instantiations, subclasses and rebuilds, so mutating it would append
    # the alias metadata again on every call.
    metadata = [*metadata, *meta]
```

Behaviour is unchanged and the suite still passes. LGTM — merging once CI is green. Thanks!

hramezani (approved):

hramezani:
Thanks @jakezwang

PR #980, opened 2026-09-20
fix: bare AliasPath complexity keyed off name length, not path length

ishan-1010:
`_extract_field_info` derived a bare `AliasPath`'s "is this value complex" flag from `len(alias)`, where `alias` was the path's own string segment, not the path itself. That check only happened to be right when the segment's name was exactly one character long.

```python
from pydantic import Field
from pydantic_settings import BaseSettings, AliasPath

class Settings(BaseSettings):
    api_key: str = Field(default="", validation_alias=AliasPath("SOME_KEY"))
```

`SOME_KEY=hello python -c "..."` raises `SettingsError: error parsing value for field "api_key" from source "EnvSettingsSource"`, because the source tries to JSON-decode a perfectly ordinary string. The mirror case (a multi-segment path whose head is a one-character name) goes the other way: the raw JSON string never gets decoded/navigated and validation fails instead of returning the indexed value.

The same loop also registered every string element of a bare `AliasPath` as its own independent env var candidate, so `AliasPath("a", "bar", "c")` treated `bar` and `c` as top-level env var names alongside `a`, instead of purely in-memory navigation steps once `a` is read and decoded.

Fixed by using only the head segment as the env var name for a bare `AliasPath`, and deriving complexity from the number of segments in the whole path (`len(v_alias) > 1`), matching how the `AliasChoices` branch already does it for each of its own choices.

Added two tests covering both directions of the bug.


hramezani (approved):

hramezani:
Thanks @ishan-1010

